PR #3, opened 2020-03-17
Don't swallow invalid URLs

### LINKED ISSUE #2: Bad URLs in the master package list are currently ignored
In `ReconcilePackageListCommand` the `fetchMasterPackageList` method currently throws away bad URLs from the fetched JSON.

There should never be bad URLs in that file, but if there are, the reconciliation process should report on them.

0xTim:
Will print out the invalid URL and send an error report to Rollbar. Resolves #2


daveverwer:
Thank you Tim! This is wonderful.

PR #4, opened 2020-03-18
Added a `Command` to reconcile the local package list from the master GitHub list.

daveverwer:
This pull request adds the first `Command` to start the periodic update process that the server will run.

`reconcile_package_list` grabs the `packages.json` file from [here](https://github.com/brightdigit/SwiftPMLibrary/blob/master/packages.json) and updates the database to have the same list of packages.


Sources/App/Commands/ReconcilePackageListCommand.swift
```diff
@@ -0,0 +1,116 @@
...
+          let package = Package(url: url)
+          return package.create(on: database)
+        }
+        
+        // Delete any removed packages
+        let packageDeletions = reconciler.packagesToDelete.map { url in
+          return Package.findByUrl(on: database, url: url).map { package -> Future<Void> in
+            print("Deleting \(package.url)")
+            return package.delete(on: database).map {
+              print("Deleted \(package.url)")
```
daveverwer: This `print` is never executed, even though the `print` two lines above it is. I don't know why.

Sources/App/Commands/ReconcilePackageListCommand.swift
```diff
@@ -0,0 +1,116 @@
...
+      for urlString in urlStrings {
+        // This should delete/re-insert a few (5 or so) random packages per run of this command.
+        // ---
+        // This is just debugging code I put in to exercise the creation/deletion code. I was going
+        // to suggest that `ReconcilePackageListCommand` didn't really need testing since the logic
+        // gets tested in `PackageListReconciler`, but the fact that nothing is being deleted makes
+        // that look a little foolish.
+        if Int.random(in: 1...700) == 1 {
+          continue
+        }
```
daveverwer: So I was going to suggest that this class didn't really need tests as the main logic is encapsulated within the `PackageListReconciler` class (which is tested) but then these deletes just don't execute, so I think this might need tests.

I haven't yet implemented them as I'm not sure how to mock the network response of the master package list. Any help here is appreciated.

Sources/App/Commands/ReconcilePackageListCommand.swift
```diff
@@ -0,0 +1,116 @@
...
+
+        // Otherwise, we're all good
+        urls.append(url)
+      }
+
+      return errorNotifications
+        .flatten(on: context.container)
+        .transform(to: urls)
```
daveverwer: @0xTim Thanks again for this code. Now I've had some time with it I don't understand how this line actually merges the two lists. Does `transform` merge two arrays of `Future` objects?

daveverwer:
This isn't ready for merging yet, but it's close and I wanted to get some other eyes on it. I'd appreciate any input on all aspects of this code 👍

I'm very new to this style of programming and have found it challenging so it's probably *full* of bad practices and mistakes.

Tests/AppTests/Models/PackageTests.swift
```diff
@@ -7,45 +7,55 @@ import XCTest
...
-    let urlString = "https://example.com/"
-    let package = Package(urlString: urlString)
-    XCTAssertEqual(package.url, URL(string: urlString))
+    let existingUrl = URL(string: "https://github.com/Alamofire/Alamofire.git")!
+    let existingPackage = try Package.findByUrl(on: database, url: existingUrl).wait()
+    XCTAssertEqual(existingPackage.url, existingUrl)
+
+    let nonExistentUrl = URL(string: "https://github.com/apple/shiny.git")!
+    let nonExistentPackage = try Package.findByUrl(on: database, url: nonExistentUrl).wait()
+    XCTAssertNil(nonExistentPackage)
```
daveverwer: This test fails with:

```
<unknown>:0: error: -[AppTests.PackageTests testFindPackageByURL] : failed: caught error: The operation couldn’t be completed. (App.PackageError error 0.)
```

Presumably, that's the equivalent of a record not found type error. Maybe that's good enough that this second assertion isn't needed?

Sources/App/Commands/ReconcilePackageListCommand.swift
0xTim: `transform` takes one futures and just returns another, it's a good way to throw away futures if all you care was that they were successful. So in this case you have an array of `Future<Void>` which you wait for `flatten`. That returns `Future<[Void]>`, since we don't really care about that we can just use transform to return the array of URLs. Make sense?

Sources/App/Commands/ReconcilePackageListCommand.swift
0xTim: In tests to mock network responses you can create a `FakeClient` and then choose to use that with `config.prefer(FakeClient.self, for: Client.self)` - I have some examples I can dig up if you want

Sources/App/Commands/ReconcilePackageListCommand.swift
daveverwer: Ahhhh so it's not merging the two Future arrays, it's executing the Futures in `errorNotifications` are being executed during the `flatten`?

That makes some sense, but creates another issue in my head now :) I thought I had it figured out that *all* the Futures in a Command were executed *after* the end of the call to `run`... So, I thought that the idea was to just line up a load of Futures and then the system processes them all after `run` completes. But, if I understand your comment correctly, that's not the case and the `errorNotification` futures are being executed during `flatten`?

Sources/App/Commands/ReconcilePackageListCommand.swift
0xTim: That would suggest the delete never actually returns - are you sure they get deleted from the database?

daveverwer: Positive. If you check this code out and run it without modification you'll see (something like) this in the log:

```
[ INFO ] Migrating 'psql' database (/Users/dave/Library/Developer/Xcode/DerivedData/Server-bcflpguawrtkznajgdsdxsgtueua/SourcePackages/checkouts/fluent/Sources/Fluent/Migration/MigrationConfig.swift:69)
[ INFO ] Migrations complete (/Users/dave/Library/Developer/Xcode/DerivedData/Server-bcflpguawrtkznajgdsdxsgtueua/SourcePackages/checkouts/fluent/Sources/Fluent/Migration/MigrationConfig.swift:73)
2020-03-18 15:40:54.146579+0000 Run[38757:5176485] SecTaskLoadEntitlements failed error=22 cs_flags=20, pid=38757
2020-03-18 15:40:54.146664+0000 Run[38757:5176485] SecTaskCopyDebugDescription: Run[38757]/0#-1 LF=0
Deleting https://github.com/SwiftNIOExtras/swift-nio-redis.git
Deleting https://github.com/vapor-community/leaf-provider.git
Deleting https://github.com/johnsusek/flowstack.git
Deleting https://github.com/vapor-community/vapormonitoring.git
Deleting https://github.com/vapor/mysql-kit.git
Deleting https://github.com/Kuniwak/MemoryLeakTestKit.git
Deleting https://github.com/vapor/postgresql.git
Program ended with exit code: 0
```

daveverwer: Lots of `Deleting`, no `Deleted`.

0xTim:
Just a general FYI - in 5.1 you don't need to bother with LinuxMain.swift and `allTests` - if you do `swift test --enable-test-discovery` it will find them for you (Note: you'll need to pass `--enable-test-discovery` to `swift build` when building on Linux as well due to a bug in swift build)

daveverwer:
> Just a general FYI - in 5.1 you don't need to bother with LinuxMain.swift and `allTests` - if you do `swift test --enable-test-discovery` it will find them for you (Note: you'll need to pass `--enable-test-discovery` to `swift build` when building on Linux as well due to a bug in swift build)

I'm so happy to hear that. I was already so sick of this and there's only like 6 tests so far 😂 Removed.

Tests/AppTests/Models/PackageTests.swift
0xTim: You do `XCTAssetThrows` to ensure it throws the correct error?

Sources/App/Commands/ReconcilePackageListCommand.swift
daveverwer: Also, just in case it's not obvious I also checked the database after running this and no records had ever been deleted.

Tests/AppTests/Models/PackageTests.swift
daveverwer: > Presumably, that's the equivalent of a record not found type error.

5 minutes ago Dave is an idiot. It was actually throwing my very own PackageError.recordNotFound error which I defined and told it to throw 😬

Fixed up this test to correctly test for that.

Sources/App/Commands/ReconcilePackageListCommand.swift
0xTim: Which would explain why the second print is never called. Try adding `databases.enableLogging(on: .psql)` in your configure.swift to see what the actual query being executed is

Sources/App/Commands/ReconcilePackageListCommand.swift
0xTim: Well, technically all futures in NIO are hot so they're started as soon as you start then - `flatten` just waits for them to finish

Sources/App/Commands/ReconcilePackageListCommand.swift
daveverwer: > databases.enableLogging(on: .psql)

That's a useful tip! 🤯

OK, so it looks like it's only executing one delete, which is weird:

```
[psql] [2020-03-18 16:14:14 +0000] SELECT current_setting('server_version') AS "version" []
[ INFO ] Migrating 'psql' database (/Users/dave/Library/Developer/Xcode/DerivedData/Server-bcflpguawrtkznajgdsdxsgtueua/SourcePackages/checkouts/fluent/Sources/Fluent/Migration/MigrationConfig.swift:69)
[psql] [2020-03-18 16:14:15 +0000] SELECT COUNT(*) AS "fluentAggregate" FROM "fluent" []
[psql] [2020-03-18 16:14:15 +0000] SELECT * FROM "fluent" ORDER BY "fluent"."batch" DESC LIMIT 1 OFFSET 0 []
[psql] [2020-03-18 16:14:15 +0000] SELECT * FROM "fluent" WHERE "fluent"."name" = ($1) LIMIT 1 OFFSET 0 ["Package" (TEXT)]
[ INFO ] Migrations complete (/Users/dave/Library/Developer/Xcode/DerivedData/Server-bcflpguawrtkznajgdsdxsgtueua/SourcePackages/checkouts/fluent/Sources/Fluent/Migration/MigrationConfig.swift:73)
[psql] [2020-03-18 16:14:15 +0000] SELECT * FROM "packages" []
2020-03-18 16:14:15.040738+0000 Run[39787:5206950] SecTaskLoadEntitlements failed error=22 cs_flags=20, pid=39787
2020-03-18 16:14:15.040814+0000 Run[39787:5206950] SecTaskCopyDebugDescription: Run[39787]/0#-1 LF=0
[psql] [2020-03-18 16:14:15 +0000] SELECT * FROM "packages" WHERE "packages"."urlString" = ($1) LIMIT 1 OFFSET 0 ["https://github.com/rymcol/swiftcron-example.git" (TEXT)]
[psql] [2020-03-18 16:14:15 +0000] SELECT * FROM "packages" WHERE "packages"."urlString" = ($1) LIMIT 1 OFFSET 0 ["https://github.com/appcron/acinteractor.git" (TEXT)]
Deleting https://github.com/rymcol/swiftcron-example.git
[psql] [2020-03-18 16:14:15 +0000] SELECT * FROM "packages" WHERE "packages"."urlString" = ($1) LIMIT 1 OFFSET 0 ["https://github.com/juri/graphqler.git" (TEXT)]
Deleting https://github.com/appcron/acinteractor.git
[psql] [2020-03-18 16:14:15 +0000] DELETE FROM "packages" WHERE "packages"."id" = ($1) [AC034449-873A-4A08-ADAD-65F4E845137A (UUID)]
Deleting https://github.com/juri/graphqler.git
Program ended with exit code: 0
```

It does three `SELECT` statements there and you can see three `print`s being executed too. Then it does one `DELETE` at the end. The UDID of the `DELETE` matches the *first* package (swiftcron-example.git).

Maybe it's possible that `DELETE` is failing?

daveverwer: Just checked and the permissions on the database role look good, so I don't think it's that.

Mordil: It's the same issue as before: you want a `flatMap` instead of a `map`, as you're chaining the `package.delete` Future to the `findByUrl` future.

PR #5, opened 2020-03-18
Fix deletion bug and make it easier to spin up a local DB

0xTim:
Fix the bug with deleting packages

Also adds a script to spin up a Docker container to make developing locally easier

### DIFF
--- a/Sources/App/Commands/ReconcilePackageListCommand.swift
+++ b/Sources/App/Commands/ReconcilePackageListCommand.swift
@@ -29,18 +29,20 @@ final class ReconcilePackageListCommand: Command
         
         // Delete any removed packages
         let packageDeletions = reconciler.packagesToDelete.map { url in
-          return Package.findByUrl(on: database, url: url).map { package -> Future<Void> in
+          return Package.findByUrl(on: database, url: url).flatMap { package -> Future<Void> in
             print("Deleting \(package.url)")
             return package.delete(on: database).map {
               print("Deleted \(package.url)")
             }
           }
         }
-
+        
         // Queue up all of the additions and deletions for execution
-        return packageAdditions.flatten(on: database)
-          .and(packageDeletions.flatten(on: database))
-          .transform(to: ())
+        
+        return map(packageAdditions.flatten(on: database), packageDeletions.flatten(on: database)) { additions, _ in
+          print("Added \(additions.count) additions")
+          print("Finished")
+        }
       }
     }
   }
@@ -56,7 +58,7 @@ final class ReconcilePackageListCommand: Command
     }.flatMap { urlStrings in
       var urls = [URL]()
       var errorNotifications = [Future<Void>]()
-
+      
       for urlString in urlStrings {
         // This should delete/re-insert a few (5 or so) random packages per run of this command.
         // ---
@@ -67,17 +69,17 @@ final class ReconcilePackageListCommand: Command
         if Int.random(in: 1...700) == 1 {
           continue
         }
-
+        
         // If this isn't a valid URL, post an error
         guard let url = URL(string: urlString) else {
           try errorNotifications.append(self.sendInvalidURLString(urlString, on: context))
           continue
         }
-
+        
         // Otherwise, we're all good
         urls.append(url)
       }
-
+      
       return errorNotifications
         .flatten(on: context.container)
         .transform(to: urls)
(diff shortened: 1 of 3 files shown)


startLocalDockerDB.sh
```diff
@@ -0,0 +1 @@
+docker run --name postgres-swiftpackageindex -e POSTGRES_DB=swiftpackageindex_dev -e POSTGRES_USER=swiftpackageindex -e POSTGRES_PASSWORD=password -p 5432:5432 -d postgres
```
Mordil: Why not a `Docker-Compose` file, as that scales a bit better if/when more services are added?

0xTim: Mainly laziness! It's a copy of a script I use for most repos to spin up a DB

Mordil: ~shocked Pickachu~ 😛

Sources/App/Commands/ReconcilePackageListCommand.swift
```diff
@@ -29,18 +29,20 @@ final class ReconcilePackageListCommand: Command
         
         // Delete any removed packages
         let packageDeletions = reconciler.packagesToDelete.map { url in
-          return Package.findByUrl(on: database, url: url).map { package -> Future<Void> in
+          return Package.findByUrl(on: database, url: url).flatMap { package -> Future<Void> in
```
daveverwer: First of all, thank you so much for your help figuring this out @0xTim. You’re being incredibly generous with your time.

Now that you’ve pointed it out, I can see the problem. It’s the Future of a Future thing, again. 😬 But the fact that this fails in such a weird way, with the first DELETE statement getting as far as the database, but not actually deleting anything, and then everything else failing silently at both compile, and run time... it’s terrifying. 😱 I don't even know how I would have started to debug this myself.

I know this was my mistake and a fairly simple one at that, but this whole approach of throwing Futures into the void feels really fragile and opaque. Honestly, it scares me.

I think what I take from this is that using this kind of framework *every part of the system* needs to be tested. Nothing can be left to chance as the entire app is a typo away from silently breaking.

startLocalDockerDB.sh
daveverwer: I started with Docker but quickly got tied in a knot trying to get it to run both a dev and test database at the same time, so I reverted to just running Postgres.app locally. That said, it's fine for this to go in if it helps you.

Sources/App/Commands/ReconcilePackageListCommand.swift
```diff
@@ -29,18 +29,20 @@ final class ReconcilePackageListCommand: Command
...
+        
         // Queue up all of the additions and deletions for execution
-        return packageAdditions.flatten(on: database)
-          .and(packageDeletions.flatten(on: database))
-          .transform(to: ())
+        
+        return map(packageAdditions.flatten(on: database), packageDeletions.flatten(on: database)) { additions, _ in
+          print("Added \(additions.count) additions")
+          print("Finished")
+        }
```
daveverwer: Is there an advantage to doing it this way over using `add`?

The cleanliness I can see is that you no longer need the `transform`, is it as simple as that?

Sources/App/Commands/ReconcilePackageListCommand.swift
0xTim: Yeah it's a weird one and definitely a bit worrying that the compiler didn't pick it up, but the type inference and lack of using the futures masked it all. I must say this is the first time I've ever seen as issue like this be completely hidden

Sources/App/Commands/ReconcilePackageListCommand.swift
0xTim: No more habit. You could do it with `transform` as well, but I removed that to add the print statements.

The advantage of using the variadic `map` or `flatMap` over `and` is that you don't end up in tuple hell when chaining more than 2 futures together

startLocalDockerDB.sh
0xTim: Multiple databases is not something the Postgres container natively supports (though it can be added on with some scripts). Personally I just spin up multiple containers and use different ports for dev and test

Sources/App/Commands/ReconcilePackageListCommand.swift
daveverwer: > you don't end up in tuple hell when chaining more than 2 futures together 

I totally see that, thanks.

PR #17, opened 2020-05-01
Review changes

finestructure:
Hey @daveverwer these are the things we discussed yesterday. I'll go ahead and merge this now into the ongoing `populate-versions` branch - just raising this as an easy way to FYI.

### DIFF
--- a/Sources/App/Commands/Analyzer.swift
+++ b/Sources/App/Commands/Analyzer.swift
@@ -3,7 +3,7 @@ import Vapor
 import ShellOut
 
 
-struct InspectorCommand: Command {
+struct AnalyzerCommand: Command {
     let defaultLimit = 1
 
     struct Signature: CommandSignature {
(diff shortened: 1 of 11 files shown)


Sources/App/Commands/Analyzer.swift
```diff
@@ -3,26 +3,26 @@ import Vapor
 import ShellOut
 
 
-struct InspectorCommand: Command {
+struct AnalyzerCommand: Command {
```
finestructure, line 6: I went with Anaylzer for now to sort of indicate that it's a "heavier" operation. Let me know if you've got better ideas!

Sources/App/Commands/Ingestor.swift
```diff
@@ -26,40 +26,43 @@ struct IngestorCommand: Command {
...
-                        pkg.status = .ok
-                        return pkg.save(on: database)
-                    }
-            } catch {
-                return recordIngestionError(database: database, error: error)
-            }
-        }
+    Package.fetchUpdateCandidates(database, limit: limit)
+        .flatMapEach(on: database.eventLoop) { fetchMetadata(for: $0, with: client) }
+        .flatMapEachThrowing { updateTables(on: database, result: $0) }
```
finestructure, line 31: As discussed, I pulled metadata fetching out into the top level `ingest` command and also broke out the table updates into their own method. Reads nicer now, I think.

Sources/App/Controllers/PackageController.swift
```diff
@@ -58,6 +63,7 @@ extension PackageController {
     enum Command: String {
         case reconcile
         case ingest
+        case analyze
```
finestructure, line 66: This is the API hook to trigger analysis that I added during our walk-through.

Sources/App/Models/Package.swift
```diff
@@ -96,11 +96,12 @@ extension QueryBuilder where Model == Package {
...
 
 
-extension QueryBuilder where Model == Package {
-    func updateCandidates(limit: Int) -> EventLoopFuture<[Package]> {
-        // TODO: filter out updated in last X minutes
-        sort(\.$updatedAt)
-        .limit(limit)
-        .all()
+extension Package {
+    static func fetchUpdateCandidates(_ database: Database, limit: Int) -> EventLoopFuture<[Package]> {
```
finestructure, line 100: Renamed as discussed

Sources/App/Commands/Analyzer.swift
daveverwer: There's no ideal names, but I think having one start with an `A` and one with an `I` will at least make it easier to keep a mental model of which does what 😂

daveverwer:
Fab!

PR #21, opened 2020-05-01
Preparing for populating versions

finestructure:
Clearing the deck before actually creating versions. This includes

- review changes
- relationship accessors and tests between Package/Respository and Package/Version
- fix for reading empty `Version.swiftVersion` from db
- `Package.defaultVersion`
- replace `submit` with `runIfActive`

### DIFF
--- a/Sources/App/Models/Package.swift
+++ b/Sources/App/Models/Package.swift
@@ -96,11 +96,12 @@ extension QueryBuilder where Model == Package {
 }
 
 
-extension QueryBuilder where Model == Package {
-    func updateCandidates(limit: Int) -> EventLoopFuture<[Package]> {
-        // TODO: filter out updated in last X minutes
-        sort(\.$updatedAt)
-        .limit(limit)
-        .all()
+extension Package {
+    static func fetchUpdateCandidates(_ database: Database, limit: Int) -> EventLoopFuture<[Package]> {
+        Package.query(on: database)
+            // TODO: filter out updated in last X minutes
+            .sort(\.$updatedAt)
+            .limit(limit)
+            .all()
     }
 }
--- a/Sources/App/Models/Package.swift
+++ b/Sources/App/Models/Package.swift
@@ -38,6 +38,9 @@ final class Package: Model, Content {
     @Children(for: \.$package)
     var repositories: [Repository]
 
+    @Children(for: \.$package)
+    var versions: [Version]
+
     init() { }
 
     init(id: UUID? = nil, url: URL, status: Status = .none) {
--- a/Sources/App/Models/Package.swift
+++ b/Sources/App/Models/Package.swift
@@ -35,12 +35,22 @@ final class Package: Model, Content {
     @Field(key: "last_commit_at")  // TODO: shouldn't this rather live in Repository?
     var lastCommitAt: Date?
 
+    @Children(for: \.$package)
+    var repositories: [Repository]
+
     init() { }
 
     init(id: UUID? = nil, url: URL, status: Status = .none) {
         self.id = id
         self.url = url.absoluteString
         self.status = status
+    }
+}
+
+
+extension Package {
+    var repository: Repository? {
+        get { repositories.first }
     }
 }
 
--- a/Sources/App/Models/Package.swift
+++ b/Sources/App/Models/Package.swift
@@ -53,7 +53,21 @@ final class Package: Model, Content {
 
 extension Package {
     var repository: Repository? {
-        get { repositories.first }
+        repositories.first
+    }
+}
+
+
+extension Package {
+    var defaultVersion: Version? {
+        // TODO: sas 2020-04-30: find a more convenient way to use this. In order to avoid
+        // fatalErrors from lack of lazy loading, the caller needs to use it on a Package that's
+        // been fetched like so:
+        //   Package.query(on: db).with(\.$versions).with(\.$repositories)
+        // That's awkward. Should instead defaultBranch take a parameter (on: db) and do this
+        // itself?
+        guard let defaultBranch = repository?.defaultBranch else { return nil }
+        return versions.first(where: { $0.branchName == defaultBranch })
     }
 }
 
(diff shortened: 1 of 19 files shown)


Sources/App/Models/Version.swift
```diff
@@ -41,6 +41,7 @@ final class Version: Model, Content {
     @Parent(key: "package_id")
     var package: Package
 
+    // TODO: sas-2020-04-30: Explore folding branch/rag into an enum with associated value String
     @Field(key: "branch_name")
     var branchName: String?
```
finestructure, line 46: I think this is quite an important thing to sort out to avoid having to do queries where need to check for either and/or add a constraint to ensure only one is set. Instead we could try and save to a `revision` field that maps to a `Revision` enum:

```
enum Revision {
  case branch(String)
  case tag(String)
  case commit(String) // unlikely we'll actually need/want this
}
```

Sources/App/Models/Package.swift
```diff
@@ -45,6 +51,27 @@ final class Package: Model, Content {
...
+
+extension Package {
+    var defaultVersion: Version? {
+        // TODO: sas 2020-04-30: find a more convenient way to use this. In order to avoid
+        // fatalErrors from lack of lazy loading, the caller needs to use it on a Package that's
+        // been fetched like so:
+        //   Package.query(on: db).with(\.$versions).with(\.$repositories)
+        // That's awkward. Should instead defaultBranch take a parameter (on: db) and do this
+        // itself?
+        guard let defaultBranch = repository?.defaultBranch else { return nil }
```
daveverwer, line 69: A comment on default branches: Fetching default branch from the GitHub API isn't strictly necessary if we're pulling the entire source locally. The default branch will be selected after a fresh pull, and we shouldn't be switching branches in our local copy of the repository. Or, if you do want to verify it then it can be read out of the local git repository rather than relying on the GitHub API for this.

Sources/App/Models/Package.swift
```diff
@@ -45,6 +51,27 @@ final class Package: Model, Content {
...
 
+extension Package {
+    var repository: Repository? {
+        repositories.first
+    }
+}
+
+
+extension Package {
+    var defaultVersion: Version? {
```
daveverwer, line 62: I'm not quite sure what `defaultVersion` is doing here? Are you referring to the git version tags here? If so, I don't think there should be a "default" version. Instead, we care about (up to) three versions.

1. **Latest**: The latest commit to the default branch.
2. **Latest Stable**: A tagged version that is a valid 3-part semantic version and does *not* have a "pre" part. For example 1.0.0.
3. **Latest Beta**: A tagged version that is a valid 3-part semantic version, and *does* have a "pre" part. For example 1.0.0-beta1.

~~If there is a "default" version it should be the latest version, but really I don't think there should be a default version. Happy to discuss what this is for and why it's needed.~~

daveverwer: Scratch my last sentence above. If there is a default version it should be the **Latest Stable** as that's what Xcode chooses if you don't specify a version when adding a package.

finestructure: Yes, that makes sense!

I think we're on the same page here - the whole version stuff will change quite a bit still tbh and I don't think this is even used anywhere yet.

I'd probably take it out for now - or I'll add a TODO and clean it up later. Let me know!

daveverwer: No need to take it out, let's just add a note around it?

Sources/App/Models/Package.swift
finestructure: Is there any reason to suspect the GH api to be wrong? Because we've already got the default branch info so it's easy to use it.

There's also the remote edge case that we pull a repo and the default branch changes afterwards. We'd sit on a wrong branch then unless we always ensure we track the default one as well, right?

finestructure: But see below, I'd probably rather ditch/ignore this for now!

Sources/App/Models/Version.swift
daveverwer: Sounds good, but I agree we won't need anything tied to a specific commit. Let's just do branch/tag. 👍

Sources/App/Models/Package.swift
daveverwer: > Is there any reason to suspect the GH api to be wrong? Because we've already got the default branch info so it's easy to use it.

Only that it's more work when we support GitLab, which we already have a case for as there are some packages already not being indexed by that lack of support. If we use the local repository, it works for all services.

Of course, it's likely that the GitLab API also supplies the default branch, so it's not a huge issue. I'm happy either way 👍

> There's also the remote edge case that we pull a repo and the default branch changes afterwards. We'd sit on a wrong branch then unless we always ensure we track the default one as well, right?

Oof, that's an excellent point, and you just made me realise my current implementation is vulnerable to this! I just switch to the default branch on clone and only ever pull changes after that. Regardless of which method we use to determine the default branch, we should make sure to check it out before pulling when updating.

Sources/App/Models/Version.swift
daveverwer: I'm curious how an enum like this would be serialised to a database field. Can't wait to see :)

Sources/App/Models/Package.swift
finestructure: WRT to Gitlab - the `projects` api resource seems to carry everything we're currently pulling in from Github: https://docs.gitlab.com/ee/api/projects.html

daveverwer: Then let's just lean on the default branch from those API calls then. It's going to be a definitive source of truth, and I can't really see people needing more services than GitHub/Lab. Maybe BitBucket, but it feels like that's well and truly lost the open-source battle to those other two.

finestructure: Yeah, that one's gone =) 👍

finestructure:
Just checking: I normally don't self-merge into master or shall I take the approval tick as signal to go ahead?

daveverwer:
Ah yes, the approval is good enough! Go for it 👍

PR #22, opened 2020-05-01
Shell mock

finestructure:
Continuation - merge after #21


finestructure:
Meh, this is bad. Unlike Gitlab, Github doesn't seem to recognise when the merge delta shrinks when a parent revision gets merged into `master`. I'll try to re-open the MR to see if it does.

PR #24, opened 2020-05-04
Get versions

finestructure:
Continuation of #23 , needs that one to be merged first


Sources/App/Commands/Analyzer.swift
```diff
@@ -24,48 +24,164 @@ struct AnalyzerCommand: Command {
...
+    }
+    let tags: EventLoopFuture<[String]> = application.threadPool.runIfActive(eventLoop: application.eventLoopGroup.next()) {
+        application.logger.info("listing tags for package \(package.url)")
+        let tags = try Current.shell.run(command: .init(string: "git tag"), at: path)
+        return tags.split(separator: "\n").map(String.init)
+    }
+    // FIXME: also save version for default branch (currently only looking at tags)
+
+    // TODO: sas 2020-05-02: is it necessary to reconcile versions or is delete and recreate ok?
+    // It certainly is simpler.
```
finestructure, line 139: FYI @daveverwer - I _think_ it'd be ok to just wipe and recreate versions (and later products) rather than try and match them up. It's just a lot less hassle than try and line up the adds and deletes with the futures but please let me know if there's a need to track the original creation date of versions. (As far as I can see the original creation date is the only bit of info we'd lose with this approach - in fact creation data would become pretty meaningless for these records.)

Sources/App/Commands/Analyzer.swift
```diff
@@ -24,48 +24,164 @@ struct AnalyzerCommand: Command {
...
+            throw AppError.invalidPackageUrl(package.id, package.url)
+        }
+        // FIXME: here we'll want to be able to use tag or default branch
+        guard let revision = version.tagName else {
+            throw AppError.invalidRevision(version.id, version.tagName)
+        }
+        try Current.shell.run(command: .gitCheckout(branch: revision), at: cacheDir)
+        let json = try Current.shell.run(command: .init(string: "swift package dump-package"), at: cacheDir)
+        // TODO: sas-2020-05-03: do we need to run tools-version? There's a toolsVersion key in the JSON
+        // Plus it may not be a good substitute for swift versions?
```
finestructure, line 169: In the ruby version you call out to `tools-version` but I've seen there's a `toolsVersion` key in the `dump-package` JSON as well (and seems to be populated).

Was there a reason you called `tools-version` or was that just an oversight?

Sources/App/Models/Version.swift
```diff
@@ -54,12 +35,14 @@ final class Version: Model, Content {
     @Field(key: "commit")
     var commit: String?
 
+    // TODO: sas-2020-05-03: currently concatenating os + version - we could save the structure instead
     @Field(key: "supported_platforms")
     var supportedPlatforms: [Platform]
```
finestructure, line 40: This one is important to discuss. Currently we'll have this filled in as

```
["macos_10.15", "ios_13"]
```

(`Platform` is just a `String` typealias). I could imagine we actually want the structured data from the package dump to poke through here:

```
struct Manifest: Codable {
    struct Platform: Codable {
        enum Name: String, Codable {
            case macos
            case ios
            case tvos
            case watchos
        }
        var platformName: Name
        var version: String
    }
    var name: String
    var swiftLanguageVersions: [String]?
    var platforms: [Platform]?
}
```

and save `Manifest.Platform` here instead (i.e. as a `jsonb[]` in the database).

It'll depend a bit on how we need to query/display this info. I mean, we can still get at the bits with PG JSON queries but if we only ever intend to display those labels as fixed strings we might as well save them that way and keep the schema simpler.

Sources/App/Commands/Analyzer.swift
daveverwer: In my experience running `tools-version` is not a great way to detect the Swift version of a package. I think it's worth considering whether we even look at the tools version at all. It's certainly giving incorrect Swift versions for some of the packages in the current system.

Maybe we don't run `tools-version` *or* look at that key in this first release and see how bad things are with missing Swift versions? What do you think?

finestructure: Big upvote from me to ditch it for now :)

I've just started crawling packages in earnest and it seems like `.swift-version` files might be a better heuristic - and one that's not in the swiftpm.co, or is it?

Sources/App/Models/Version.swift
daveverwer: So before looking at the specifics of how this gets represented in code, I think it's worth taking a second to think about how this data is going to be used.

I think the biggest use for this data is not really about the specific version numbers, but more as an indication of general plaform support. If I pick a package, and it supports iOS, and it's been updated in the last 12 months, it almost certainly supports the version of iOS I'm interested in. I think having this data act as a boolean flag for platform support is 90% of what's needed.

I do think we should store the version requirement, but it's only for that last 10%.

As you'll have seen, these are stored as strings for each platform in the current schema:

```
supported_macos_version
supported_ios_version
supported_watchos_version
supported_tvos_version
```

I don't have a huge preference whether these are just strings, or integers mapped to enum values, but I think having them as four columns is a big advantage when it comes to search. 

To answer your original question. I can't see anything right now that would need this data in a format other than a string. So my gut feeling says just go with that? What are your thoughts?

Sources/App/Commands/Analyzer.swift
daveverwer: Yes, `.swift-version` files are going to be a *much* better indicator of version support, and actually, if you look at the to-do list I had it was at the very top of my list for post-release features 😂

Screenshot 2020-05-05 at 09 13 14@2x

daveverwer: In case it wasn't clear, I'm saying ditch it. It's a bad indicator 👍

Sources/App/Commands/Analyzer.swift
daveverwer: I don't think the creation date of package version records is important at all, what's much more important is the version release date but that's not affected by this. So, I don't think there's any reason to need full reconciliation there.

I think I did reconciliation in the original implementation, there's something about the constant churning of integer database IDs that rubs my old-school brain the wrong way, but this uses UUID keys, so that's also not an issue.

Something about constantly deleting and re-inserting the same data (for versions that haven't changed) feels wrong to me, but there's no good reason apart from that.

finestructure: :+1: it's certainly something we could change going forward but I'll keep it simple at the start then!

Sources/App/Models/Version.swift
finestructure: I think this only makes a difference if we're looking to query or aggregate across versions - which seems unlikely, at least on a frequent enough basis that it could turn into a performance issue. Even then, I've dealt with PG databases and millions of rows with JSON queries and not had issues. Plus there's indexes on JSON field properties.

All that is to say, since I expect us to pretty much mostly pull in the whole record and then show the contents, it doesn't really make a big difference whether the fields are in columns or in a JSON field.

An advantage the latter brings is that the AR glasses that'll be introduced next year won't require a schema change :D Or perhaps some way of adding linux for that matter.

What I'd envisioned is not an integer mapped enum but a JSON array that encodes the structure pretty much exactly as laid out:

```
[
	{
	  "name": "macos",
	  "version": "10.10"
	},
	{
	  "name": "ios",
	  "version": "13.0"
	}
]
```

Since we're using Vapor on both backend and frontend, in the end that's a completely transparent decode that just gives you a `[Platform]` array to work with and we don't even have to worry about the details.

You'd still just do something like

```swift
extension Array where Element == Platform {
  var supportedPlatforms: Set<Platform.Name> { Set(map(\.name)) }
}

package.versions.map {
  // some table row template pseudo code
  "<tr>\($0.tagName) \($0.supportedPlatforms)</tr>")
}
```

i.e. we put the "smarts" into `Platform` extensions client side and make the field just a decodable store. If we ever need deeper inspection of the field content relationally, PG still has us covered with its JSON support.

And if we want to add Linux, we'd just add an enum `case linux` and that's it. As long as we don't take out any cases we don't need to do anything else.

Does that make sense?

finestructure:
Ok, I've marked this ready now that #23 is in master.

Unfortunately the diff from #23 is still showing but if I close and re-open we lose the discussion :/

finestructure:
Ah, there might be a workaround: https://stackoverflow.com/a/26986320/1444152

daveverwer:
This is fantastic progress!

Sources/App/Models/Version.swift
daveverwer: Sorry I missed that I hadn't replied to this. I think this is fine.

> All that is to say, since I expect us to pretty much mostly pull in the whole record and then show the contents, it doesn't really make a big difference whether the fields are in columns or in a JSON field.

You're absolutely right, with one potential exception that may affect speed. Conditional search. I'm writing up some other thoughts on Search today (will post shortly) which could potentially affect this decision. Let's revisit it after having that discussion.

PR #26, opened 2020-05-05
End to end testing

finestructure:
@daveverwer These are a few small fixes that cropped up while end-to-end testing processing all whole 2.7k packages.

I’ve now successfully processed everything locally and on my Digital Ocean box *) except for 5 packages I had to delete manually because they brought up the git login prompt (see #27 ) -EDIT: the commit I just added fixes this. It went well all in all!

I’ll send you the db details for the Digital Ocean DB to have a look at the data. We’ve got close to 30k versions and around 26k products (plus the 2.7 packages and repositories) in there now.

Have a look through the data - it should be mostly complete, except for a few minor omissions, I think (default branch versions and fork links, most notably). The main purpose was to run it end to end once to make sure processing isn’t interrupted by errors bubbling up and to have the full table structure with all entities in place.

*) The DO box ran out of disk space with 100 packages left to process 😅 - so that data is almost complete as well. The checkout size is 16.6GB now.

### DIFF
--- a/Sources/App/Commands/Analyzer.swift
+++ b/Sources/App/Commands/Analyzer.swift
@@ -127,6 +127,8 @@ func pullOrClone(application: Application, package: Package) throws -> EventLoop
     return application.threadPool.runIfActive(eventLoop: application.eventLoopGroup.next()) {
         if Current.fileManager.fileExists(atPath: path) {
             application.logger.info("pulling \(package.url) in \(path)")
+            // git reset --hard to deal with stray .DS_Store files on macOS
+            try Current.shell.run(command: .init(string: "git reset --hard"), at: path)
             let branch = package.repository?.defaultBranch ?? "master"
             try Current.shell.run(command: .gitCheckout(branch: branch), at: path)
             try Current.shell.run(command: .gitPull(), at: path)
--- a/Sources/App/Commands/Analyzer.swift
+++ b/Sources/App/Commands/Analyzer.swift
@@ -48,7 +48,7 @@ func analyze(application: Application, limit: Int) throws -> EventLoopFuture<Voi
 
 func analyze(application: Application, packages: EventLoopFuture<[Package]>) throws -> EventLoopFuture<Void> {
     // get or create directory
-    let checkoutDir = Current.fileManager.checkouts
+    let checkoutDir = Current.fileManager.checkoutsDirectory
     application.logger.info("Checkout directory: \(checkoutDir)")
     if !Current.fileManager.fileExists(atPath: checkoutDir) {
         application.logger.info("Creating checkout directory at path: \(checkoutDir)")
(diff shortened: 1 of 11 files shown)


Sources/App/Commands/Analyzer.swift
```diff
@@ -57,18 +81,27 @@ func analyze(application: Application, limit: Int) throws -> EventLoopFuture<Voi
 
     // FIXME: sas 2020-05-04: Workaround for partial flush described here:
     // https://discordapp.com/channels/431917998102675485/444249946808647699/706796431540748372
+    // FIXME: this breaks exposing this via the API:
+    // Precondition failed: BUG DETECTED: wait() must not be called when on an EventLoop.
```
finestructure, line 85: I’ll add an issue to track this, maybe someone can jump in to give a hand.

Sources/App/Commands/Analyzer.swift
```diff
@@ -161,6 +201,11 @@ func getManifest(package: Package, version: Version) -> Result<Manifest, Error>
             throw AppError.invalidRevision(version.id, version.tagName)
         }
         try Current.shell.run(command: .gitCheckout(branch: revision), at: cacheDir)
+        guard Current.fileManager.fileExists(atPath: cacheDir + "/Package.swift") else {
+            // It's important to check for Package.swift - otherwise `dump-package` will go
+            // up the tree through parent directories to find one
+            throw AppError.invalidRevision(version.id, "no Package.swift")
+        }
```
finestructure, line 208: This whole `analyze` function needs some refactoring and cleanup but I don’t want to do that before we’ve fixed the `wait` workaround.

Sources/App/Commands/Analyzer.swift
```diff
@@ -57,18 +81,27 @@ func analyze(application: Application, limit: Int) throws -> EventLoopFuture<Voi
...
+                .flatMapEach(on: application.db.eventLoop) { result -> EventLoopFuture<Void> in
+                    switch result {
+                        case .success:
+                            pkg.status = .ok
+                        case .failure(let error):
+                            application.logger.error("Analysis error: \(error.localizedDescription)")
+                            pkg.status = .analysisFailed
+                    }
+                    return pkg.save(on: application.db)
+            }
```
finestructure, line 104: This whole `analyze` function needs some refactoring and cleanup but I don’t want to do that before we’ve fixed the `wait` workaround.

Sources/App/Commands/Analyzer.swift
daveverwer: I'm not sure I can be much help here, but let's raise an issue for it so it doesn't get forgotten after this is merged.

finestructure: Ah yes, thanks for the reminder to actually paste that in :D

#29

PR #32, opened 2020-05-07
Processing stage

finestructure:
This reviews our candidate selection and adds a `processing_stage` field to `Packages` to track where we are.

### DIFF
--- a/Tests/AppTests/PipelineTests.swift
+++ b/Tests/AppTests/PipelineTests.swift
@@ -18,7 +18,7 @@ class PipelineTests: AppTestCase {
             Package(url: "2".url, status: .ok, processingStage: .reconciliation),
             ].save(on: app.db).wait()
         // fast forward our clock by the deadtime interval
-        Current.date = { Date().advanced(by: Constants.reingestionDeadtime) }
+        Current.date = { Date().addingTimeInterval(Constants.reingestionDeadtime) }
         let batch = try Package.fetchCandidates(app.db, for: .ingestion, limit: 10).wait()
         XCTAssertEqual(batch.map(\.url), ["1", "2"])
     }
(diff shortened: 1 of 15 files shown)


Tests/AppTests/PipelineTests.swift
```diff
@@ -0,0 +1,208 @@
+@testable import App
+
+import SQLKit
+import Vapor
+import XCTest
+
```
finestructure, line 6: The tests in this file essentially encode our processing logic as it stands. It's probably easiest if you give these a read through if you agree with the decisions.

Just as a reminder, the stages are:

1. Reconciliation: fetching the master package list and reconciling it against the db (adding or deleting rows in the `packages` table)
1. Ingestion: fetch GH metadata for a set of package candidates returned by `Package.fetchCandidates(app.db, for: .ingestion, limit: 10) ` and create or update rows in `repositories`
1. Analysis: clone/pull repos for a set of package candidates returned by `Package.fetchCandidates(app.db, for: .analysis, limit: 10) ` and create rows in `versions` and `products` according to contents 

The final test, `test_processing_pipeline`, is the most important one: it encodes end-to-end how the whole processing pipeline works over time if the various stages get triggered repeatedly.

(I should probably add this text as a comment to the test itself... :D )

Tests/AppTests/PipelineTests.swift
```diff
@@ -0,0 +1,208 @@
...
+}
+
+
+private extension String {
+    var gh: String { "https://github.com/foo/\(self)" }
+}
+
+private extension Array where Element == String {
+    var gh: Self { map(\.gh) }
+}
```
finestructure, line 208: Just two little helpers to make the tests more readable by turning "1" into a url that can be processed without cluttering everything up with lots of `https://github.com/foo/1` ceremony.

Tests/AppTests/PipelineTests.swift
daveverwer: Added this to the README in #39. 👍

PR #36, opened 2020-05-08
Support beta versions

### LINKED ISSUE #30: Supporting beta versions of Packages
Looking through the differences between the existing dataset from swiftpm.co, and the new dataset from this project there are a few differences. One that's important is that swiftpm.co correctly parses semantic versions that have a `pre` part, more commonly known as beta versions 😂

For example, for one package on swiftpm.co:

Screenshot 2020-05-06 at 11 14 33@2x

and for the same package in this project:

Screenshot 2020-05-06 at 11 14 49@2x

finestructure:
Fixes #30 

Merge after #32


Sources/App/Commands/Analyzer.swift
```diff
@@ -220,7 +220,7 @@ func updateVersion(on database: Database, version: Version, manifest: Result<Man
     switch manifest {
         case .success(let manifest):
             version.packageName = manifest.name
-            version.swiftVersions = manifest.swiftLanguageVersions?.compactMap(SemVer.parse) ?? []
+            version.swiftVersions = manifest.swiftLanguageVersions ?? []
```
finestructure, line 223: Due to our stricter SemVer parsing I've had to change saving swift language versions as plain strings. They're not really SemVer - 5.0, 5.1 ... I don't think it makes a difference really as we're not using the actual version info anyway other than display (where 5.0 will read better than 5.0.0)

Sources/App/Models/SemVer.swift
```diff
@@ -5,43 +5,76 @@ struct SemVer: Content, Equatable {
...
+    }
 }
 
 
-extension SemVer: CustomStringConvertible {
-    var description: String { "\(major).\(minor).\(patch)" }
-}
+// Source: https://regex101.com/r/Ly7O1x/3/
+// Linked from https://semver.org
+let semVerRegex = NSRegularExpression(#"""
```
daveverwer, line 58: Did you say that this regex came from the SPM project itself? If so, let's reference where it came from in the comment.

finestructure: It’s linked from the semver page, at the bottom (i.e. the first link is the link from the semver.org page)

Sources/App/Commands/Analyzer.swift
daveverwer: For the Swift version, I agree! Apple don't refer to Swift 5.2.0, they refer to 5.2. I made the same decision in the SwiftPM Library implementation.

Just a note while we're on the subject. I think we should show a full, strict semantic version for *packages* everywhere, even if it's a .0 patch. Xcode *only* sees full semantic versions (including a patch) as valid, and we should do the same.

finestructure: Great, we're on the same page then!

Re SemVer for packages - that's actually a great point. In the PR after this one you'll see that I've collapsed branches and tags into `References`. Currently those are `.branch(String)` and `.tag(String)` but given that I'll change that to `.tag(SemVer)` then.

Does that make sense?

daveverwer: Yep, I saw the column change and it looks good, and it's even better if it's a `tag(SemVer)` 👍

PR #38, opened 2020-05-08
Html dsl

finestructure:
- Add JohnSundell/Plot HTML DSL
- Add snapshot testing lib for pages

Merge after #37


Sources/App/Pages/Home.swift
```diff
@@ -1,49 +1,37 @@
...
-        ]),
-        row([
-            .label(["Reconcile the master package list with the Package Index."]),
-        ]),
-    ])
+func reconcileButton() -> Node<HTML.BodyContext> {
+    .form(
+        .action("/packages/run/reconcile"),
+        .method(.get),
+        .input(.class("btn btn-primary"), .type(.submit), .value("Reconcile")),
```
finestructure, line 24: This has mostly translated well, except  I couldn't manage for the life of me to nest the entities like before: `form` -> `div class=row` -> `input`. As you can see in the snapshot image that messes up the spacing.

Obviously the looks of this initial page are completely irrelevant and I'm not even sure if my original html is the right way to do it. Just wanted to mention it in case that might be a problem further down the line.

Also, it shows how the snapshots can be useful to see accidental changes.

It might make sense to skip them though until the the design is somewhat stable. I'm adding some notes below how to manage them.

Tests/AppTests/WebpageSnapshotTests.swift
```diff
@@ -0,0 +1,20 @@
...
+
+import SnapshotTesting
+import XCTest
+
+
+let defaultSize = CGSize(width: 800, height: 600)
+
+class WebpageSnapshotTests: XCTestCase {
+    func test_home() throws {
+        // record = true
```
finestructure, line 11: Uncommenting this will re-record the snapshots, overwriting the files in `__Snapshots__` (see below) _and_ mark the test as failed (just so you don't accidentally and silently change the expectations without review).

Setting the test to record while tweaking the design is a nice way to check design changes against the committed baseline in a local diff viewer.

Sources/App/routes.swift
```diff
@@ -3,7 +3,7 @@ import Vapor
 
 func routes(_ app: Application) throws {
     app.get { req in
-        render(page: homePage())
+        homePage()
```
finestructure, line 6: The `ResponseEncodable` extension on `HTML` above allows us to return `HTML` straight from the route handler and this is the entry point for `/`.

Sources/App/Pages/Home.swift
daveverwer: I've just spent some time looking at this and it's certainly not happy with it. I can get a `div` and other elements inside a form, but as soon as I put an input inside that div, I get:

```
Referencing static method 'input' on 'Node' requires the types 'HTML.BodyContext' and 'HTML.FormContext' be equivalent
```

My guess is that there's an allowed list of tags, which isn't actually correct as you can pretty much put anything inside a `form`.

This is very much a common problem with HTML DSLs in my experience! Programmers want to put logic into them that doesn't exist in HTML itself. I may be wrong, and I haven't investigated properly yet, but that's my current theory.

finestructure: Yeah, exactly. That’s the error I got and I couldn’t find a way to make the types lines up. The contexts just don’t support it - it’s the same as with swift-html, just a bit further in... meh.

Not sure if this is something we need to worry about. I'm guessing the layout can be fixed to look the same even without that nesting but the difficulties may be harbingers of things to come?

_I_ would be stuck here and revert to Leaf or the previous "non-typed" DSL but that's mainly because nesting `div`s is my hammer for HTML layout 🤣

Tests/AppTests/__Snapshots__/WebpageSnapshotTests/test_home.1.txt
```diff
@@ -0,0 +1,35 @@
...
+        <div class="row">
+          <input class="btn btn-primary" type="submit" value="Ingest"/>
+        </div>
+        <div class="row">
+          <label>Ingest metadata for a batch of packages.</label>
+        </div>
+      </form>
+    </div>
+  </body>
+</html>
```
finestructure, line 35: Nice - the html snapshot only shows cosmetics differences now (the baseline commit is 63eee1325bb54d2879e86253a87370597fd59fe2) and the image is identical against that revision rendered with the previous dsl. All good now!

PR #39, opened 2020-05-08
README and Makefile changes

daveverwer:


Makefile
```diff
@@ -20,11 +20,11 @@ test-docker:
 	@# run tests inside a docker container
 	docker run --rm -v $(PWD):/host -w /host --network="host" vapor/swift:5.2-bionic make test
 
-test-e2e: reset reconcile ingest
+test-e2e: db-reset reconcile ingest analyze
 	@# run import sequence test
```
daveverwer, line 24: Is there a reason analyze wasn't part of this? My guess is just an oversight as it was added?

Makefile
```diff
@@ -57,6 +57,4 @@ db-down-dev:
 db-down-test:
 	docker rm -f spi_test
 
-reset: db-down-dev db-up-dev migrate
-	@# reset dev db (test db is automatically reset by tests)
-
+db-reset: db-down db-up migrate
```
daveverwer, line 60: The README said `db-reset` but the Makefile said `reset`, so I picked `db-reset` 👍

daveverwer: Also, when I ran this locally it consistently hung at the `migrate` stage if *only* the dev db had been reset. Resetting both made it work, and there should be no downside to it.

Makefile
```diff
@@ -20,11 +20,11 @@ test-docker:
 	@# run tests inside a docker container
 	docker run --rm -v $(PWD):/host -w /host --network="host" vapor/swift:5.2-bionic make test
 
-test-e2e: reset reconcile ingest
+test-e2e: db-reset reconcile ingest analyze
 	@# run import sequence test
 
 migrate:
-	$(VAPOR) run migrate
+	echo y | $(VAPOR) run migrate
```
daveverwer, line 27: There may be a reason we wouldn't want to do this, but I can't think of one. Can you?

Makefile
finestructure: Ah yes, I used this target a lot early on but by the time I started testing analyse I really only needed to run that in isolation - good to have it in!

Makefile
finestructure: I think that may be because the db is not quite ready and the migrate kicks in too soon. There is a way to wait on it but it’s probably not worth fiddling that in tbh.

PR #43, opened 2020-05-10
Todo cleanup

finestructure:
Fixing some TODOs, clarifying others

### DIFF
--- a/Sources/App/Models/Package.swift
+++ b/Sources/App/Models/Package.swift
@@ -42,7 +42,7 @@ final class Package: Model, Content {
     @OptionalEnum(key: "processing_stage")
     var processingStage: ProcessingStage?
 
-    @Field(key: "last_commit_at")  // TODO: shouldn't this rather live in Repository?
+    @Field(key: "last_commit_at")  // TODO: shouldn't this rather live in Repository? Is it needed at all?
     var lastCommitAt: Date?
 
     @Children(for: \.$package)
--- a/Sources/App/Models/Repository.swift
+++ b/Sources/App/Models/Repository.swift
@@ -2,8 +2,6 @@ import Fluent
 import Vapor
 
 
-// TODO: sas 2020-04-26: discuss whether we want to inline Repository with Package rather than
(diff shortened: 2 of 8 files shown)


Sources/App/Models/Package.swift
```diff
@@ -42,7 +42,7 @@ final class Package: Model, Content {
     @OptionalEnum(key: "processing_stage")
     var processingStage: ProcessingStage?
 
-    @Field(key: "last_commit_at")  // TODO: shouldn't this rather live in Repository?
+    @Field(key: "last_commit_at")  // TODO: shouldn't this rather live in Repository? Is it needed at all?
     var lastCommitAt: Date?
```
finestructure, line 46: Do we need this field at all? Or was this something the previous version used for processing stage management, which we're doing via `processing_stage` now?

Sources/App/Models/Repository.swift
```diff
@@ -36,7 +34,7 @@ final class Repository: Model, Content {
     @Field(key: "forks")
     var forks: Int?
 
-    @OptionalParent(key: "forked_from_id")
+    @OptionalParent(key: "forked_from_id")  // TODO: remove or implement
     var forkedFrom: Repository?
```
finestructure, line 38: Do we need this for display? Or would be enough to record that a repo _is_ a fork, without necessarily linking to the original? It would save us a lookup across `reposities` when populating the db and "raceyness" on first ingestion when the fork comes before the parent.

Tests/AppTests/AnalyzerTests.swift
```diff
@@ -70,7 +70,7 @@ class AnalyzerTests: AppTestCase {
         assert(commands: commands, expectations: expecations)
 
         // validate versions
-        // TODO: This is monstrous... create a helper? There has to be a better way?
+        // A bit awkward... create a helper? There has to be a better way?
         let pkg1 = try Package.query(on: app.db).filter(by: urls[0].url).with(\.$versions).first().wait()!
```
daveverwer, line 74: "This is monstrous" -> "A bit awkward" with no code changes...

The next commit on this line is going to change it to "This is awesome!" 😂

PR #45, opened 2020-05-10
Fixes #42 - supported platforms enum

### LINKED ISSUE #42: Save supported platforms as enum?
https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/blob/todo-cleanup/Sources/App/Models/Version.swift#L34-L35

Discussed here: https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/pull/24#discussion_r419964032

finestructure:


Sources/App/Commands/Analyzer.swift
```diff
@@ -224,7 +224,7 @@ func updateVersion(on database: Database, version: Version, manifest: Result<Man
         case .success(let manifest):
             version.packageName = manifest.name
             version.swiftVersions = manifest.swiftLanguageVersions ?? []
-            version.supportedPlatforms = manifest.platforms?.map { $0.description } ?? []
+            version.supportedPlatforms = manifest.platforms?.compactMap { Platform(from: $0) } ?? []
```
finestructure, line 227: There is a slight potential issue here in case `Manifest.Platform`'s and `Platform`s list of platforms diverge. If the manifest had a new key but the db entity didn't add it as well, this would silently discard the platform.

The test `ManifestTests.test_platform_list` below ensures this cannot happen.

daveverwer: Adding the test is good, but this brings up an approach which might be worth discussing. I'm generally a fan of crashing if something unexpected happens. I know it's not going to happen every day, but if/when Apple add a new platform we'll just silently eat those errors because of this?

Is there a way we can instead start things crashing when that happens? With a service like Rollbar, those notifications don't get overwhelming, and it means that it's hard to ignore what should be a fairly urgent problem.

daveverwer: The test is good, but because it's based on fixture data it won't ever fail because of external factors.

finestructure: I definitely want to roll up all these errors somewhere into a logging and reporting system!

I'll add a card to add a test catching any platform changes, ok? I think it'd already bubble up into the error bucket but it's best to make sure.

I'll also add a card for an error reporting mechanism.

PR #55, opened 2020-05-11
License enum

finestructure:
- makes `License` an enum
- map unknown licenses to `.other`
- I've looked at the distinct values present in both the new and the old server, should be the complete list currently seen

### DIFF
--- a/Sources/App/Models/License.swift
+++ b/Sources/App/Models/License.swift
@@ -1,0 +1,30 @@
+
+enum License: String, Codable {
+    case agpl_3_0 = "agpl-3.0"
+    case apache_2_0 = "apache-2.0"
+    case bsd_2_clause = "bsd-2-clause"
+    case bsd_3_clause = "bsd-3-clause"
+    case cc0_1_0 = "cc0-1.0"
+    case epl_1_0 = "epl-1.0"
+    case gpl_3_0 = "gpl-3.0"
+    case isc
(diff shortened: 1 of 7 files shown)


Sources/App/Models/License.swift
```diff
@@ -0,0 +1,30 @@
...
+    case cc0_1_0 = "cc0-1.0"
+    case epl_1_0 = "epl-1.0"
+    case gpl_3_0 = "gpl-3.0"
+    case isc
+    case lgpl_2_1 = "lgpl-2.1"
+    case lgpl_3_0 = "lgpl-3.0"
+    case mit
+    case mpl_2_0 = "mpl-2.0"
+    case other
+    case unlicense  // TODO: is this an actual license name or meant to say "unlicensed"?
```
finestructure, line 16: Is this an actual license?

daveverwer: This is an actual license, and what I wanted to check was whether any packages in the current SwiftPM Library database use it, and they do!

Screenshot 2020-05-11 at 21 48 34@2x

https://choosealicense.com/licenses/unlicense/

daveverwer:
Happy for this to be merged as soon as that TODO is removed.

Sources/App/Models/License.swift
finestructure: Ah cool, thanks for checking!

PR #59, opened 2020-05-12
Added methods needed for license display to License.

daveverwer:
I think these fit in the model, rather than having another class, but happy to discuss it!


finestructure:
I agree, that's exactly where I think they should be!

PR #61, opened 2020-05-13
Deployment

### LINKED ISSUE #56: make checkouts directory configurable

### LINKED ISSUE #57: Deployment automation
I've started having a look at out deployment options.

What I know works without any further research is the following:

- mirror repo to Gitlab
- instrument Linode box as a runner
- drive deployment to it via `.gitlab-ci.yml`
- run docker-compose up on the runner (the Linode box)

I'm doing this for several projects and it works well. The downside is that deployment runs separate-ish. While `.gitlab-ci.yml` could happily live inside the repo, what it triggers happens elsewhere. (Although arguably it has to happen elsewhere by definition - the Linode box - it's just a question of how much of it does.)

For Github actions I believe we can go the following:

- build docker image as we do now
- push to docker hub
- set up ssh action that runs on Linode box
- pull docker image
- run docker-compose up

The slight issue I'm seeing right now is that there's no built in ssh action (AFAICT) and I'm not excited to use a 3rd party one that handles private keys as an input. However, I think it shouldn't be too hard to either fork one and use a version we control or just run the key handling inline.

Or maybe I'm to sensitive about the key? B/c unless we host our own runner (which we easily could, I'm doing it for three other GH projects) we'll always be passing our private key to some machine we don't control.

finestructure:
- fixes #56 - checkouts directory configurable via CHECKOUTS_DIR env variable (defaults to previous value)
- set ingestion dead time to 60mins
- /api/version endpoint
- fixes #57 - automatic deployment via Gitlab pipelines
	- manual for any revision
	- fully automatic for tags

This latest version is deployed: http://173.255.229.82:8080


Sources/App/routes.swift
```diff
@@ -7,6 +7,8 @@ func routes(_ app: Application) throws {
     }
 
     app.group("api") { api in
+        api.get("version") { req in API.Version(version: appVersion) }
```
finestructure, line 10: I typically add an endpoint that reports the deployed version back for easier confirmation - this will return a JSON response `{"version": <sha or tag>}`

daveverwer: That's really interesting. Is there something you use that uses this?

docker-compose.commands.yml
```diff
@@ -0,0 +1,49 @@
...
+    # set these variables via the environment or a `.env` file, which
+    # docker-compose reads and uses to populate variables
+    CHECKOUTS_DIR: ${CHECKOUTS_DIR}
+    DATABASE_HOST: ${DATABASE_HOST}
+    DATABASE_PORT: ${DATABASE_PORT}
+    DATABASE_NAME: ${DATABASE_NAME}
+    DATABASE_USERNAME: ${DATABASE_USERNAME}
+    DATABASE_PASSWORD: ${DATABASE_PASSWORD}
+    GITHUB_TOKEN: ${GITHUB_TOKEN}
+    LOG_LEVEL: ${LOG_LEVEL}
```
finestructure, line 20: This file contains single-run commands

Issue #62, opened 2020-05-14
Review Ingestor/Analyzer error paths


finestructure:
I need to rework both stages around their error handling. There are still gaps where errors are not caught, for instance this test fails:

```
    func test_invalidPackageCachePath() throws {
        // setup
        try savePackages(on: app.db, ["1", "2"], processingStage: .ingestion)

        // MUT
        try analyze(application: app, limit: 10).wait()

        // validation
        let packages = try Package.query(on: app.db).sort(\.$url).all().wait()
        XCTAssertEqual(packages.map(\.status), [.invalidUrl, .invalidUrl])
    }
```

by raising an `invalidPackageCachePath` exception out of the `analyze` call that should not bubble up.

In practise this might not happen, because an invalid url like that doesn't get this far in the chain. However, both stages, suffer from being written in the first few days and contain a mix of throwing and `Result` code which is hard to reason about.

Explore rewriting everthing around `Result`, if possible.

PR #65, opened 2020-05-14
The beginnings of a public package page.

daveverwer:
I'm submitting this for merging now as I'm not going to be able to work on this tomorrow (newsletter!) and it's already getting big so I think it's best to get it in to avoid messy merges.

The basic structure of the design is done, but much of the detail (for example all the icons) are not yet complete. It's coming along really well though.

It's far from finished, and the most obvious thing that's missing is a structure to the code in `PackageShowView`. I know how I want to split that up, it's just not done yet and I'm *finished* for today.

Code review is very welcome with the exception of what I mentioned above.

In terms of how it looks... Here it is:

Screenshot 2020-05-14 at 22 04 30@2x

and for mobile:

Screenshot 2020-05-14 at 22 04 55@2x


Sources/App/Controllers/PackageController.swift
```diff
@@ -0,0 +1,18 @@
+import Fluent
+import Plot
+import Vapor
+
+struct PackageController {
+
+    func index(req: Request) throws -> EventLoopFuture<Response> {
+        // No such thing as a full index of packages.
+        return req.eventLoop.future(req.redirect(to: "/"))
```
finestructure, line 9: Oh nice, I had no idea there's an `eventLoop.future` :D

Sources/App/Views/PackageController/PackageShowView.swift
```diff
@@ -0,0 +1,118 @@
+import Plot
+
+
+class PackageShowView: PublicPage {
+
+    let package: Package
+
+    init(_ package: Package) {
+        self.package = package
```
finestructure, line 9: I guess here's where we interface with our view model. After tackling #62 - depending on how that goes 😅 - shall I start filling in the query and plug it in here?

finestructure:
This looks great, can't wait to hook this up with data and start making the content browsable!

Sources/App/Controllers/PackageController.swift
daveverwer: Well, when you've been doing Vapor development as long as I have, you learn a few things... 😂 Orrrrr... Google is *awesome* 🤷‍♂️

Actually I learned something cool about redirects when figuring this out. My initial feeling was just to make this a 302, but a 303 enforces the redirect to be a GET regardless of the method used for the initial request. It's not actually changing anything here as `index` is always a GET in this case, but it's good to know.

Sources/App/Views/PackageController/PackageShowView.swift
daveverwer: Yes, that'd be wonderful. 👍

My gut feeling says that the view model (for example) shouldn't construct the full sentences that are going to be used, but instead populate everything that might be needed. 

Maybe something like this:

To construct this sentence:

"In development for over 5 years, with 1,433 commits and 79 releases."

The view model could contain:

* First commit date (`Date`)
* Number of commits (`Int`)
* Number of releases (`Int`)

and to construct this sentence:

"By Christian Noon, Mattt, Jon Shier, Kevin Harwood, and 186 other contributors."

* Array of:
  * Full Name (`String`)
  * Profile URL (`URL`)
* Number of other contributors (`Int`)

Is that what you were thinking?

daveverwer: Actually, the place for this discussion isn't a merged PR. I'm going to make a new issue for this so we can discuss it and track it.

PR #67, opened 2020-05-15
Issue 62 improve error handling

finestructure:
This reworks the analysis part of #62 

I still needs to go through the Ingestor and there are some other things I need to clean up but I wanted to push this out before it gets even bigger.

This main cleans up `Analyzer`, avoid a mix of throwing, futures, and Results in favour of just futures and Results. It also adds more tests to the individual stages and makes the main `analyze` function (line 49++) more organised and cleaner. I hope it's more readable now!

I've also added an issue to go in and add some doc strings (#68 ) and a follow up around testing (#69 )

Just maybe a couple of lines to the point of this exercise: it's very easy to accidentally return early from an array of futures and there are a few ways this can happen:

- calling `flatten` on an array of futures
- throwing exceptions into that flatten or any other "gatherer" that calles `.andAllSuceed` or one of the family

Esp the throwing is easily done accidentally by forgetting to wrap an error into a future instead of throwing. What this does then is collapse arrays of futures into returning as soon as the first future fails (short circuiting).

So the exercise here was to remove or encapsulate throws and make sure we use `whenAllComplete` and `andAllComplete`. This also allowed me to reduce some of the nesting inside the main `analyze` function.

I hope all this makes some sense, happy to take a look at it together!

### DIFF
--- a/Sources/App/Commands/Analyzer.swift
+++ b/Sources/App/Commands/Analyzer.swift
@@ -64,50 +64,18 @@ func analyze(application: Application, packages: EventLoopFuture<[Package]>) -> 
         }
     }
 
-    // refresh checkouts
-    let checkouts = packages
-        // FIXME: flatMapEach calls flatten, i.e. andAllSucceed, i.e. it short-circuits
-        .flatMapEach(on: application.eventLoopGroup.next()) {
-            pullOrClone(application: application, package: $0)
-    }
+    let checkouts = packages.flatMap { pullOrClone(application: application, packages: $0) }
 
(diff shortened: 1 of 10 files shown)


daveverwer:
Love the idea of moving this towards better error handling, and agree with the concept. I'm afraid I'm not following every detail of the code in a PR like this, but let's get it merged and maybe go through it a bit on our next call.

finestructure:
Great! Thanks for taking a look and let's catch up on Monday or so, ok? Have a nice rest of the weekend! :)

PR #75, opened 2020-05-17
First view model

finestructure:
I've hooked up the package page with a proposal for a view model.

It's currently using the same stub data but supplied from a function that populates that model. I'll look into actually assembling that from data next but wanted to push this up so you can have a look asap.

I've used our snapshots to ensure everything's unchanged in the page rendering, which worked really well except for the fact that css isn't being rendered with `loadHTMLString`. Not super important but may you've got an idea what's going on!


Tests/AppTests/WebpageSnapshotTests.swift
```diff
@@ -1,20 +1,52 @@
...
+        assertSnapshot(matching: page.render(indentedBy: .spaces(2)), as: .lines)
+
+        #if os(macOS)
+        if !isRunningInCI {
+            assertSnapshot(matching: page, as: .image(size: defaultSize, baseURL: baseURL()))
+        }
+        #endif
+    }
+
+    func test_PackageShowView() throws {
```
finestructure, line 28: I've added this snapshot mainly for my own testing. It does pass but might be annoying for you while working on the page so maybe you'd prefer if we gate this with an `XCTSkip` on CI until it's settled down a bit?

daveverwer: Yea I deliberately didn't include a snapshot with my work so far as it was just going to break with every commit at the moment. It's fine to include it from now on as that HTML is largely done now. It's mainly the CSS that needs work now.

Sources/App/Views/PackageController/PackageShowView.swift
```diff
@@ -191,3 +140,107 @@ class PackageShowView: PublicPage {
...
+            let lastPullRequestClosedMerged: String
+        }
+    }
+}
+
+
+extension PackageShowView.Model {
+    func authorsClause() -> [Node<HTML.BodyContext>] {
+        switch authors.count {
+            case 0:
```
daveverwer, line 179: I was planning to pull this functionality out into a helper class. There's a lot of things on this page that need making into a comma-separated sentence. The logic is good though 👍

Sources/App/Views/Utils.swift
```diff
@@ -0,0 +1,12 @@
+
+func pluralize(count: Int, singular: String, plural: String? = nil) -> String {
```
daveverwer, line 2: Missing a test for this?

finestructure: Yes, I rushed to push when I saw you're active :)

finestructure: I'll add some real quick

Sources/App/Views/PackageController/PackageShowView.swift
finestructure: Yes, please don't hesitate to move things around. I've just picked a couple of file and type names as they seemed to make sense right now but I image we'll need to reorganise that a bit as it grows.

daveverwer:
> I've used our snapshots to ensure everything's unchanged in the page rendering, which worked really well except for the fact that css isn't being rendered with loadHTMLString. Not super important but may you've got an idea what's going on!

I just realised I forgot to answer this. In some ways is this an advantage? I'm not sure I want to test the CSS, it's really not that important that it's pixel perfect and any changes in markup should be enough to trigger test failures.

finestructure:
We've got a separate, markup only test here:

```
assertSnapshot(matching: page.render(indentedBy: .spaces(2)), as: .lines)
```

This dumps the html [in a file](https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/blob/08493992afa2c42131191a6e368d960d18b18500/Tests/AppTests/__Snapshots__/WebpageSnapshotTests/test_PackageShowView.1.txt).

The other test - if it worked properly - could take care of the visual rendering so you can spot things that are broken in the css. It's not a huge deal but at the moment neither test captures issues with the css.

daveverwer:
I'm happy either way, but I'd say an image-based test may be overkill. Testing the markup might be enough.

PR #77, opened 2020-05-18
Package show model query

finestructure:
Sorry for the commit noise but as I was adding the mock for the package query I noticed that a lot of other mocks were in the App target instead of the test target and I moved them over as part of this change.

The relevant bit is really just the `query` method where I commented and the new test `test_query_basic`.


Sources/App/Views/ViewModelQueries/PackageShowViewQuery.swift
```diff
@@ -1,32 +1,77 @@
...
+                          authors: [],
+                          history: .init(since: "foo",
+                                         commits: .init(name: "commits", url: "1"),
+                                         releases: .init(name: "releases", url: "2")),
+                          activity: .init(
+                            openIssues: .init(name: "issues", url: "3"),
+                            pullRequests: .init(name: "PRs", url: "4"),
+                            lastPullRequestClosedMerged: "XX days ago"),
+                          products: (libraries: 17, executables: 42))
+        }
```
finestructure, line 63: Dave, I'm thinking of making some of the model fields optional to allow us to chip away at the query while not everything is there yet and to better handle error cases where data is partially missing. (See also the TODO comment just above this initialiser).

So for instance if we have

```
struct Model {
  ...
  let history: History?
  ...
}
```

I'd leave this `nil` in the query and then in the display part we could conditionally leave out the whole section if data is lacking.

Does that make sense?

The alternative would be to leave the fields out until we can populate them but then we'd have to comment out the display part.

daveverwer: Yes, it's better for the views to be able to deal with optional data. That way, if we can't get data for a specific package for any reason the page can still be displayed, just without that element.

finestructure: Excellent. Later on we can look at maybe returning `nil` from the model query if too many things are missing but that's not something we need to deal with right now.

Sources/App/Views/ViewModelQueries/PackageShowViewModel.swift
```diff
@@ -0,0 +1,49 @@
...
+                // If for instance we have just a package entry with no repository, we should
+                // probably error out (because too many fields would be blank).
+                Self.init(title: p.name ?? "–",
+                          url: p.url,
+                          license: p.repository?.license ?? .none,
+                          summary: p.repository?.summary ?? "–",
+                          authors: [],
+                          history: nil,
+                          activity: nil,
+                          products: p.productCounts)
```
finestructure, line 26: I'm a bit on the fence whether to make `title` and `summary` non-optional and avoid the `?? "–"` here. I guess it's better to do it here than having to sprinkle it throughout the view code.

On the other hand, we should probably consider some values as technically optional, because of the way we populate the db but effectively implicitly unwrapped because they need to be there.

Should we make this  this initialiser failable and return `EventLoopFuture<Self?>` such that the caller can bail out? I.e. in `PackageController`:

```
    func show(req: Request) throws -> EventLoopFuture<HTML> {
        guard let id = req.parameters.get("id").flatMap(UUID.init(uuidString:)) else {
            return req.eventLoop.future(error: Abort(.notFound))
        }
        return PackageShowView.Model.query(database: req.db, packageId: id)
            .unwrap(or: Abort(.notFound)  // <- bail out if model query fails or returns insufficient data
            .map { PackageShowView($0).document() }
    }
```

This should be rate but not impossible, for instance if a package is added but then not immediately processed, we'll be missing things title and summary.

When we tackle search we should ensure we filter those results out but since we won't know that here, this might be a workable approach.

daveverwer: We need to cope with a missing summary. The GitHub description field where this comes from is not required. There are plenty of packages that have no description.

Having a missing name is much trickier. We should always have a name after an analysis run has taken place, so maybe the package page returns a 404 (and is excluded from search results) until an analysis has processed. I think we can use the existing status field for this?

finestructure: Ok, I'll make the initialiser failable and raise a 404 for a missing package name (title) then, ok? (as part of this PR)

Regarding search, that's what I was thinking, too. Just like we have candidates for analysis we'll have candidates for search and freshly added ones just wouldn't be eligible. And mechanism to steer that is the status and processing stage field. (It's going to be slightly more complicated that that, because we bounce between stages when we re-ingest packages but we can overcome that by introducing another stage or so).

I've got this sort of mapped out but that's probably not something I'd want to mix into this PR here.

daveverwer: Sounds good 👍

Sources/App/Views/ViewModelQueries/PackageShowViewQuery.swift
finestructure: Ok, that's updated and good to go!

daveverwer:
Is this ready to go?

finestructure:
Yes, it is!

Issue #47, opened 2020-05-10
Site design


daveverwer:
I've done a lot of thinking about how we want to show package metadata this weekend, and I'm posting my initial draft designs here for discussion. Even though I'm planning to start implementing some parts of this tomorrow, everything is still up for open and frank review in this thread.

When I started this process, my main goal was to move away from the approach of trying to display all metadata about multiple versions of a package **inline** in the search results. It's already too much to take in and also doesn't show anywhere near what people need to know to make informed decisions about package quality.

The idea, as discussed previously, was to move to a "page per package" approach giving much more scope for displaying information that people can use to judge the maturity, quality, and maintenance history of a package.

I also rolled back my thinking displaying metadata from multiple package versions as part of this process. The fact that we capture all metadata from multiple versions it critical, but it's only occasionally useful for people to see. When it's useful, it's instrumental, but that tends to only be around the release of new Swift/operating system versions. The fact that we have this data is a huge advantage, but I want to be smart about displaying it.

So, here's my first draft of the per-package page, where I spent most of my time and thinking this weekend.

![Package Page@2x](https://user-images.githubusercontent.com/5180/81510077-27341400-9307-11ea-8fa1-c5fcce994545.png)

Some notes:

## Display of multiple version metadata

You'll see an example of my idea for dealing with changing metadata between significant versions in the Languages and Platforms section. Most packages will use the same version of Swift/iOS/macOS/etc. for every version. It doesn't change *that* often. So let's only display it when it changes!

If all three significant versions use the same Swift versions, it just gets displayed once. Only if (in this example) the new beta upgraded Swift version would it be shown like it is here.

## Prominence (or lack of) of the Swift version

You'll notice that one of the (seemingly) most essential bits of metadata, the version of Swift that's supported, is quite far down the listing. At first, this worried me too, but I think I'm OK with it after playing with it today. I think if we can get the metadata from git/GitHub about the number of releases, project start date, commits, etc. that gives an excellent indication of how active a project is. Active projects tend to support the latest version of Swift.

The Swift version needs to be there, and it is, but I don't think it's as important as that other metadata.

## Contributors

Here's a basic plan that I think will work well for this line of metadata.

* Any contributor with more than 10 commits is eligible to be listed here. This rule will save projects where someone fixes a typo in a README in a newly released project with just an "Initial commit."
* For all eligible contributors, they are listed if they have made more than 5% of the total commits.
[...]

PR #80, opened 2020-05-18
Redrawn icons!

daveverwer:
Redrawn all from scratch (no stock artwork) so that we can have a consistent line width. I also went with a filled style rather than outline as I thought it'd work better.

Happy to make changes though! Feedback welcome.

Before:

Before

After:

After


daveverwer:
Personal views on how the redrawn ones came out. I'm happy with the top four icons. The Apple and Swift ones are much less readable, but they were too inconsistent with their weightings without being surrounded by something.

The branch/stable/beta icons will look better when they're in-place in the lozenges.

I might try making the Apple and Swift icons bigger. Will post a screenshot.

daveverwer:
These are too big, they stand out far left of the titles, which is bad 😬 but these two specific icons look better at this size.

Gut feeling says that the'll be fine at the smaller size.

Screenshot 2020-05-18 at 18 19 07@2x

daveverwer:
This is about as good as I can get it in terms of a trade-off between the size of those bottom two icons not extending to the left of the titles, and legibility.

Screenshot 2020-05-18 at 18 26 44@2x

finestructure:
Yes, I think the top four look better this way as well. However, I feel now the release icons lost a bit of weight. I _did_ like the circle style. I wonder if it’d work if we made all of them circle styles? So regardless of their actual density they’d still line up well enough.

Although I wonder if some of the symbols would force the circles to be too big.

But looking at the variants, personally I’d prefer the new top four and then the original releases and platform icons.

daveverwer:
> I feel now the release icons lost a bit of weight. I did like the circle style.

You're absolutely right, that was intentional as they're about to be put inside the lozenges. I just haven't put them inside that markup yet. They'll be [like they are here](https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/issues/47#issuecomment-630180904).

> But looking at the variants, personally I’d prefer the new top four and then the original releases and platform icons.

I have one more idea for the language and platform icons, I think a rounded rect might just give them a little something over a plain rect. I'll give it one more try.

daveverwer:
Pushed this version:

Screenshot 2020-05-18 at 22 31 56@2x

finestructure:
Ah of course, they go in the lozenges!

And rounded rect for the platforms - that's 👌

PR #92, opened 2020-05-19
Add Repository.owner/name fields from metadata

### LINKED ISSUE #81: Parse out owner and repository name
Store these in two fields:

* username
* reponame

finestructure:
Fixes #81


Sources/App/Core/Github.swift
```diff
@@ -15,8 +15,14 @@ enum Github {
...
         var license: License?
-        var stargazersCount: Int
+        var name: String?
+        var owner: Owner?
         var parent: Parent?
+        var stargazersCount: Int
+
+        struct Owner: Decodable, Equatable {
+            var login: String?
+        }
```
finestructure, line 25: It'd probably be safe to parse the owner/name out of the existing `fullName` field but since it's easy to just get the original fields I've added those instead.

Sources/App/Migrations/CreateRepository.swift
```diff
@@ -25,3 +25,21 @@ struct CreateRepository: Migration {
...
+            .update()
+    }
+
+    func revert(on database: Database) -> EventLoopFuture<Void> {
+        return database.schema("repositories")
+            .deleteField("name")
+            .deleteField("owner")
+            .update()
+    }
+}
```
finestructure, line 45: I've added the new fields as a migration for now. It's probably best to squash all these into a single schema and reprocess everything but maybe for now it's a good exercise to deploy with a schema migration.

Sources/App/Core/Github.swift
daveverwer: Gotcha, looking at the GitHub API docs login seems a strange name for it but it matches the API response. I agree since we've already got it coming down, it's better to read it from the API response than parse it.

Sources/App/Migrations/CreateRepository.swift
daveverwer: Either way is fine by me until we get to a public launch. Agree it's not a bad idea to test the migration process though.

PR #102, opened 2020-05-21
Release and platforms clauses

finestructure:
This updates the view model with the releases and L&P fields. The snapshot is now generated entirely from the mock data provided via `PackageShowView.Model.mock`.

I don't believe we had an issue associated with this, did we?


Sources/App/Views/PackageController/PackageShowView.swift
```diff
@@ -198,14 +95,21 @@ extension PackageShowView {
...
         }
 
+        struct DatedLink: Equatable {
+            let date: String   // FIXME: use RelativeDateTimeFormatter
+            let link: Link
+        }
+
         struct History: Equatable {
-            let since: String  // TODO: use Date and derive on the fly
+            let since: String  // FIXME: use RelativeDateTimeFormatter
```
finestructure, line 112: I'll convert these to `Date` and tackle the relative dates when I actually populate them from data.

Tests/AppTests/__Snapshots__/WebpageSnapshotTests/test_PackageShowView.1.txt
```diff
@@ -121,7 +121,7 @@
               </ul>
             </li>
             <li>
-              <p>Version 
+              <p>Versions
```
finestructure, line 124: This is a genuine fix in the markup, we were listing this as `Version X and Y` before

Sources/App/Views/Utils.swift
```diff
@@ -10,3 +10,18 @@ func pluralize(count: Int, singular: String, plural: String? = nil) -> String {
             return "\(count) \(plural)"
     }
 }
+
+
+extension String {
+    func pluralized(for count: Int, plural: String? = nil) -> String {
+        let plural = plural ?? self + "s"
+        switch count {
+            case 0:
```
daveverwer, line 19: It's only when you see lines of code like this that you're reminded how ridiculous English is 😂

Sources/App/Views/PackageController/PackageShowView.swift
```diff
@@ -288,13 +193,186 @@ extension PackageShowView.Model {
...
+extension PackageShowView.Model {
+
+    static func listPhrase(opening: Node<HTML.BodyContext>,
+                           nodes: [Node<HTML.BodyContext>],
+                           ifNoValues: [Node<HTML.BodyContext>] = []) -> [Node<HTML.BodyContext>] {
+        switch nodes.count {
+            case 0:
+                return ifNoValues
+            case 1:
+                return [opening, nodes[0]]
```
daveverwer, line 347: Subtle one here, and one that complicates the solution a bit. There shouldn't be a comma for just two items.

* One item.
* One item and another item.
* One item, another item, and a final item.

finestructure: That's how it shoud work. Only the default case adds any commas (in the `middle` section).

I don't have an explicit test for this because the nodes are hard to compare but the snapshot should confirm it - or is this not covering those cases?

Tests/AppTests/__Snapshots__/WebpageSnapshotTests/test_PackageShowView.1.txt
finestructure: Actually, this is generated from `listPhrase` so confirming there's no comma I believe. (see comment below)

Sources/App/Views/Utils.swift
finestructure: Let me tell you about German. I'm glad we're not localising this 🤣🤣

Sources/App/Views/PackageController/PackageShowView.swift
daveverwer: I'm sure you're right, I was just eyeballing the code and I didn't see a difference with the commas in the code. Apologies!

PR #106, opened 2020-05-21
Search results

### LINKED ISSUE #48: Package Search
First up, how does search currently work in the SwiftPM Library?

* Information is donated from the various package/product/version records into a local ElasticSearch instance. The default branch is always the source search information as that's the most up to date.
* The following information is donated:
  * Package name (extracted from the Package.swift)
  * Package description (currently from the GitHub description)
  * The repository identifier (username/project, excluding github.com and other URL paraphernalia)
  * The search score (for ordering only, it's not searched).
* Search can be customised to look only in certain fields. For example, try [this search](https://swiftpm.co/?query=name%3Ahtml) compared to [this search](https://swiftpm.co/?query=description%3Ahtml), or [this search](https://swiftpm.co/?query=html).

The plan was that at some point you might also want to customise search by other useful metadata like, for example:

* Swift version
* Platform compatibility
* Authors real names, as well as the "free" search we get from including the repository identifier which includes a GitHub username.
* More... 

Without ElasticSearch, some of this becomes much harder (but not impossible) to achieve. As I was thinking about the design of the search form today, something struck me. Even though the colon syntax for searching is fairly standard, not everyone knows about it. So maybe we lean into a basic Postgres full-text search across a de-normalised version of the search data from the default branch, processed only to contain searchable data. Then, some checkboxes/input controls people can tick as they start to search. There's plenty of UI space on that home page for extra controls, and it might aid discovery of the feature.

Not all of this needs implementing straight away, but I want to discuss it about it as it affects the design.

daveverwer:
Closes #48.


Public/javascripts/main.js
```diff
@@ -13,4 +12,83 @@ document.addEventListener('DOMContentLoaded', function(event) {
...
+    resultsElement.appendChild(noResultsElement)
+  } else {
+    // Create an unordered list with the results.
+    const resultsListElement = document.createElement('ul')
+    searchResults.forEach((result, index) => {
+      const resultsItemElement = document.createElement('li')
+
+      // A link surrounds the whole content of the list item.
+      const resultsLinkElement = document.createElement('a')
+      resultsLinkElement.href = '/packages/' + result.id
```
daveverwer, line 65: Ideally, and especially if we tackle routing server-side, this URL should come down in the API response.

Public/javascripts/main.js
```diff
@@ -13,4 +12,83 @@ document.addEventListener('DOMContentLoaded', function(event) {
...
+      resultsLinkElement.appendChild(resultNameAndRepositoryContainer)
+
+      // Name.
+      const resultNameElement = document.createElement('h4')
+      resultNameElement.textContent = result.name
+      resultNameAndRepositoryContainer.appendChild(resultNameElement)
+
+      // Repository identifier.
+      const resultRepositoryElement = document.createElement('small')
+      resultRepositoryElement.textContent = result.owner + '/' + result.package_name
```
daveverwer, line 79: Similar here, we could include repository_identifier in the API response rather than the two components parts. I don't think there's a reason that the individual parts would be needed. That may change in the future of course, and so I'm less concerned about this one over the URL/routing issue above.

Public/javascripts/main.js
```diff
@@ -13,4 +12,95 @@ document.addEventListener('DOMContentLoaded', function(event) {
...
+}
+
+// Helpers
+
+function createSearchResultListItemElement(result, containerElement) {
+  const resultListItemElement = document.createElement('li')
+
+  // A link surrounds the whole content of the list item.
+  const linkElement = document.createElement('a')
+  linkElement.href = '/packages/' + result.id
```
daveverwer, line 83: Ideally, and especially if we tackle routing server-side, this URL should come down in the API response.

Public/javascripts/main.js
```diff
@@ -13,4 +12,95 @@ document.addEventListener('DOMContentLoaded', function(event) {
...
+  linkElement.appendChild(nameAndRepositoryContainer)
+
+  // Name.
+  const nameElement = document.createElement('h4')
+  nameElement.textContent = result.name
+  nameAndRepositoryContainer.appendChild(nameElement)
+
+  // Repository identifier.
+  const repositoryElement = document.createElement('small')
+  repositoryElement.textContent = result.owner + '/' + result.package_name
```
daveverwer, line 97: Similar here, we could include repository_identifier in the API response rather than the two components parts. I don't think there's a reason that the individual parts would be needed. That may change in the future of course, and so I'm less concerned about this one over the URL/routing issue above.

daveverwer:
This adds two library dependencies, but they're not very heavy. They are also well used, and reliable. If we were being super strict, the only reason I'm bringing `lodash` in is for the debounce function, which could be seen as overkill as that library is ~70k.

`axios` is tiny (16kb) and very focused in what it does so that's less of an issue.

Public/javascripts/main.js
finestructure: That's easy enough to add - do we have an issue to track it? I'll pick that up then.

finestructure:
Ok to undraft and merge?

daveverwer:
Yep!

Public/javascripts/main.js
daveverwer: I'll add an issue now, but it's low priority as this is already working.

daveverwer: #109

Issue #88, opened 2020-05-19
Populate release date metadata


daveverwer:
* Release date for every package version we store

@daveverwer to look at git command for fetching this.

daveverwer:
Alright, the git command for this one is pretty simple. Given a list of tag strings, which we have this command will find the commit hash for a specific tag:

`git rev-list -n1 0.0.2`

The output is a commit hash:

`43588020a5484d78417f3740c7fc4d409fc40f18`

Which can be turned into a date easily with this:

`git show -s --format=%ct 43588020a5484d78417f3740c7fc4d409fc40f18`

The output is a Unix time:

`1589652294`

Of all the output formats, I'd guess Unix time is easiest to parse (the `%ct`), but there are lots of options.

These could be chained together into one command of course, but we're going to use that `git show` command to process lots of commit hashes into dates for several things, so it might be good to keep that as a separate function that we can run.

daveverwer:
This can also be done in one command:

`git log -n1 --format=format:"%ct" 0.0.2`

Outputs the date of that commit as a Unix time.

daveverwer:
Interestingly, if we wanted to grab the hash at the same time for any of these, the format can be expanded:

`git log -n1 --format=format:"%H-%ct" 0.0.2`

Outputs:

`43588020a5484d78417f3740c7fc4d409fc40f18-1589652294`

PR #110, opened 2020-05-21
Release metadata

finestructure:
This is the first part of #88, getting the version metadata (release date, commit) from the git repositories.

I'll wire this up in a follow up PR.

### DIFF
--- a/Sources/App/Core/Git.swift
+++ b/Sources/App/Core/Git.swift
@@ -1,6 +1,9 @@
 import ShellOut
 
 
+typealias CommitHash = String
+
+
 enum Git {
 
     static func tag(at path: String) throws -> [Reference] {
@@ -11,4 +14,7 @@ enum Git {
             .map { Reference.tag($0) }
     }
 
+    static func revList(_ reference: Reference, at path: String) throws -> CommitHash {
+        try Current.shell.run(command: .init(string: "git rev-list -n 1 \(reference)"))
+    }
 }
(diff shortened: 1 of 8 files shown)


finestructure:
Unfortunately, this seems to have made the tests flakey. I'm getting random failures around unfulfilled futures again, like we used to.

Not sure what's going on - this shouldn't be making it any worse. Either we've just masked it and been lucky this hole time or I'm missing something.

I'll take another look.

finestructure:
Pushed a version that's closer to the previous version. I can't shake the nagging suspicion this is somehow related to output logging. I seem to be able to trigger failures on slower terminals or with more console output.

If this continues to be an issue, we should look into making the processing a continuous job. If we do that we won't run the risk of cutting off futures, because there won't be any waits and not process termination.

finestructure:
Well, that's silly. I'm appending the run commands to a list to double check what we processed and of course with more commands not being run it exposes that these commands are being appended to an array concurrently... I need to synchronise that.

Screenshot 2020-05-21 at 15 40 58

Sources/App/Core/Git.swift
```diff
@@ -0,0 +1,52 @@
...
+            .map(String.init)
+            .compactMap(SemVer.init)
+            .map { Reference.tag($0) }
+    }
+
+    static func revList(_ reference: Reference, at path: String) throws -> CommitHash {
+        try Current.shell.run(command: .init(string: "git rev-list -n 1 \(reference)"), at: path)
+    }
+
+    static func showDate(_ commit: CommitHash, at path: String) throws -> Date {
```
daveverwer, line 25: I don't think this is being used now? I only see it here and in a test. Are we keeping it around in case it's useful later? Same with `revList` above.

PR #111, opened 2020-05-22
Wire up releases

finestructure:
Implements #88 

Releases are now driven by live data:

Screenshot 2020-05-22 at 12 06 48


Sources/App/Views/PackageController/PackageShow+Query.swift
```diff
@@ -0,0 +1,116 @@
...
+            .unwrap(or: Abort(.notFound))
+    }
+}
+
+
+extension Package {
+    var defaultVersion: Version? {
+        guard
+            let versions = $versions.value,
+            let repositories = $repositories.value,
```
finestructure, line 49: NB: after discussing with Tanner Nelson in Discord I've arrived at this way to safely access relationship data: `model.$relationship.value` will return a double optional if the data is not loaded and a single optional if the data is loaded but nil (in case of optional relationships).

That means we can now safely use these helpers. Safely in the sense that we won't see crashes. The interface is still less than ideal, because inspecting a double optional to detect non-fulfilled relationships isn't great.

I'll probably move some of these extension over to `Package.swift` now that they're ok to use but I want to see how wiring everything up goes first.

daveverwer:
Haven't looked at this properly yet but from a quick look at the screenshot, it seems like there might be an "ago ago" problem ;)

finestructure:
Note that I took the screenshot before fixing the "ago ago" :)

finestructure:
get out of my head!!

daveverwer:
haha!

finestructure:
Screenshot 2020-05-22 at 12 24 26

finestructure:
Aw bugger, `RelativeDateTimeFormatter` doesn't appear to be available on Linux :(

```
2020-05-22T10:25:06.6665134Z /__w/SwiftPackageIndex-Server/SwiftPackageIndex-Server/Sources/App/Views/PackageController/PackageShow+Query.swift:95:25: error: use of unresolved identifier 'RelativeDateTimeFormatter'
2020-05-22T10:25:06.6665259Z         let formatter = RelativeDateTimeFormatter()
```

finestructure:
Looks like I'll need to implement a replacement.

finestructure:
Ok, that should be sorted for now

PR #116, opened 2020-05-22
Fixes #100

### LINKED ISSUE #100: Limit the number of search results in the JSON API
Two parts to this:

* The number of results for any query should be limited to a maximum of 20.
* If there are more than 20 results, include a boolean to indicate that so that a note can be shown to people to say they should narrow their search.

In order to achieve this, we should change the top-level JSON object to be a dictionary that includes a `results` key with the current array in it.

finestructure:
This changes the `search` endpoints return value to

```
{
  hasMoreResults: false
  results: []
}
```

The content of `results` is of the same type as the previous complete response.


Sources/App/Controllers/API/APISearchController.swift
```diff
@@ -14,13 +14,23 @@ extension API {
 
 
 extension API {
+    static let searchLimit = 20
+    static let searchLimitLeeway = 5
```
finestructure, line 18: The idea here is to allow a lilttle leeway so we don't cut off at 20 when there are 21 results. We'll return the full result set up to 24 and from 25+ we'll trim back to 20 and set `hasMoreResults` to `true.

Happy to cut off straight at 20 as well if you'd rather have a fixed value.

daveverwer: No, I think this is smart. 👍

daveverwer:
Let's not merge this now, I'll take it from here and add the client-side code for displaying when there are more results than we are showing.

daveverwer:
Screenshot 2020-05-23 at 10 34 38@2x


I think this is now ready for merging.

Issue #86, opened 2020-05-19
Populate repository history metadata


daveverwer:
* First commit date
* Last commit date
* Total number of commits
* Total number of releases

Source will be from local git repository.

@daveverwer to provide git commands for these.

daveverwer:
The first commit hash is easy to get:

`git rev-list --max-parents=0 HEAD`

The output is:

```

98283c3be7074a6938fd6f9becd14433c9a907a3
```

(**Note:** There does seem to be a consistent blank line at the top of this output 😬)

That can then be turned into a Unix time with [this](https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/issues/88#issuecomment-630750323)

daveverwer:
Scratch that, there's an easier way to get the Unix time for the first commit in one command:

`git log --max-parents=0 -n1 --format=format:"%ct"`

and then for the last commit:

`git log -n1 --format=format:"%ct"`

Much easier!

daveverwer:
The total number of commits is easy too:

`git rev-list --count HEAD`

Total release is also trivial:

`git tag`

and then parse that to see how many of those versions are semantic versions. Actually we've probably already got this right there in the database as the versions are definitely being parsed already.

PR #123, opened 2020-05-23
History metadata

finestructure:
Implements #86 

Wiring it into the top section is separate but this populates the database from the git repos.


Sources/App/Models/Package.swift
```diff
@@ -45,9 +45,6 @@ final class Package: Model, Content {
     @OptionalEnum(key: "processing_stage")
     var processingStage: ProcessingStage?
 
-    @Field(key: "last_commit_at")  // TODO: shouldn't this rather live in Repository? Is it needed at all?
-    var lastCommitAt: Date?
```
finestructure, line 49: I've moved this over to `Repository` to group all the git metadata together. It doesn't really make a difference with respect to lookups as we're already joining `Package` and `Repository`.

finestructure:
Good to go as well!

PR #124, opened 2020-05-23
JavaScript error handling for the search API call.

### LINKED ISSUE #120: Handle errors from the search API.
Fix [this](https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/blob/master/Public/javascripts/main.js#L37).

daveverwer:
Fixes #120.


daveverwer:
Alright, this is ready to go. Simple error messages look like this:

Screenshot 2020-05-23 at 15 57 00@2x

If there's a `reason` in the JSON response returned by the server, and it's *not* the same as the HTTP error message, it also gets displayed, like this:

Screenshot 2020-05-23 at 15 56 37@2x

Public/javascripts/main.js
```diff
@@ -9,20 +9,19 @@ document.addEventListener('DOMContentLoaded', function(event) {
...
-  // If there's a results element, its initial state should be hidden.
-  const resultsElement = document.getElementById('results')
-  if (resultsElement) { resultsElement.hidden = true }
+  // There are never results when the page loads, so hide the element.
+  setElementHiddenById('results', true)
 
   // If there is a search element, configure the search callbacks.
   const queryFieldElement = document.getElementById('query')
-  if (queryFieldElement) {
+  if (!!queryFieldElement) {
```
finestructure, line 17: Oh, Javascript, never change... 🤣

daveverwer: Yes :)

PR #127, opened 2020-05-23
Fixed bug where search results cleared after navigating back from a Package page

### LINKED ISSUE #108: Search results are lost when navigating between pages.
Steps to reproduce:

* Home page
* Type search query
* Pick a result and click to navigate to that page
* Click back

At this point, the search box still has the query in it, but the results are gone.

Fix is easy enough. When the DOM loaded, if there is text in the query field, re-do the search. It could even be as easy as just un-hiding the results div as I bet the results are still there. Needs a bit of testing.

daveverwer:
This was an absolute **pain** to fix! Full of different incompatibilities for Safari/Chrome/Firefox as well as some truly bizarre behaviour that means I now know more than I ever wanted to know about the page load lifecycle in every different browser. 😂

Fixes #108.

### DIFF
--- a/Public/javascripts/main.js
+++ b/Public/javascripts/main.js
@@ -1,3 +1,8 @@
+// Constants for session key storage.
+const SessionKey = {
+  searchResults: 'searchResults'
+}
+
 document.addEventListener('DOMContentLoaded', function(event) {
   // Force external links to open with a _blank target.
   document.addEventListener('click', function(event) {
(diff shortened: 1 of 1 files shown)


Public/javascripts/main.js
```diff
@@ -9,12 +14,10 @@ document.addEventListener('DOMContentLoaded', function(event) {
...
-  // There are never results when the page loads, so hide the element.
-  setElementHiddenById('results', true)
-
   // If there is a search element, configure the search callbacks.
   const queryFieldElement = document.getElementById('query')
   if (!!queryFieldElement) {
+    // When user input is entered into the query field, perform the search.
     document.addEventListener('input', function(event) {
       const searchQuery = queryFieldElement.value.trim()
       if (searchQuery.length > 0) {
```
finestructure, line 23: BTW, unrelated to this PR - I think we should debounce the search. Or is that happening and I’m not seeing it? It seems like we’re search per keystroke in my brief testing.

daveverwer: Yep, debouncing was in there originally... but the fix to #128 was actually that the debounce was in slightly the wrong location! Pushing a fix to that now.

PR #129, opened 2020-05-23
Moved the location of the debouncing to cope with fast input.

### LINKED ISSUE #128: When deleting a search query quickly, the search results display a previous value.
Screenshot 2020-05-23 at 21 42 17

daveverwer:
Need to debounce the event itself, not the call to performSearch. 👍

Fixes #128.

### DIFF
--- a/Public/javascripts/main.js
+++ b/Public/javascripts/main.js
@@ -68,7 +68,7 @@ function performSearch(searchQuery) {
   setElementHiddenById('results', false)
 }
 
-window.clearSearchResults = function() {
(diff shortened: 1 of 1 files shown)


Public/javascripts/main.js
```diff
@@ -49,7 +49,7 @@ window.addEventListener('pageshow', function(event) {
   }
 })
 
-window.performSearch = _.debounce(function(searchQuery) {
```
daveverwer, line 52: @finestructure Here was the original debounce call, but it was in slightly the wrong place..

PR #132, opened 2020-05-24
Wire up history data

finestructure:
This pulls in the live data from our git repository crawl.

Now there's a slight issue here in that `RelativeDateTimeFormatter` returns phrases like `in 5 days` and `2 months ago`, which don't work as is with our wording "In development for ...":

Screenshot 2020-05-24 at 10 12 55

We can either just use our own relative date time formatter which already has the capability to return the phrase without `in` or `ago` or we could rephrase the sentence to say `Development started ...`, for instance.

I'm leaning towards the former, in which case I'd update the interface of the formatter for use on both platforms and add some more tests as it'd be here to stay.


finestructure:
Unrelated (and low prio) but I just happened to notice it: should we maybe try a checkbox or something like that instead of the mail icon for the issues and PR section?

daveverwer:
> Now there's a slight issue here in that `RelativeDateTimeFormatter` returns phrases like `in 5 days` and `2 months ago`, which don't work as is with our wording "In development for ...":

I'm not sure what the formatter in the standard library returns, but the one we were looking at in Rails deliberately omits the "ago" leaving you to add that manually, if needed. This would work in this situation.

"In development for 5 years."
"In development for 20 minutes."
etc...

PR #136, opened 2020-05-24
Logging improvements

### LINKED ISSUE #133: Dial back Rollbar logging
We're way past the request limit of 5k/month, even without the GH issues:

IMG_0070
IMG_0071

I'm not sure there's a whole lot we can do to stay under 5k/month and the plans seem to be pricey for what we would need from a paid plan.

It might be best if we identified a few critical messages we need to report up to Rollbar and only send those and deployment items.

Later on I can set up a Grafana/Loki dashboard that pulls in the console output in a few of hours that should give us more of less the same level of insight.

### LINKED ISSUE #135: Log Shellout errors in more detail
Dave's seeing an issue running git commands and it's impossible to tell which package is affected and which commands exactly is being run. Log those to the console.

finestructure:
Merge after #132 

Fixes #133 by adding a `ROLLBAR_LOG_LEVEL` env variable (defaulting to `critical` if unset), allowing us to manage the level externally. We're only reporting one critical error at the moment, so this will reign in errors we're reporting.

Fixes #135 

While it's hard to capture the package id or url without wrapping all shell.run calls in error handlers, it's quite simple to at least add the command and path to the existing exceptions. Give that most commands provide enough context from that alone, this should help us with debugging quite a bit. If we still need more, we can go in an wrap the areas where we're lacking detail.

### DIFF
--- a/Sources/App/Commands/Analyzer.swift
+++ b/Sources/App/Commands/Analyzer.swift
@@ -181,8 +181,9 @@ func reconcileVersions(application: Application, package: Package) -> EventLoopF
         return try Git.tag(at: cacheDir)
     }
     .flatMapError {
-        Current.reportError(application.client, .error,
-                            AppError.genericError(pkgId, "Git.tag failed: \($0.localizedDescription)"))
+        let appError = AppError.genericError(pkgId, "Git.tag failed: \($0.localizedDescription)")
+        application.logger.report(error: appError)
+        return Current.reportError(application.client, .error, appError)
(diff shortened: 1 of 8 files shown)


daveverwer:
Yea just the path is fine, that includes plenty of data about which repository it is. Is this ready to merge? I'll get it running on my machine if it is 👍

daveverwer:
Alright, I've been running this locally and the error is now crystal clear. Logged as #139.

Issue #137, opened 2020-05-24
Clean up and refactor JavaScript


daveverwer:
The JavaScript is getting out of control already!

* Split JavaScript into multiple source files and classes.
* Combine and minify the multiple source files when changing code.

PR #138, opened 2020-05-24
Keyboard search query navigation

### LINKED ISSUE #107: Add keyboard navigation of search results.
Arrow keys up and down to move between results, enter to visit the result package page.

daveverwer:
Fixes #107 

The JavaScript is way past the level and complexity where I'm comfortable working in a single file just doing everything inline, it needs cleanup and refactoring. Logged this as #137.


finestructure:
Not sure if it's from the PR and it's not a big deal but I'm noticing some flickering in the results drop down after almost every keystroke. There seems to be a dark/black interstitial shown between refreshs. I'll try to record but it might be too fast.

daveverwer:
Yea it’s not from this PR, it’s been there since the beginning. It’s annoying and I should fix it 😂

PR #140, opened 2020-05-24
Issue 139 tagnames

### LINKED ISSUE #139: Bug when running `git log` commands
After the logging in #136 it's a little clearer that there's an error with the code to gather the history  metadata from git. When running against some repositories, there are errors like this:

```
[ ERROR ] analysis error: Shell command failed:
command: "git log -n1 --format=format:"%H-%ct" 0.1.0"
path:    "/Users/dave/Development/SwiftPackageIndex/Checkouts/github.com-fabianfett-vapor-lambda-runtime"
message: "ShellOut encountered an error
Status code: 128
Message: "fatal: ambiguous argument '0.1.0': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'"
Output: """
```

Running `git tag` on that repository shows the error clearly:

```
v0.1.0
v0.2.0
v0.3.0
```

The `v` is being omitted from the tag, probably as it's being stripped as it's not technically a valid version. We'll need to keep the original tag name around as well as the parsed semantic version.

Running the exact same command with the correct tag name works well:

```
git log -n1 --format=format:"%H-%ct" v0.1.0
```

Outputs:

```
8ecc257775d0567cf404287d2c7a40cee7d033d4-1579047775
```

finestructure:
Fixes #139

### DIFF
--- a/Sources/App/Models/Reference.swift
+++ b/Sources/App/Models/Reference.swift
@@ -3,7 +3,11 @@ import Foundation
 
 enum Reference: Equatable {
     case branch(String)
-    case tag(SemVer)
+    case tag(SemVer, _ tagName: String)
+
+    static func tag(_ semVer: SemVer) -> Self {
+        return .tag(semVer, "\(semVer)")
+    }
 
     var isBranch: Bool {
         switch self {
@@ -18,7 +22,7 @@ enum Reference: Equatable {
         switch self {
             case .branch:
                 return nil
-            case .tag(let v):
+            case .tag(let v, _):
                 return v
         }
     }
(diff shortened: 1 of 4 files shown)


finestructure:
Regarding deployment: if we push this as is, we'll get the following error from decoding existing tags with the old format:

```
{"error":true,"reason":"invalid field: reference type: Optional<Reference> error: dataCorrupted(Swift.DecodingError.Context(codingPath: [], debugDescription: \"none of the required keys found\", underlyingError: nil))"}
```

This will correct itself once all versions have been re-processed.

I'll also open up a PR soon that squashes the schema, which will fix this the blunt way =)

Sources/App/Models/Reference.swift
```diff
@@ -3,7 +3,11 @@ import Foundation
 
 enum Reference: Equatable {
     case branch(String)
-    case tag(SemVer)
+    case tag(SemVer, _ tagName: String)
```
daveverwer, line 6: Ah, interesting. Avoiding another database field. At some point, this can go too far but I think this is probably OK since they are basically the same data at this point.

daveverwer:
> Regarding deployment: if we push this as is, we'll get the following error from decoding existing tags with the old format:

Yep, this is one reason I haven't pointed the domain at the site yet. We have free reign to break everything at this point. 👍

daveverwer:
I'll deploy this now.

Issue #126, opened 2020-05-23
Page rendering and missing data


daveverwer:
This is obviously something that's happening more for me since I have a partially analysed database (I normally only let `./analyze_loop.sh` run for 20-30 minutes usually after a reset) but there's something that concerns me a bit about how we're dealing with missing data when rendering the package page.

Sometimes, I click through from a search result to the package page and there's a piece of data needed to render that package page that results in the page being a JSON error response. It would make more sense to me if the page were able to render under *almost* any circumstance, with missing data rather than a total refusal to render. 

I know this is not going to happen in a live situation anywhere near as often, but I bet it will happen occasionally, and it's hard to test for. I can see two potential solutions for this:

1. We make the Package page much more tolerant about missing data. I'd much rather render an empty bullet (which isn't ideal!) than have the whole page fail.
2. We exclude any packages that will not render from the search results, meaning that it'd be much less likely that people would see a page in this state.

Oh, and regardless of which of these we go with, we should have some HTML formatted error pages rather than just outputting a JSON error. I've logged that separately as #125.

finestructure:
Yes, that’s something we need to look at when refining `PackageShow.Model`. I’m starting to have to make fields optional here in order to be able to deal with missing data. The nice thing is the optionals really guide us and allow us to either fail late (on page rendering) or early, when querying.

We’re automatically benefitting from packages that tend to fail to render due to missing versions etc also not being picked up by the search query. But especially around other metadata like releases or commit info we’ll have to either transform the optional into placeholder text on the show page or collapse/hide the whole element.

BTW I have not yet encountered any JSON errors bubbling up (but I’m also not browsing that many packages). Would be good to know when/how that happens because I don’t think that _should_ actually be the case when it’s just missing data.

Issue #122, opened 2020-05-23
Handle absence of all L&P info


finestructure:
<img width="1059" alt="Screenshot 2020-05-23 at 13 32 03" src="https://user-images.githubusercontent.com/65520/82729707-ef12d500-9cf9-11ea-8e7d-cfc722eb8e1b.png">

PR #142, opened 2020-05-25
Make top section elements optional and hide them if empty

### LINKED ISSUE #119: Complete `PackageShow.View` top section
- [x] populate history data #86
- [x] wire up history data #118
- [ ] make author and activity optional until we parse that data

finestructure:
Merge after #141 

Fixes #119 

From the new snapshot test:

Screenshot 2020-05-25 at 08 49 48


daveverwer:
Unlike #122, I think these are much more reasonable to hide if for some reason we can't get the information. The four clauses are:

1. Authors - We should be able to get this from GitHub once that's implemented. The package author can't really do anything to fix that.
2. History - We should be able to get this information in *every* case. I can't see a situation where this could be blank.
3. Activity - Again, this is a GitHub thing, so the package author can't do anything. If we can't get it for some reason it's OK to hide.
4. Products - This one is tougher, if there really isn't a library or an executable, the package author probably wants to know about it. I think there's an argument for showing this with a link to the FAQ, but it's an edge case as I don't think it'll happen very often, if at all.

What do you think?

finestructure:
Yes, I think some these are only technically optionals but probably not in practise. I think dealing with them by hiding them in a first pass makes us degrade gently right now and then we can revisit how we're going to deal with pathological cases in a second pass, perhaps in the context of #126?

daveverwer:
Sounds good to me.

PR #143, opened 2020-05-25
Hide L&P section when there is no lpInfo at all

### LINKED ISSUE #122: Handle absence of all L&P info

finestructure:
Fixes #122 

This makes the whole L&P section optional when there is no info at all. I.e. we would previously just display the "Languages and Platforms" header without any content below.

Screenshot 2020-05-25 at 09 10 47


daveverwer:
This is an interesting one, I think there's a good argument against removing the heading completely. Instead, having the heading with a link to a FAQ that I'll write about why there's no information there, and how the package author can fix it.

People might not even know we can parse that information if it just disappears.

Let me prioritise writing some things, and maybe we can amend this to link to that after I've written it... Or, maybe we just make that change now with just a link to example.com for now and add an issue to change it to the correct link once I've written the FAQ.

finestructure:
I totally agree with not hiding and instead link to explainers etc. My thinking was that we do the easy thing now to make it look ok in the worst case and then fill in with links etc in a follow-up pass. Does that make sense?

daveverwer:
I agree, but I think this one is important enough to prioritise before launch. Even if it's just a markdown file on GitHub this one specifically should be a link, so let's do that now and we can just replace the URL when it exists.

finestructure:
FYI - the other PR, #147 , is continuation from this one. Let me know if you want to patch this one here up with a placeholder link or something to move the merge train along.

Shall I just embed something like "There are no language versions or platforms available." for now, perhaps with a link to `/faq`?

daveverwer:
> Shall I just embed something like "There are no language versions or platforms available." for now, perhaps with a link to `/faq`?

Yes, that's perfect.

PR #151, opened 2020-05-25
Fix textarea line height, and stop the text area wrapping text.

daveverwer:
Super simple fix to a couple of issues with the textarea.


daveverwer:
Oops, I just noticed I removed a line accidentally...

PR #156, opened 2020-05-26
Logging

### LINKED ISSUE #152: Add Grafan/Loki to capture console logs

finestructure:
Parking this for now. Grafana and Loki are running but promtail is not forwarding logs at the moment.

- the built-in config only picks up `/var/log` and doesn't get the container logs
- supplying a custom config works when I try it locally but fails to load on prod:
```
spi@localhost ~/SwiftPackageIndex-Server (logging)> docker logs spi_promtail_1
Unable to parse config: read /etc/promtail/promtail.yaml: is a directory
```
- the Loki docker plugin to allow forwarding without promtail doesn't work for some reason

I'm running docker-compose 1.25.5 on both systems, so it's not clear to me why one would map in the config file as a directory.

--

Fixed by baking in the config for now.

Fixes #152


daveverwer:
I'm no expert with this stuff as you know, but if you want to "rubber duck" it with me, let me know and we can jump on a call.

PR #181, opened 2020-05-27
Recents on homepage

### LINKED ISSUE #149: Populate the "Recent Packages" and "Recent Releases" data
- [x] create mat views for queries
- [ ] wire up queries in home page

The Recent Packages and Recent Releases lists on the home page need data.

## Recent Packages

This should be the last 10 (valid) packages ordered by created_at on the packages table. In other words, the last 10 packages that were added to the index.

## Recent Releases

The last 10 versions across all packages. Limited to be unique per package (so even if a package does two releases in quick succession, it's not listed twice).

finestructure:
This wires up the recents on the home page (with sparse local data):

Screenshot 2020-05-27 at 10 35 36

Fixes #149


Sources/App/Views/Home/HomeIndex+Model.swift
```diff
@@ -0,0 +1,41 @@
+import Fluent
+import Plot
+
+
+extension HomeIndex {
+    struct Model {
+        var recentPackages: [DatedLink]
+        var recentReleases: [DatedLink]
```
finestructure, line 8: Tempted to rename `DatedLink` to `DateLink` - or am I being too obsessed with naming? 😅

daveverwer: DatedLink is better 👍

finestructure:
Hold on, I just realised I'm taking the full set of records from the view. Should we limit this to 6 each in the query? Or just take what the materialized view returns, which is 10 at the moment?

daveverwer:
10 sounds good.

After that's done, I'm happy with this so go ahead and merge.

finestructure:
Marking draft - I broke a test

PR #182, opened 2020-05-27
Update missing package info copy

### LINKED ISSUE #171: Missing languages and platforms metadata copy
I said I'd think about the design for how the languages and platforms metadata looks when there's no metadata to show for any of the versions we're indexing.

It currently displays this:

Screenshot 2020-05-27 at 07 01 27@2x

When we have no language or platform data at all, that should be replaced with a simple `p` tag with the following copy:

Screenshot 2020-05-27 at 07 12 12@2x

Text is also below, for copying and pasting purposes!

> The manifest for this package doesn't include metadata on which versions of Swift, and which platforms it supports – Are you the package author? [Learn how to fix this](#).

finestructure:
Fixes #171 

I've created a stub wiki page where the link is pointing to: https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/wiki/Fixing-missing-Language-and-Platform-information


daveverwer:
The change itself here looks good 👍 but when I look at what's changing in the code, I'm not sure the original code was actually working well, and I don't know whether the original text was ever displayed.

For example, `JohnSundell/Files` has no version or platform information, but shows this: 

Screenshot 2020-05-27 at 20 23 26@2x

If this PR is to fix that, I'd have expected to see the original text, but I don't think I've seen that original text on any package. It's been a long day though, so I may be missing something obvious.

finestructure:
Mmm, weird. You can see in the snapshot that it’s displaying that paragraph if the clause is `nil` but I wonder if there’s an edge case with certain data. When you say it has no version or platform info - can you set a breakpoint here and check if it returns nil?

https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/blob/9556de836860d1d37b95841eefab51894fc90eaa/Sources/App/Views/PackageController/PackageShow%2BModel.swift#L164

daveverwer:
Just did a bit of poking at it and something strange is definitely happening. I ended up placing a breakpoint on this line:

https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/blob/9556de836860d1d37b95841eefab51894fc90eaa/Sources/App/Views/PackageController/PackageShow%2BView.swift#L76

... and it hit that breakpoint, and then rendered this:

Screenshot 2020-05-27 at 20 46 41@2x

finestructure:
That’s bizarre... how is that even possible?

daveverwer:
Which package does it render the correct text for? I can check that one, I think I have a mostly full analyze done on my local database.

finestructure:
Oh wait, I think the html might be wrong. In the non-nil case we send `<li>` and in the other `<p>`, don’t we?

finestructure:
Ah, no, it’s `<ul>`, never mind.

finestructure:
It’s just the mock data in the snapshot test.

daveverwer:
Ah right, can you reproduce this with real data? I always err on the side of testing with real data.

finestructure:
I just realised that a lot of my data was broken, because the shell command picks up my environment which has `swiftenv` installed but I don't have all history swift versions installed:

```
[ ERROR ] Generic error: getManifests failed: Shell command failed:
command: "swift package dump-package"
path:    "/Users/sas/Projects/SPI-Server/SPI-checkouts/github.com-johnsundell-files"
message: "ShellOut encountered an error
Status code: 127
Message: "swiftenv: version `4.1' (set by /Users/sas/Projects/SPI-Server/SPI-checkouts/github.com-johnsundell-files/.swift-version) is not installed"
```

This does not affect the Linode box (nor I guess you?). This doesn't have anything to do with the issue - just want to mention it as an FYI in case you run into issues where local data is different than prod.

finestructure:
Ok, that's fixed now!

Screenshot 2020-05-28 at 08 50 37

finestructure:
I've also added #191 because that's actually an easy thing for us to do to help populate that section

Issue #64, opened 2020-05-14
Add two RSS feeds


daveverwer:
* Newly added packages.
* New package versions.

finestructure:
I might start looking into this one, unless you're keen on tackling it?

daveverwer:
Go for it!

A couple of points for implementation:

* Include 100 posts in a feed, but it'd be great if this were easily customisable (by us, it doesn't need to be a user-configurable thing or a URL parameter).
* Every URL in the feed, including ones in the body should be a fully qualified, absolute URL.
* For the "Recent Packages" feed I was thinking the fields would be something like this:
  * Title: [Package Name]
  * Link: https://swiftpackageindex.com/[Repository Owner]/[Repository Name]
  * Description:
```html
<h2><a href="https://swiftpackageindex.com/[Repository Owner]/[Repository Name]">[Package Name]</a></h2>
<p>[Package Summary]</p>
<small><a href="https://swiftpackageindex.com/[Repository Owner]/[Repository Name]">[Repository Owner]/[Repository Name]</a></small>
```
* For the "Recent Releases" feed I was thinking the fields would be something like this:
  * Title: [Package Name] - [Version X]
  * Link: https://swiftpackageindex.com/[Repository Owner]/[Repository Name]
  * Description:
```html
<h2><a href="https://swiftpackageindex.com/[Repository Owner]/[Repository Name]">[Package Name] - [Version X]</a></h2>
<p>[Package Summary]</p>
<small><a href="https://swiftpackageindex.com/[Repository Owner]/[Repository Name]">[Repository Owner]/[Repository Name]</a></small>
```

We can look at adding some of the other available metadata later. This would be a great start.

Also, here's the official RSS validator. If it passes that, we're good! https://validator.w3.org/feed/

finestructure:
In RSS terms, recent packages and release would each go into its on `channel`, right?

daveverwer:
There would be completely separate feeds for each of these, with separate autodiscovery meta tags.

finestructure:
We have a couple of data gaps here, if we're looking to drive both feeds by our recent views:

- package summary
- released version

We'll need to extend the view to provide those or create another view with those available.

daveverwer:
Let's drop package summary then.

Package version is important, and should also probably be shown in Recent Releases on the home page too.

PR #189, opened 2020-05-27
Recent Packages and Releases styling tweaks

daveverwer:
As I was looking at this, the limit of 10 seemed just a little too long. On a 15" laptop, it pushes the footer off the end of the page, and the dimensions just seem a little too long and thin to look natural.

This is how it looked:

Screenshot 2020-05-27 at 21 49 51@2x

Reducing that to 7, and applying a little styling I think it looks much more balanced:

Screenshot 2020-05-27 at 21 56 59@2x

You'll notice that I changed the limits in Core/Constants.swift rather than in the view. This is intentional. I think that view is going to need to be much bigger than 10 to support the RSS feeds (in #64) that will use these same queries. I'd say maybe 100 rows is more like what we need to return for those. We can increase the view as part of implementing that, but this is why I've changed the constants here, not the view.


finestructure:
Good point about the RSS feed! I'll look into bumping the mat view limit.

PR #194, opened 2020-05-28
Activity metadata

### LINKED ISSUE #87: Back-end support for pulling issue and pull request data from GitHub
This is an additional (or even multiple additional) GitHub API request to fetch:

- [x] Number of open issues
- [x] Number of open pull requests
- [x] Date of last issue being closed
- [x] Date of last pull request being closed/merged

finestructure:
Fixes #87 

Screenshot 2020-05-28 at 15 37 42

There are a few where the last issue dates are `null` but I'd like to leave that for a follow up.


finestructure:
NB: you'll notice a slight discrepancy when comparing numbers against actual repos, which is due to drafts.

For instance, Carthage right now has

- 147 open issues
- 21 pull requests, one of which is a draft

Screenshot 2020-05-28 at 18 29 34

In my local db I've got 148 open issues and 20 open pull requests. The discrepancy stems from us getting a report of a total of 168 issues and then subtracting the 20 open pull requests instead of 21.

I'll check if I can get a pull request count including drafts.

finestructure:
But not today :)

I think it's ok to merge that provided there are no other issues with it and then I'll do a follow up tomorrow!

finestructure:
Actually, the issue isn't the draft. The issue is that the base branch of one PR isn't the default branch: https://github.com/Carthage/Carthage/pull/2881

I think if I simply drop the query parameter to require the PR to go to the default branch everything would align. It sort of feels like that'd be the sensible thing to do.

What do you think?

daveverwer:
> I think if I simply drop the query parameter to require the PR to go to the default branch everything would align. It sort of feels like that'd be the sensible thing to do.
> 
> What do you think?

100% agree. 

It almost doesn't matter what criteria we use, these numbers are mainly so people get a sense of how active, and roughly how many issues are open on a project. The only thing people might notice is if the counts don't line up with what GitHub says, so whatever criteria they use for their counts is what we should try and match.

PR #198, opened 2020-05-29
Count all prs

### LINKED ISSUE #196: Remove branch criterion on pulls request
See https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/pull/194#issuecomment-635556959

finestructure:
Fixes #196 

Screenshot 2020-05-29 at 07 33 47


Sources/App/Commands/Ingestor.swift
```diff
@@ -8,25 +8,51 @@ struct IngestorCommand: Command {
     struct Signature: CommandSignature {
         @Option(name: "limit", short: "l")
         var limit: Int?
+        @Option(name: "id")
+        var id: String?
```
finestructure, line 12: Allowing

```
vapor-beta run ingest --id 8b371914-95bc-4917-aa73-d4778fc6fc5d
```

to avoid busting prod's ingestion capability 😅

Sources/App/Core/Github.swift
```diff
@@ -82,35 +82,24 @@ enum Github {
             let repoUri = try? apiUri(for: package, resource: .repo),
             let issuesUri = try? apiUri(for: package, resource: .issues, query: ["state": "closed",
                                                                                  "sort": "closed",
-                                                                                 "direction": "desc"])
+                                                                                 "direction": "desc",
+                                                                                 "page_size": "100"]),
```
finestructure, line 86: I've also opened up issue fetching to 100 per page. This was perhaps the reason we saw some gaps in the "last closed" fields. I.e. if all last 30 closed items were of only one kind, the last closed date of the other kind would be `nil`, as they are pushed off the first page.

This is something we could also correct by fetching the next page conditionally but I think the better approach is to revamp this to use the GraphQL API.

PR #209, opened 2020-05-29
Add authors field

### LINKED ISSUE #208: Add authors fields
Add the author fields to the schema (without populating) in order to move forward with the schema squash.

finestructure:
Fixes #208 in preparation to parse author data from git repos

### DIFF
--- a/Sources/App/Models/Author.swift
+++ b/Sources/App/Models/Author.swift
@@ -1,0 +1,5 @@
+
+struct Author: Codable, Equatable {
+    var name: String
+    var url: String
+}
--- a/Sources/App/Migrations/CreateRepository.swift
+++ b/Sources/App/Migrations/CreateRepository.swift
@@ -89,7 +89,7 @@ struct AddActivityFields: Migration {
 struct AddAuthors: Migration {
     func prepare(on database: Database) -> EventLoopFuture<Void> {
         return database.schema("repositories")
-            .field("authors", .array(of: .json))
+            .field("authors", .array(of: .json), .sql(.default("{}")))
             .update()
     }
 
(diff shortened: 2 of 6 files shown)


Sources/App/Models/Author.swift
```diff
@@ -0,0 +1,5 @@
+
+struct Author: Codable, Equatable {
+    var name: String
+    var url: String
```
finestructure, line 4: Does this suffice or do think we need more fields?

daveverwer: The only possible one would be a GitHub username, but it's easy enough to add in the future and I don't think we have any immediate plans for it, so let's just keep it to the fields we need, or know we're going to need.

I presume we're not going to try and do any author optimisation? This looks like a simple one -> many relationship, and I think that's wise. Someone might want one URL for one project, and another URL for another project and the easiest way to do that is just to have 1->many  rather than many<->many.

finestructure: This would be an inline field, just like swift versions and platforms. So adding a field later is not a problem and doesn't require a schema change, either.

I didn't want to go into adding a whole relation because as you say we don't really need the normalisation and it'll make the queries more involved as we'd have to add yet another `.with(\.$authors)` clause in places.

Should I maybe throw in a field `alias` that we can use for various purposes like Github/Gitlab name or Twitter handle or whatever people prefer instead of their real name?

daveverwer: I think if adding things later doesn't require a schema change, we should leave it as is for now. I think as we narrow down what the external metadata file looks like, this will become clearer and I don't want to assume the outcome of that process.

finestructure: Good to go then :)

PR #215, opened 2020-05-30
Slightly easier control of which snapshot tests to run

daveverwer:
Merge after #214


finestructure:
I was just going to mention this on the other test: you can set `record = true` in `setUp` and it'll re-record all snapshots. I think that's the same as this, isn't it?

daveverwer:
Ah, I didn't know that. I had been copying it to every test manually. I'll edit this a bit then. I'd still like the lines in there, just set to false so I don't have to add them every time.

daveverwer:
Looking at this, would it be OK if we left it like this? The alternative would be to leave a commented out `record = true` in every test, and also in the `setUp` method. I think this method makes it a little clearer what's happening.

PR #219, opened 2020-05-30
Added Google Tag Manager code

### LINKED ISSUE #178: Integrate Google Analytics

daveverwer:
Fixes #178 

When you merge this, could you also tag and deploy it, please? I'll check it's all working later this evening.

### DIFF
--- a/Sources/App/Views/PublicPage.swift
+++ b/Sources/App/Views/PublicPage.swift
@@ -33,8 +33,34 @@ class PublicPage {
             ),
             .script(
                 .src("/javascripts/main.min.js?\(resourceReloadQueryString())")
-            )
+            ),
+            analyticsHead()
         )
+    }
+
+    /// The Google Tag Manager code to be inserted into the <head> element.
+    /// - Returns: A <script> containing the Google Tag Manager template code.
+    final func analyticsHead() -> Node<HTML.HeadContext> {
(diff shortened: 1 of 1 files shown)


Sources/App/Views/PublicPage.swift
```diff
@@ -33,10 +33,36 @@ class PublicPage {
...
+            ),
+            analyticsHead()
         )
     }
 
+    /// The Google Tag Manager code to be inserted into the <head> element.
+    /// - Returns: A <script> containing the Google Tag Manager template code.
+    final func analyticsHead() -> Node<HTML.HeadContext> {
+        let environment = (try? Environment.detect()) ?? .development
+        return .if(environment == .production,
```
daveverwer, line 45: There's a question here of whether this should be `== .production` or `!= development`. I went with the former in the end as it's not super important that this gets a test, but we don't *ever* want this to execute in an environment other than production.

For example, I can see us adding a staging environment, and this code is future proof against that.

finestructure: I think this is the right call for now. I think when we add another env we'll want to drive this from a `GOOGLE_ANALYTICS_KEY` env variable instead and pass it in from deployment.

daveverwer:
Tested this on live and it's all working 👍

Issue #125, opened 2020-05-23
Enable server errors to render in HTML format, not just JSON.


daveverwer:
We need a page to handle 404/500/etc... errors.

It should show the error message that currently gets returned as JSON, but in a more attractive format and with the rest of the PubicPage template around it so that people have somewhere to navigate to after an error occurs.

PR #220, opened 2020-05-30
Issue 211

### LINKED ISSUE #211: Broken package MockingJ
http://173.255.229.82:8080/packages/8010F85D-D099-43AA-876E-BE05FB69BD45

```
{"error":true,"reason":"Something went wrong."}
```

finestructure:
Fixes #211

### DIFF
--- a/Sources/App/Models/Reference.swift
+++ b/Sources/App/Models/Reference.swift
@@ -3,7 +3,7 @@ import Foundation
 
 enum Reference: Equatable {
     case branch(String)
-    case tag(SemVer, _ tagName: String?)
+    case tag(SemVer, _ tagName: String)
 
     static func tag(_ semVer: SemVer) -> Self {
         return .tag(semVer, "\(semVer)")
@@ -30,9 +30,9 @@ enum Reference: Equatable {
 
 
 extension Reference: Codable {
-    struct Tag: Codable, Equatable {
+    private struct Tag: Codable {
         var semVer: SemVer
-        var tagName: String?
+        var tagName: String
 
         var asReference: Reference { .tag(semVer, tagName) }
     }
@@ -45,11 +45,6 @@ extension Reference: Codable {
         }
         if let value = try? container.decode(Tag.self, forKey: .tag) {
             self = value.asReference
-            return
-        }
-        // try legacy format last
-        if let value = try? container.decode(SemVer.self, forKey: .tag) {
-            self = .tag(value, nil)
             return
         }
         throw DecodingError.dataCorrupted(
@@ -78,7 +73,7 @@ extension Reference: CustomStringConvertible {
     var description: String {
         switch self {
             case .branch(let value): return value
-            case let .tag(semVer, tagName): return tagName ?? "\(semVer)"
+            case .tag(_, let value): return value
         }
     }
 }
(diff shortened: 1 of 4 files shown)


Sources/App/Models/Reference.swift
```diff
@@ -47,6 +47,11 @@ extension Reference: Codable {
             self = value.asReference
             return
         }
+        // try legacy format last
+        if let value = try? container.decode(SemVer.self, forKey: .tag) {
+            self = .tag(value, nil)
+            return
+        }
```
finestructure, line 54: I'm actually not convinced we should try and recover from this given that we're going to squash the scheme and clear out that bad data. It's going to be cruft immediately after that.

I think the main lesson learned is to be careful (and a test) when making changes to our JSON payloads. What happened here was that when you go from one field to two the JSON format changes from an inline field to an object. And that's not something you can simply recover from by adding an optional field.

I think I'd just revert that last commit, only add the error reporting, and then let the schema squash handle the rest.

What do you think?

finestructure: PS: This failure to clear out wasn't a widespread issue, because we deployed the initial format change a while back. Most fields were processed fine.

The issue where the format was a problem was only introduced when I added scoring. At that point `updateStatus` started to fail if it encountered records with old `Reference` data. And there were a few holdouts that haden't been processed yet, presumably because they'd failed for other reasons until now.

daveverwer: > What do you think?

I'm fine with what you say here, but I think it's worth talking about something else around the same subject. I've noticed a marked reduction in Rollbar errors (actually, to zero over the last several days). This kind of crash is exactly what I'd like Rollbar to notify us of. Someone hits a page, and it 500s while generating it.

Should that be happening already? If not, I think we should try to make it happen.

finestructure: Yes, there's really only one critical error we're currently reporting: if the checkouts directory can't be created.

I agree, this here is a worthy candidate :D I'll roll back the field change and add reporting a critical error up to rollbar.

daveverwer: Can we make more generic, so all 500s during a page render result in a Rollbar error?

finestructure: I'm sure we can hook that into the middleware. I'll add an issue to track, ok?

daveverwer: Basically, if someone expected to see a page and instead saw an error, I'd like to at least be aware.

Errors like this are super important. If someone tries the site for the first time and see an error when clicking through to a package, they're going to judge the whole site by that one result, and potentially never come back. It could be failing for days/weeks on some packages if we don't know about this kind of error.

daveverwer: > I'm sure we can hook that into the middleware. I'll add an issue to track, ok?

Perfect, thanks!

daveverwer: I also just bumped up the priority of #125 into Essential for the same reason. That raw JSON error is a killer to people's confidence in the site. We need to make it render errors in the page template.

finestructure: Yep, totally!

finestructure:
This should be good to go now!

PR #225, opened 2020-05-30
Don't auto-restart migrations

finestructure:
This is correcting a small issue that the migration container was also set to restart. Because it's a on-off command, that meant it was being restarted every 30s or so. Which is not harmful - it just does nothing.

What happened though was that I reset the schema to deploy the schema squash and to my surprise it suddenly reappeared 😆


finestructure:
It's a little confusing: `restart = always` was part of the `shared` block that got passed into all commands. What this does is remove it from the shared block and add it only to the ones that should have it! :)

PR #227, opened 2020-05-30
Add basic error page

### LINKED ISSUE #125: Enable server errors to render in HTML format, not just JSON.

finestructure:
Fixes #125 

Shippable design and content 😅 

Screenshot 2020-05-30 at 18 35 24

### DIFF
--- a/Sources/App/Views/Error/ErrorPage.swift
+++ b/Sources/App/Views/Error/ErrorPage.swift
@@ -1,0 +1,17 @@
+import Plot
+import Vapor
+
+
+final class ErrorPage: PublicPage {
+    let status: HTTPStatus
+
+    init(_ status: HTTPStatus) { self.status = status }
+
+    override func content() -> Node<HTML.BodyContext> {
+        .div(
+            // FIXME
+            .h1(.text("ERROR \(status)"))
+        )
+    }
+
+}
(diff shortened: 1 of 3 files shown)


Sources/App/Views/Error/ErrorPage.swift
```diff
@@ -0,0 +1,17 @@
...
+    let status: HTTPStatus
+
+    init(_ status: HTTPStatus) { self.status = status }
+
+    override func content() -> Node<HTML.BodyContext> {
+        .div(
+            // FIXME
+            .h1(.text("ERROR \(status)"))
+        )
+    }
```
finestructure, line 15: Here's where we can switch on error codes and return custom content.

finestructure: If you've got some copy/design I can add it to this PR. Or we merge and fill it in later? Let me know!

daveverwer:
This is wonderful, and already 100x better than what we had with the raw JSON. 👍

I'll make a follow-on PR now to pretty it up a little.

PR #228, opened 2020-05-30
Error page view formatting and view model.

daveverwer:
Ended up making this into a view model as the logic for the error message isn't trivial.

This isn't complete yet, but I could use a hand 👋


Tests/AppTests/ErrorPageModelTests.swift
```diff
@@ -0,0 +1,41 @@
...
+        let model = ErrorPage.Model(status: status, error: nil)
+
+        // validate
+        XCTAssertEqual(model.errorMessage, "500 - Internal Server Error")
+    }
+
+    func test_500_with_error() throws {
+        // setup
+        let status = HTTPResponseStatus(statusCode: 500)
+        let error: AbortError = nil // I need an AbortError here but don't want to stand up the whole stack just to test the view model.
```
daveverwer, line 32: I really don't want to set up the whole stack here, but I do need to throw (or somehow get hold of) an `AbortError`. Any ideas @finestructure?

finestructure: There’s a function `Abort()` that you should be able to call (but I’m not quite sure without a computer to hand if it returns an `AbortError` or throws)

daveverwer:
Should be ready for review now. Thanks Sven!

Sources/App/Core/ErrorMiddleware.swift
```diff
@@ -11,15 +11,16 @@ public final class ErrorMiddleware: Middleware {
         next.respond(to: req)
             .flatMapError { error in
                 if let abort = error as? AbortError {
-                    return self.handleError(for: req, status: abort.status)
+                    return self.handleError(for: req, status: abort.status, error: abort)
```
finestructure, line 14: Since status is a field on `error` I think we should just pass `error` and make it non-optional in `Model`.

Sources/App/Core/ErrorMiddleware.swift
```diff
@@ -11,15 +11,16 @@ public final class ErrorMiddleware: Middleware {
...
                 } else {
                     return self.handleError(for: req, status: .internalServerError)
```
finestructure, line 16: And here we could then just use

```
return self.handleError(for: req, Abort(.internalServerError))
```

daveverwer:
All fixed up, I think.

PR #234, opened 2020-05-31
Swift language versions

daveverwer:
We should practice what we preach 😂


Package.swift
```diff
@@ -31,5 +31,6 @@ let package = Package(
             "ShellOut",
             "SnapshotTesting"
         ])
-    ]
+    ],
+    swiftLanguageVersions: [.v5]
```
finestructure, line 35: It took me way too long to realise what this is 🤣

PR #239, opened 2020-05-31
Package name camel case

### LINKED ISSUE #121: Standardise on camelCase in Codable responses
Currently, package_name is snake case

Screenshot 2020-05-23 at 10 07 23@2x

finestructure:
Fixes #121


Sources/App/Core/Search.swift
```diff
@@ -25,6 +33,14 @@ enum Search {
...
             case summary
         }
+
+        var asRecord: Record {
+            .init(packageId: packageId,
+                  packageName: packageName,
+                  repositoryName: repositoryName,
+                  repositoryOwner: repositoryOwner,
+                  summary: summary)
+        }
```
finestructure, line 43: So this probably looks a bit convoluted - the issue is that we need to set the `packageName` key to _decode_ as `package_name` because we're also using `Record` to decode the results from the db.

In order to encode it for the API we're then stuck with the snake case unless we map to a distinct type. Which is arguably the cleaner thing to do, although it leads to these pass-through gymnastics.

Another thing we _could_ do is change the search query to set a field alias `package_name AS packageName` with a migration.

This here is probably better, isn't it?

Public/javascripts/search_core.js
```diff
@@ -200,7 +200,7 @@ export class SPISearchCore {
 
     // Repository identifier.
     const repositoryElement = document.createElement('small')
-    repositoryElement.textContent = result.owner + '/' + result.package_name
+    repositoryElement.textContent = result.owner + '/' + result.packageName
```
finestructure, line 203: Is this sufficient to regen the minified versions?

daveverwer: Yea, I'll commit a new minified version on this branch now.

PR #240, opened 2020-05-31
Fail silently when ROLLBAR_TOKEN is not set

finestructure:


finestructure:
The problem is that we don't really have proper means to fail non-silently - I don't see how I can get a logger without a major wiring of `application` instead of `client` through various layers.

I think in prod we'd probably not want to bubble that up to the users, either, do we?

Let me have a quick look if I can get a logger from global state somehow, or introduce one.

daveverwer:
> I think in prod we'd probably not want to bubble that up to the users, either, do we?

Maybe, but without *something* we'll just never know if our Rollbar integration breaks.

This is tougher than it seemed it would be.

daveverwer:
> > I think in prod we'd probably not want to bubble that up to the users, either, do we?
> 
> Maybe, but without _something_ we'll just never know if our Rollbar integration breaks.

But, you're absolutely right... if the user sees this, it's just worse in every way than them seeing the *actual* error.

I think we should fail silently on Rollbar errors in production too. Not reporting it to the user doesn't stop the issue if our Rollbar integration is broken, it just exposes it to the user. I'm convinced.

finestructure:
I think we'll see this by absence of rollbar items and the deployment script will also alert when an env variable isn't set (although it won't fail). Should be good enough, I'd say!

PR #246, opened 2020-06-01
Use in-memory fs for ~20% test speedup

finestructure:
Speeds up tests and avoid leaving behind a volume on disk


daveverwer:
This is a great idea.

PR #248, opened 2020-06-01
Url for

### LINKED ISSUE #172: URL generation from routing
With the Vapor router not being able to generate URLs for routes, we should implement our own URL generation mechanism. It doesn't need to be super generic, or complex, but we do need a *central* source of generating URLs to pages in the site.

For now, this could be as simple as a few static funcrions:

* `urlForHomePage(relative: Bool)`
* `urlForAboutPage`
* `urlForFAQPage`
* `urlForPrivacyPolicyPage(relative: Bool)`
* `urlForPublicPackagePage(relative: Bool)`

The about page, FAQ page, and other functions that don't include a `relative` flag should always be fully qualified URLs as they are going to be off-site links for this launch version. 

Did I miss any URLs that we need to generate?

finestructure:
Fixes #172 

This ended up being quite a bit tricker in the end. Mainly, because I wanted to allow support both for referencing the urls in the DSL (via `.absolutePath` or `.relativePath`) but also in the routing defintion, via `.pathComponents`.

This way there's a single place where we define our Resources now: `Resources.swift`.

I've modified only a few places so far, as an example, but if we decide to go with this I'll go through and do the rest as well.

### DIFF
--- a/Sources/App/Views/PublicPage.swift
+++ b/Sources/App/Views/PublicPage.swift
@@ -111,9 +111,9 @@ class PublicPage {
             .div(
                 .class("inner"),
                 .a(
-                    .href("/"),
+                    .href(Root.home.absolutePath),
                     .h1(
-                        .img(.src("/images/logo.svg")),
+                        .img(.src(Root.images("logo.svg").absolutePath)),
                         "Swift Package Index"
                     )
                 ),
--- a/Sources/App/Views/PublicPage.swift
+++ b/Sources/App/Views/PublicPage.swift
@@ -209,7 +209,7 @@ class PublicPage {
                         ),
                         .li(
                             .a(
-                                .href("/privacy"),
+                                .href(.privacy),
                                 "Privacy and Cookies"
                             )
                         ),
--- a/Sources/App/Views/PublicPage.swift
+++ b/Sources/App/Views/PublicPage.swift
@@ -209,7 +209,7 @@ class PublicPage {
                         ),
                         .li(
                             .a(
-                                .href(.privacy),
+                                .href(Root.privacy.absolutePath),
                                 "Privacy and Cookies"
                             )
                         ),
--- a/Makefile
+++ b/Makefile
@@ -60,7 +60,15 @@ db-up-dev:
 	docker run --name spi_dev -e POSTGRES_DB=spi_dev -e POSTGRES_USER=spi_dev -e POSTGRES_PASSWORD=xxx -p 6432:5432 -d postgres:12.1-alpine
 
 db-up-test:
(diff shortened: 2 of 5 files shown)


Sources/App/Views/PublicPage.swift
```diff
@@ -111,9 +111,9 @@ class PublicPage {
             .div(
                 .class("inner"),
                 .a(
-                    .href("/"),
+                    .href(Root.home.absolutePath),
                     .h1(
-                        .img(.src("/images/logo.svg")),
+                        .img(.src(Root.images("logo.svg").absolutePath)),
```
finestructure, line 116: This is an example of how we can transition other stringly static paths into the scheme, for instance JS & CSS paths.

I'm not 100% in love with the name `Root` for the top level resource but I can't think of anything better right now.

daveverwer: Maybe `SiteURL`?

finestructure:
No worries, no harm done! This should be an easy extension to add the FQDN as well as the other paths and AFAICT the mechanics of it are well served by this change overall, aren’t they?

I’ll add the FQDN etc tomorrow :)

daveverwer:
Yep, absolutely! The structure of it is 👍

I just hate to see *any* wasted work as I was lazy or assumed too much writing it up 😬

Sources/App/Core/AppEnvironment.swift
```diff
@@ -27,7 +28,8 @@ extension AppEnvironment {
             Environment
                 .get("ROLLBAR_LOG_LEVEL")
                 .flatMap(AppError.Level.init(rawValue:)) ?? .critical },
-        shell: .live
+        shell: .live,
+        siteURL: { Environment.get("SITE_URL") ?? "http://localhost:8080" }
```
finestructure, line 32: I've added `SITE_URL` to our Gitlab variables as `http://173.255.229.82:8080` for now.

Sources/App/Core/SiteURL.swift
```diff
@@ -0,0 +1,99 @@
...
+
+
+// MARK: - Types for use in resource declaration
+
+
+protocol Resourceable {
+    var absoluteURL: String { get }
+    var relativeURL: String { get }
+    var path: String { get }
+    var pathComponents: [PathComponent] { get }
```
finestructure, line 80: I've renamed these to better distinguish the "non-rooted" paths we use for the routing, `path` and `pathComponents` vs the absolute/relative to `SITE_URL` urls we construct from that underlyting `path`.

finestructure:
Should be good to go now!

Sources/App/Core/AppEnvironment.swift
daveverwer: Perfect, this is the ideal way to solve this problem. 👍

Issue #177, opened 2020-05-27
Update the search API JSON response to include the package URL


daveverwer:
Once we have a robust way to generate URLs (#172), we should also then supply those generated URLs to the web client via the API response, as [mentioned here](https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/pull/106#discussion_r428603609).

It might be worth tackling #121 at the same time as this.

PR #253, opened 2020-06-03
Owner package urls

finestructure:
Adds owner/repository url support:

Screenshot 2020-06-03 at 07 40 17

I think this works really well and I'll look to surface these everywhere in a follow-up.

This branch is a continuation of #252 . It's best to maybe look at #252 for its change but merge this one here to merge both, because their migrations are related and sit in the directory `002` together. (Although they can be run independenty just fine.)

### DIFF
--- a/Sources/App/Controllers/PackageController.swift
+++ b/Sources/App/Controllers/PackageController.swift
@@ -17,4 +17,15 @@ struct PackageController {
             .map { PackageShow.View($0).document() }
     }
 
+    func _show(req: Request) throws -> EventLoopFuture<HTML> {
+        guard
+            let owner = req.parameters.get("owner"),
+            let repository = req.parameters.get("repository")
+            else {
+                return req.eventLoop.future(error: Abort(.notFound))
+        }
+        return PackageShow.Model.query(database: req.db, owner: owner, repository: repository)
+            .map { PackageShow.View($0).document() }
+    }
+
 }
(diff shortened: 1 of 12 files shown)


Sources/App/Controllers/PackageController.swift
```diff
@@ -17,4 +17,15 @@ struct PackageController {
...
+    func _show(req: Request) throws -> EventLoopFuture<HTML> {
+        guard
+            let owner = req.parameters.get("owner"),
+            let repository = req.parameters.get("repository")
+            else {
+                return req.eventLoop.future(error: Abort(.notFound))
+        }
+        return PackageShow.Model.query(database: req.db, owner: owner, repository: repository)
+            .map { PackageShow.View($0).document() }
+    }
```
finestructure, line 29: In this PR I'm leaving the `/packages/{id}` route in and add the new `_show` and `_packages(_ owner:_ repository)` route with `_` for now.

I've had a quick look at taking the id based lookup out but this will require a set of further changes:

- RecentPackages:
  - update mat. view and model
  - update link generation on homepage
- RecentReleases
  - update mat. view and model
  - update link generation on homepage
- Search links (JS)

This would make for quite a big PR.

It's also a risky PR, because we don't have coverage for all the required the link changes.

daveverwer: That's fine for this PR, but let's consider this change fairly high priority feature as that'll mean those URLs have been public if we preview/launch with that.

Search links JS should be done via #177 which will also centralise the generation of URLs

PR #263, opened 2020-06-03
Keep status as .none until we complete the first full processing pass

### LINKED ISSUE #250: Prioritise new packages for analysis
Follow up of #232

finestructure:
Merge after #259 

Fixes #250

### DIFF
--- a/Sources/App/Commands/Common.swift
+++ b/Sources/App/Commands/Common.swift
@@ -13,8 +13,8 @@ func updatePackage(application: Application,
                     .load(on: application.db)
                     .flatMap { pkg.$versions.load(on: application.db) }
                     .flatMap {
-                        if stage == .ingestion && pkg.status == .none {
-                            // newly ingested package: leave status == .none for fast-track
+                        if stage == .ingestion && pkg.status == .new {
+                            // newly ingested package: leave status == .new for fast-track
                             // analysis
                         } else {
                             pkg.status = .ok
--- a/Sources/App/Commands/Common.swift
+++ b/Sources/App/Commands/Common.swift
@@ -13,7 +13,12 @@ func updatePackage(application: Application,
                     .load(on: application.db)
                     .flatMap { pkg.$versions.load(on: application.db) }
                     .flatMap {
-                        pkg.status = .ok
+                        if stage == .ingestion && pkg.status == .none {
+                            // newly ingested package: leave status == .none for fast-track
+                            // analysis
+                        } else {
+                            pkg.status = .ok
+                        }
                         pkg.processingStage = stage
                         pkg.score = pkg.computeScore()
                         return pkg.update(on: application.db)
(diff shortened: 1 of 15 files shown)


Sources/App/Commands/Common.swift
```diff
@@ -13,7 +13,12 @@ func updatePackage(application: Application,
                     .load(on: application.db)
                     .flatMap { pkg.$versions.load(on: application.db) }
                     .flatMap {
-                        pkg.status = .ok
+                        if stage == .ingestion && pkg.status == .none {
+                            // newly ingested package: leave status == .none for fast-track
+                            // analysis
+                        } else {
+                            pkg.status = .ok
+                        }
```
finestructure, line 21: This is updates the processing order in the simplest way: we keep status a `.none` until a package has passed all processing stages. Our stage selection already prioritises packages with status = `.none`, so this works out nicely and I don't think it's too much of a stretch to interpret status in this way.

The alternative isn't too involved either but would require us to introduce a `.new` state. If we were to do that we should make a schema migration and make `status` non-optional with a default of `.new`. At that point we're essentially simply replacing `.none` (or `NULL`) with `.new` and I'm not really sure that's worth it.

What do you think?

Tests/AppTests/PipelineTests.swift
```diff
@@ -165,7 +162,7 @@ class PipelineTests: AppTestCase {
         do {  // validate - only new package moves to .ingestion stage
             let packages = try Package.query(on: app.db).sort(\.$url).all().wait()
             XCTAssertEqual(packages.map(\.url), ["1", "3", "4"].asGithubUrls)
-            XCTAssertEqual(packages.map(\.status), [.ok, .ok, .ok])
+            XCTAssertEqual(packages.map(\.status), [.ok, .ok, .none])
```
finestructure, line 165: The full pipeline test `test_processing_pipeline` shows nicely how the `status` field progresses and ends up being `.ok` eventually.

Sources/App/Commands/Common.swift
finestructure: One advantage of this change in general, whether we go with `.none` or `.new`, is that we can easily see packages that are pending first full processing. Previously, setting them to `.ok` after ingestion made them indistinguishable from re-ingested packages.

daveverwer: This looks good to me, and makes perfect sense.

If I was looking at this without the debt of an existing database field, it'd be a clear vote for `.new` being a default. It's explicit, and obvious, and that's a good thing.

I'd say there's an argument for changing it. It may be super obvious what it all means now as it's fresh in our minds, but will that still be the case when it's a few months in? I don't think we *have* to do it, but I'd say it's worth considering.

finestructure: Yeah, I agree. The `.none` reads weird in the tests, even knowing what it means now. I'll transition that over!

PR #264, opened 2020-06-03
Smoke test

### LINKED ISSUE #242: Set up a search and package smoke test
- Query for some common terms and check results
- hit a few popular packages
- schedule a run after deployment and regularly

finestructure:
Merge after #263

Fixes #242


finestructure:
If we schedule this, or something like it, to run regularly, we may want to make sure the traffic can be excluded in GA. If there something we can stuff in a header or something?

PR #265, opened 2020-06-04
Migrate Package.status to non-optional, defaulting to .new

finestructure:
This modifies #263 to change `Package.status: Status?` to `Package.status: Status` and adds a new default case `new`.

A bit trickier than expected, because I couldn't see a way to write the migration in Fluent. I'm also glad that we've got our tests, because while the type system guides many things, it does _not_ catch when you compare a non-optional to `.none`. I.e. this here compiles

```
enum Status {
    case ok
}
print(Status.ok == .none)  // false
```

which I found surprising.

They also caught the raw SQL which of course escaped the type system:

```
            .sort(.sql(raw: "status IS NOT NULL"))
```

That all is to say: please take a close look if I missed anything else 😅

### DIFF
--- a/Sources/App/Commands/Common.swift
+++ b/Sources/App/Commands/Common.swift
@@ -13,8 +13,8 @@ func updatePackage(application: Application,
                     .load(on: application.db)
                     .flatMap { pkg.$versions.load(on: application.db) }
                     .flatMap {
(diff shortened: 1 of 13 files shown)


finestructure:
Good to go now

PR #271, opened 2020-06-04
Fix search query

### LINKED ISSUE #269: Fix search query

finestructure:
Fixes #269


finestructure:
Just for posterity: This rewrites the search query in SQLKit, allowing the use of `SQLBind`. There's no real functional change beyond that, except that I've also added a search term cutoff at 20 and dropped `coalesce` from package name, repo owner + name. This means that if these are `NULL`, they won't show up in search. (They are very likely broken.)

PR #277, opened 2020-06-05
Perf test

### LINKED ISSUE #262: Perform load test

finestructure:
Just adding some snippets for performance testing

Fixes #262 

With the new Mac mini I don't think we'll need to be concerned about performance. I've just run a test with 20 random search terms and it served (on the box, to cut out latency) more than 50 requests per second.

A test running random gets against our full package list (`all-packages-test.restfile`) gave 45 req/sec (again directly on the host).

### DIFF
--- a/restfiles/all-packages-test.restfile
+++ b/restfiles/all-packages-test.restfile
@@ -1,4 +1,5 @@
 # auto-generated via `vapor run create-restfile all`
+mode: random
 requests:
   /0111b/JSONDecoder-Keypath:
     url: ${base_url}/0111b/JSONDecoder-Keypath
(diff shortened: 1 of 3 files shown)


.gitlab-ci.yml
```diff
@@ -76,6 +76,8 @@ deploy (ad hoc): &deploy
 
 smoke-test:
   stage: smoke-test
+  only:
+    - tags
```
finestructure, line 80: Running the smoke test only after deployments

restfiles/perf-test-search.restfile
```diff
@@ -0,0 +1,124 @@
+# run as follows
+# env base_url=... rester perf-test-search.restfile --count 100 --stats
+mode: random
```
finestructure, line 3: Rester can be run in a repeat loop and with `mode: random` it will pick a random request from the list.

The request list below is just 20 random search terms.

restfiles/all-packages-test.restfile
```diff
@@ -5928,6 +5929,10 @@ requests:
     url: ${base_url}/jhoogstraat/StandardRouting
     validation:
       status: 200
+  /jjjkkkjjj/Matft:
```
daveverwer, line 5932: The benefits of being the latest app added to the index is that you get to be the test data! 😂

PR #280, opened 2020-06-05
Package url in search results

### LINKED ISSUE #177: Update the search API JSON response to include the package URL

finestructure:
Fixes #177


daveverwer:
Tagging this version to deploy everything.

PR #286, opened 2020-06-06
Fix url escaping for all urls, not just search

finestructure:
I just realised while working on the RSS fees that we need to escape the package urls in general and that this should have been in `SiteURL` to begin with.

### DIFF
--- a/Sources/App/Core/SiteURL.swift
+++ b/Sources/App/Core/SiteURL.swift
@@ -45,8 +45,11 @@ enum SiteURL: Resourceable {
             case let .images(name):
                 return "images/\(name)"
 
-            case let .package(.value(owner), .value(repository)):
-                return "\(owner)/\(repository)"
+            case let .package(.value(owner), .value(repo)):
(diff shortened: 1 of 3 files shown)


Sources/App/Core/SiteURL.swift
```diff
@@ -45,8 +45,11 @@ enum SiteURL: Resourceable {
             case let .images(name):
                 return "images/\(name)"
 
-            case let .package(.value(owner), .value(repository)):
-                return "\(owner)/\(repository)"
+            case let .package(.value(owner), .value(repo)):
+                let owner = owner.addingPercentEncoding(withAllowedCharacters: .urlPathAllowed) ?? owner
+                let repo = repo.addingPercentEncoding(withAllowedCharacters: .urlPathAllowed) ?? repo
+                return "\(owner)/\(repo)"
```
finestructure, line 51: I don't think we want to fail here if `addingPercentEncoding` returns `nil`, so I'm just falling back to the original value.

PR #297, opened 2020-06-06
Update to self-hosted images

daveverwer:


Resources/Markdown/faq.md
```diff
@@ -2,7 +2,7 @@
 
 The Swift Package Index is a search engine for packages that support the Swift Package Manager.
 
-But it’s about more than just indexing packages because choosing the right package is about more than just finding code that does what you need. Is it well maintained? What versions of Swift does it support? Is it well tested? How long has it been in development? **Picking high-quality packages is hard**, and *that’s* where the Swift Package Index aims to be most useful.
+But it's about more than just indexing packages because choosing the right package is about more than just finding code that does what you need. Is it well maintained? What versions of Swift does it support? Is it well tested? How long has it been in development? **Picking high-quality packages is hard**, and *that's* where the Swift Package Index aims to be most useful.
```
finestructure, line 5: 😂

daveverwer: I'm 100% with you on this... On my blog, it automatically converts them so I never even look 😂

daveverwer: We should submit a PR to Ink to make it convert them.

finestructure: Oh, that’s interesting. Because I noticed the other day that iOSDevWeekly’s webpage has the curly quotes but the emails (often) don’t. Is that why?

daveverwer: Ah no, Curated doesn't replace them. I wish it did... I was talking about https://daveverwer.com

iOS Dev Weekly has a mix of standard and curly... don't look too closely or you'll never be able to read it again. 😂

PR #319, opened 2020-06-07
Add "Recent Releases" RSS feed

### LINKED ISSUE #64: Add two RSS feeds

finestructure:
+ some fixes for "Recently Added" RSS feed

Fixes #64


finestructure:
Good to go!

PR #322, opened 2020-06-07
Rss fixes

finestructure:
I think that should fix the links - I'll deploy ad hoc to test

That fixed it:

837F4B3D-0C77-4AA4-8CEF-C4D1C86122C8

The issue was that while I had set `SITE_URL` in the secrets, I hadn’t actually wired it through to the containers in `docker-compose.yml`


finestructure:
Still getting an error when trying to subscribe via Reeder though :(

finestructure:
Odd, packages showed up a moment later despite the error, but not releases, even after retry.

4EBF02FA-FE96-415A-AD52-F1BEEEA534C9
5DA23524-7362-4F4E-9C4A-38638D824189

finestructure:
List seems to be out of order though - something isn’t right here.

daveverwer:
So the packages one looks like it's working, and the items in it seem sensible from what I can see.

The releases.rss is a 404 for me. No RSS content at all, just HTML as a 404.

finestructure:
Well, it kinda helps to add the route 🤪

I’ll go over the whole thing again tomorrow though - I was way too unfocused today 😅

finestructure:
I am a bit at a loss what's going on here. The feeds validate and can subscribe to them without error now. However, what I see has virtually nothing to do with the most recent items in our respective lists:

Screenshot 2020-06-08 at 09 14 34
Screenshot 2020-06-08 at 09 14 18

My best and only guess is that my client is in a weird state, because it pulled in the feeds in random order before I added `pubDate`. Although even that shouldn't have messed up the order, because there was order in the feed.

I've tried deleting and re-adding the feeds but no difference.

finestructure:
What feed content do you see for the deployed version (83b5a760eb223039c04a9f3f074cb1e0ff51c601) which is what the above is referring to?

PR #331, opened 2020-06-08
Dedupe recents

### LINKED ISSUE #315: Dedupe recent packages in case of package name changes
Example:

Screenshot 2020-06-07 at 17 05 06

finestructure:
Fixes #315 

I came a across a better of doing this "greatest N per group" type of query without having to do lots of `group by` gymnastics (which is fragile as colums tend to be added):

https://stackoverflow.com/a/9795768/1444152

### DIFF
--- a/Sources/App/Migrations/003/UpdateRecentPackages1.swift
+++ b/Sources/App/Migrations/003/UpdateRecentPackages1.swift
@@ -11,6 +11,7 @@ struct UpdateRecentPackages1: Migration {
         }
         let updatedViewSQL: SQLQueryString =
             """
+            -- v1
             CREATE MATERIALIZED VIEW recent_packages AS
             SELECT
                 p.id,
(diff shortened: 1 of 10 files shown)


Sources/App/Migrations/003/UpdateRecentPackages1.swift
```diff
@@ -38,18 +39,16 @@ struct UpdateRecentPackages1: Migration {
...
-                MAX(p.created_at) AS created_at
+              p.id,
+              v.package_name,
+              MAX(p.created_at) AS created_at
             FROM packages p
             JOIN versions v ON v.package_id = p.id
-            JOIN repositories r ON r.package_id = p.id
             WHERE v.package_name IS NOT NULL
-            GROUP BY p.id, r.owner, r.name, v.package_name
+            GROUP BY p.id, v.package_name
```
finestructure, line 51: While labeling the different versions for a better overview I noticed that our revert from v1 -> v0 actually reverted to the wrong query. (Not that it's a big problem with these caching views.)

PR #333, opened 2020-06-08
Make hour time differences less verbose

### LINKED ISSUE #313: Remove the word "about" from recent releases and recent packages
Based on feedback:

> "Released about 4 hours ago" can just be "Released 4 hours ago". I don't think anyone will somehow complain that you were off by 20 minutes, and it just removes reading "about" seven times on that page.

Absolutely 👍

Also from the RSS feed, and anywhere else we're talking about times.

finestructure:
It'd be quite involved to remove about only for the recent releases so I've trimmed it in our generic time delta function.

Fixes #313


daveverwer:
Yea, absolutely. It's never important on a site like this. "About" is implicit.

PR #335, opened 2020-06-08
Http error codes

### LINKED ISSUE #323: Error pages should return the correct status code
Googles really don't like it when sites return 200 for 404/500 errors.

Screenshot 2020-06-07 at 20 22 49@2x

finestructure:
Fixes #323


daveverwer:
From the commit message, is this *only* for 4xx/5xx codes? From the code, it doesn't look like it, but the commit message says it.

finestructure:
Yeah, that's misleading - we're passing through _any_ error code that we're getting.

Issue #74, opened 2020-05-17
Add a JavaScript copy button to the URL on the public package page.


daveverwer:
Leaving this as a reminder I still need to do it. I haven't added any JavaScript yet, but will.

PR #342, opened 2020-06-09
Fix 74

daveverwer:
I use JavaScript to add the button intentionally. Because it's a JavaScript **only** button, it's added via JavaScript. If JavaScript isn't working for some reason, the button won't even be added.

Tested in Safari, Chrome, and FireFox. It does require Safari 13.1 (it's not even available in Safari 13.0) which was only released at the end of March. However,  I don't have a lot of patience with people who don't keep Safari up to date so I'm not coping gracefully with older versions of Safari until people complain. 😂 It's been in Chrome for ~2 years, so plenty of time there. 👍

Here it is in action:

Screenshot 2020-06-09 at 01 08 11


finestructure:
Nice!

finestructure:
Fixed #74

Issue #343, opened 2020-06-09
Avoid 404s during version reconciliation


finestructure:
The [last smoke test](https://gitlab.com/finestructure/swiftpackageindex/-/jobs/586574984) picked up a few 404s:

- https://swiftpackageindex.com/BenEmdon/CenteredCollectionView (still 404)
- https://swiftpackageindex.com/GEOSwift/GEOSwift (now 200)
- https://swiftpackageindex.com/Moya/Moya (now 200)
- https://swiftpackageindex.com/ashleymills/Reachability.swift (now 200)
- https://swiftpackageindex.com/llvm-swift/LLVMSwift (now 200)
- https://swiftpackageindex.com/pointfreeco/swift-tagged (now 200)
- https://swiftpackageindex.com/square/Valet (now 200)
- https://swiftpackageindex.com/stephencelis/SQLite.swift (now 200)

finestructure:
A [rerun](https://gitlab.com/finestructure/swiftpackageindex/-/jobs/586893074) triggers 404 on the same https://swiftpackageindex.com/BenEmdon/CenteredCollectionView and then 6 other different urls.

Weird.

daveverwer:
Intermittent 404s isn't a great sign. I can't think what could be causing that.

daveverwer:
https://github.com/BenEmdon/CenteredCollectionView is a valid package too. It also doesn't show up in search results.

finestructure:
The good news is we know exactly when it started, because that test has been running since 3c6510c0d2e33565932ba5832a78af20d46f66ea 5 days ago, over many releases. It must be something that came with last night's 0.4.3 release.

finestructure:
I wonder if it's the ErrorMiddleware change. I'll try and repro locally and see if a roll back fixes it.

daveverwer:
That's certainly the most likely of yesterdays change to cause this behaviour.

But that wouldn't explain why BenEmdon/CenteredCollectionView is being excluded from search results. I just checked on SwiftPM.co to see if that struggled, but it's fine. https://swiftpm.co/?query=CenteredCollectionView

finestructure:
That is very likely due to the query rejecting it because of a nil value. I suspect that's a separate issue.

finestructure:
I can repro everything locally, just a matter of debugging through now.

finestructure:
So, CenteredCollectionView has no versions in the db, which leads to

```
[...]

PR #347, opened 2020-06-09
Wrap version update in transaction

finestructure:
Not marking the issue #343 as fixed yet.

In order to confirm we'll need to run a smoke test after this has been deployed and there's been a full pass across the dataset. Only then can we be certain that any version gaps have been filled.

I suspect CenteredCollectionView was in the midst of an update as we rolled out a version. That's why we didn't have versions for a longer time until it came around in the next pass.

This fix here would prevent that from happening or us seeing a gap while the update is in progress.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,7 +10,7 @@ jobs:
       image: swift:5.2.4-bionic
     services:
       postgres:
-        image: postgres:12
+        image: postgres:12.1-alpine
         env:
           POSTGRES_DB: spi_test
           POSTGRES_USER: spi_test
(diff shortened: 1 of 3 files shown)


daveverwer:
I think the tests here might have stalled, they've been running for almost 30 minutes.

finestructure:
Mmm, not good. I'll try the docker (linux) run locally. I hope it's not a platform specific issue with the transaction.

finestructure:
`make docker-test` passed several times locally but CI seems to be stalling again :/

finestructure:
Well, interesting!

There was a bug in the transaction wrapper: I forgot to replace one of the inner `application.db` parameters with the transaction, meaning I was mixing tx contexts. The CI actually _helpfully_ stalled here and now I'm baffled that over more than a dozen runs across macOS and Linux on my machine it didn't once deadlock.

PR #352, opened 2020-06-09
Robots txt

### LINKED ISSUE #339: Add a Robots file
The server should serve a plain text file as a 200 as `/robots.txt`

Content something like this:

```
# All robots allowed
User-agent: *
Disallow:

# Sitemap files
Sitemap: https://swiftpackageindex.com/sitemap.xml
```

That allows every agent to crawl every page it can find, and shows them where to find the sitemap

finestructure:
Fixes #339

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,3 +1,7 @@
 .build/
 .swiftpm/
 SPI-checkouts/
+Public/build
+Public/node_modules
+Public/.config.codekit3
+Public/package*.json
--- a/Public/robots.txt
+++ b/Public/robots.txt
@@ -1,0 +1,6 @@
+# All robots allowed
+User-agent: *
+Disallow:
+
+# Sitemap files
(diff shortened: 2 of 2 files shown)


.dockerignore
```diff
@@ -1,3 +1,7 @@
 .build/
 .swiftpm/
 SPI-checkouts/
+Public/build
+Public/node_modules
+Public/.config.codekit3
+Public/package*.json
```
finestructure, line 7: We're currently serving these files: https://swiftpackageindex.com/package.json but they're just inputs for CodeKit. Adding them here will make sure they're skipped during

```
COPY --from=build /build/Public /run/Public
```

I _think_ this is all we need/want to skip but please double check!

daveverwer: You're absolutely right, none of those are needed. I just checked and everything else in the Public folder *is* needed. I really wish we could split these up a bit, but it wasn't easy to do with CodeKit.

PR #357, opened 2020-06-10
RSS feed formatting changes

daveverwer:
I also added the summary into the tests as they were not showing up on my local feeds and it wasn't being tested.

### DIFF
--- a/Sources/App/Core/RSS.swift
+++ b/Sources/App/Core/RSS.swift
@@ -62,10 +62,20 @@ extension RecentPackage {
             .title(packageName),
             .link(link),
             .pubDate(createdAt, timeZone: .utc),
-            .content(
-                .h2(.a(.href(link), .text(packageName))),
+            .description(
+                .p(
+                    .a(
+                        .href(link),
+                        .text(packageName)
+                    )
+                ),
                 .p(.text(packageSummary ?? "")),
-                .element(named: "small", nodes: [.a(.href(link), .text(packageName))])
+                .element(named: "small", nodes: [
+                    .a(
+                        .href(link),
+                        .text(packageName)
+                    )
+                ])
             )
         )
     }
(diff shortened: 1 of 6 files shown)


Sources/App/Core/RSS.swift
```diff
@@ -62,10 +62,20 @@ extension RecentPackage {
...
+                        .href(link),
+                        .text(packageName)
+                    )
+                ),
                 .p(.text(packageSummary ?? "")),
-                .element(named: "small", nodes: [.a(.href(link), .text(packageName))])
+                .element(named: "small", nodes: [
+                    .a(
+                        .href(link),
+                        .text("\(repositoryOwner)/\(repositoryName)")
```
finestructure, line 76: Is this the lack of a leading `/` intentional? Since this is a link I think we should use one of

```
SiteURL.package(.value(repositoryOwner), .value(repositoryName)).absoluteURL()
SiteURL.package(.value(repositoryOwner), .value(repositoryName)).relativeURL()
SiteURL.package(.value(repositoryOwner), .value(repositoryName)).path
```

so in case of the above it'd be `.path`.

Sources/App/Core/RSS.swift
```diff
@@ -78,14 +88,33 @@ extension RecentRelease {
...
+                        .text(packageName)
+                    ),
+                    .element(named: "small", text: " – \(version)")
+                ),
                 .p(.text(packageSummary ?? "")),
-                .element(named: "small", nodes: [.a(.href(link), .text(packageName))])
+                .element(named: "small", nodes: [
+                    .a(
+                        .href(link),
+                        .text("\(repositoryOwner)/\(repositoryName)")
```
finestructure, line 106: See above

Sources/App/Core/RSS.swift
finestructure: Oh wait, this is the text part - never mind

daveverwer: Yep, intentional. It was supposed to look like the [owner]/[repo] that people might recognise from GitHub, rather than being a URL.

Like this:

Screenshot 2020-06-10 at 09 50 34@2x

PR #358, opened 2020-06-10
Stars

### LINKED ISSUE #150: Use a checkmark icon instead of the envelope icon on the Package page
I forget where you suggested this @finestructure, but I liked the suggestion!

### LINKED ISSUE #338: Include the number of stars as top level metadata
How on earth did this not cross my mind when I was designing the new package page?!? I must be losing it.

The bullet point should be in the top section, and let's put it at the bottom, after all the others. The text will be short and it'll look good at the bottom. We can always move it later. The bullet should simply say:

* XXX stars.

I'll knock an icon together for this before launch, and also tackle #150 at the same time. @finestructure, if you could handle the bringing of the data through from the query into the HTML that'd be much appreciated!

finestructure:
Fixes #338 
Fixes #150


Public/stylesheets/main.css
```diff
@@ -187,6 +187,7 @@ li.author:before { content: "\e901"; }
 li.history:before { content: "\e902" }
 li.activity:before { content: "\e903"; }
 li.products:before { content: "\e904"; }
+li.stars:before { content: "\e904"; }
```
finestructure, line 190: I'm not sure how the icon is being referenced - I've kept it referecing the product one.

Leaving the PR as draft so we can add in the icon and change this reference here - that's all it should take. Snapshots should be unaffected since they already use the correct class and don't display the actual icon in the screenshots.

Sources/App/Views/PackageController/PackageShow+Model.swift
```diff
@@ -1,18 +1,20 @@
...
-        var activity: Activity?
+        var languagePlatforms: LanguagePlatformInfo
+        var license: License
         var products: ProductCounts?
         var releases: ReleaseInfo
-        var languagePlatforms: LanguagePlatformInfo
+        var stars: Int?
+        var summary: String
+        var title: String
+        var url: String
```
finestructure, line 17: Sorry for the commit noise - I've now ordered the properties alphabetically.

daveverwer:
I’ll pick this up and add the new icon now.

PR #364, opened 2020-06-10
Dark mode image replacement and FAQ images

daveverwer:
Ask and ye shall receive!


daveverwer:
It even watches for colour scheme changes live and replaces. 🎉

Screenshot 2020-06-10 at 23 46 20

I deliberately changed the background colour of the dark mode screenshots so they stand out slightly from the background of the page as you can't see the drop shadow I added in light mode.

finestructure:
Fantastic :)

We'll probably need some sort of link checking so we don't accidentally forget to ship a dark image variant, right?

PR #366, opened 2020-06-11
Bump INGEST_SLEEP to 600

finestructure:


finestructure:
I can't self-approve this so I think I'll disable approvals briefly as it's pretty early in the UK...

finestructure:
Well, of course we're burning through our request budget at twice the rate...

🤦‍♂️

The Linode box is running with the same API key.

PR #386, opened 2020-06-12
Added extra instructions for a 404 page.

### LINKED ISSUE #385: Include an additional messaging on the error page
On 404 errors, a message saying to come back later as it might just be indexing.

daveverwer:
Fixes #385


Sources/App/Views/Error/ErrorPage+View.swift
```diff
@@ -18,7 +18,10 @@ enum ErrorPage {
                 ),
                 // Note: The copy in this header tag intentionally matches the copy in `search_core.js`.
                 .h4("Something went wrong. Sorry!"),
-                .p(.text(model.errorMessage))
+                .p(.text(model.errorMessage)),
+                .if(model.errorInstructions.count > 0,
```
daveverwer, line 22: I thought this was a little cleaner than having this be optional just because we don't have easy access to if let here and so if it's optional, I also needed to add a nil coalesce on the next line down.

daveverwer:
Screenshot to save you running it.

Screenshot 2020-06-12 at 08 00 13@2x

Sources/App/Views/Error/ErrorPage+View.swift
finestructure, line 22: `.if(!.model.errorInstructurions.isEmtpy)`?

finestructure: There is `.unwrap` though

finestructure: i.e. that could be (with an optional),

```
.wrap(model.errorInstructions) { .p(.text($0)) }
```

daveverwer: Yep, fixing it up.

PR #398, opened 2020-06-13
Fix broken link to Rester in README

iainsmith:
Noticed this while browsing, so here's a quick trivial fix.


finestructure:
Thanks Iain!

PR #400, opened 2020-06-13
Stats view

finestructure:
I couldn't help myself 🤣

Screenshot 2020-06-13 at 17 50 29


Tests/AppTests/Mocks/HomeIndexModel+mock.swift
```diff
@@ -4,6 +4,7 @@
 extension HomeIndex.Model {
     static var mock: HomeIndex.Model {
         .init(
+            stats: .init(packageCount: 2544, versionCount: 38410),
```
finestructure, line 7: This line here ensures we recreate the exact same snapshot as before. It's a good test to change the numbers here and see the snapshot test fail.

Sources/App/Commands/Analyzer.swift
```diff
@@ -87,6 +87,7 @@ func analyze(application: Application, packages: EventLoopFuture<[Package]>) ->
         .flatMap { RecentPackage.refresh(on: application.db) }
         .flatMap { RecentRelease.refresh(on: application.db) }
         .flatMap { Search.refresh(on: application.db) }
+        .flatMap { Stats.refresh(on: application.db) }
```
daveverwer, line 90: How much of a load do these refreshes put on the box? I know we've got *loads* of headroom right now, and I know that's a hard question to answer (I don't expect an answer) but it's worth keeping an eye on.

PR #403, opened 2020-06-14
Re-use stats for page description on home page

finestructure:


Sources/App/Views/PublicPage.swift
```diff
@@ -120,8 +120,7 @@ class PublicPage {
     final func description(maxLength: Int = 200) -> String {
         guard let pageDescription = pageDescription() else {
             return """
-            The Swift Package Index is the place to find the best Swift packages. \
-            Indexing 2,482 packages and 36,483 versions.
+            The Swift Package Index is the place to find the best Swift packages.
```
finestructure, line 123: I've had to move this to `HomeIndex`, because we don't have a model available in the generic class. We could make the `request` available and thereby get a `db` with which to query but I think we only really care about the stats for links to the home page anyway.

We can add the stats to other pages that have models quite easily as well but I've left it at that.

Tests/AppTests/__Snapshots__/WebpageSnapshotTests/test_HomeIndexView.1.txt
```diff
@@ -9,9 +9,9 @@
     <title>Swift Package Index</title>
     <meta name="twitter:title" content="Swift Package Index"/>
     <meta name="og:title" content="Swift Package Index"/>
-    <meta name="description" content="The Swift Package Index is the place to find the best Swift packages. Indexing 2,482 packages and 36,483 versions."/>
-    <meta name="twitter:description" content="The Swift Package Index is the place to find the best Swift packages. Indexing 2,482 packages and 36,483 versions."/>
-    <meta name="og:description" content="The Swift Package Index is the place to find the best Swift packages. Indexing 2,482 packages and 36,483 versions."/>
+    <meta name="description" content="The Swift Package Index is the place to find the best Swift packages. Indexing 2,544 packages and 38,410 versions."/>
+    <meta name="twitter:description" content="The Swift Package Index is the place to find the best Swift packages. Indexing 2,544 packages and 38,410 versions."/>
+    <meta name="og:description" content="The Swift Package Index is the place to find the best Swift packages. Indexing 2,544 packages and 38,410 versions."/>
```
finestructure, line 14: Number changed because now they are taken from the mock and not hard-coded anymore.

Tests/AppTests/__Snapshots__/WebpageSnapshotTests/test_ErrorPageView.1.txt
```diff
@@ -9,9 +9,9 @@
     <title>Swift Package Index</title>
     <meta name="twitter:title" content="Swift Package Index"/>
     <meta name="og:title" content="Swift Package Index"/>
-    <meta name="description" content="The Swift Package Index is the place to find the best Swift packages. Indexing 2,482 packages and 36,483 versions."/>
-    <meta name="twitter:description" content="The Swift Package Index is the place to find the best Swift packages. Indexing 2,482 packages and 36,483 versions."/>
-    <meta name="og:description" content="The Swift Package Index is the place to find the best Swift packages. Indexing 2,482 packages and 36,483 versions."/>
+    <meta name="description" content="The Swift Package Index is the place to find the best Swift packages."/>
+    <meta name="twitter:description" content="The Swift Package Index is the place to find the best Swift packages."/>
+    <meta name="og:description" content="The Swift Package Index is the place to find the best Swift packages."/>
```
finestructure, line 14: The error page is a page where we lose the `meta` but we're probably not linking to those 😆

Sources/App/Views/PublicPage.swift
daveverwer: Yea that works. Thanks

PR #405, opened 2020-06-14
Db backup

### LINKED ISSUE #346: Auto-snapshot the database
Every 24 hours, let's automate a snapshot of the database that we can easily then pull down to local development machines. We should stop the analyser, snapshot, and restart it. 4 am seems like a good time to do it, nice and out of the way.

We should also have another script that (in addition to what we already have) can easily grab the latest snapshot and install it. I'd run that every morning.

finestructure:
We can now manually trigger db backups that ship to the iCloud shared folder and there's also a nightly job, running at 6am UTC: https://gitlab.com/finestructure/swiftpackageindex/-/pipeline_schedules

Fixes #346

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -97,13 +97,13 @@ smoke-test:
 
 db-backup:
   stage: scheduled
-  only:
-    - schedules
+  # only:
+  #   - schedules
   when: manual
   tags:
     - ceres
   script: |
-    FILENAME=spi_prod_$(date +%Y-%m-%d).dump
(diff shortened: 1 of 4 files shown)


.gitlab-ci.yml
```diff
@@ -69,20 +72,49 @@ deploy:
...
+    - schedules
   
   
-deploy (ad hoc): &deploy
-  <<: *deploy-common
-  when: manual
+# deploy (ad hoc): &deploy
+#   <<: *deploy-common
+#   when: manual
+
```
finestructure, line 82: I've also ad-hoc deployment for now. I'll bring that back once we have a staging env and then I'll enable it there (and only there).

PR #407, opened 2020-06-15
Search Input Delay

maxxfrazer:
- Using debounce with a delay of 200ms on input event listener

Currently the search runs every time a letter is added, and it looks like the 300 value on the same line was meant to be part of debounce - I've left it in until you can confirm it is or isn't for that.

I've compiled the project to create the `.min` files locally using CodeKit; but many files were changed, including image binaries(?). Let me know how to complete this PR if you do want the delay between adding chars to the search box!


daveverwer:
Thank you, Max! This was definitely working at some point! 😬 Thanks for noticing it, and for the fix.

PR #417, opened 2020-06-18
Rss improvements

### LINKED ISSUE #360: RSS feed granularity
I would love to read a feed that included every major and minor feed release, but I don't care much about patch releases. We should have a query parameter on the releases feed that allows this:

* `https://swiftpackageindex.com/releases.rss` - All releases
* `https://swiftpackageindex.com/releases.rss?granularity=minor` - All major (x) and minor (x.x) releases, exclude patch (x.x.x) releases.
* `https://swiftpackageindex.com/releases.rss?granularity=major` - All major (x) releases, exlude minor (x.x) and patch (x.x.x) releases.

We should add two more autodiscovery meta tags for the two new feeds too.

finestructure:
Fixes #360 

I'll keep this in draft for now so that we can deploy it to `dev` once that env is up (hopefully tomorrow, if the box comes back online).


daveverwer:
Just a quick one, I mentioned it in chat but I don't think it got through due to ... notifications. I have a feeling we'll need a 4th boolean parameter 😬 for beta/pre versions.

`major`
`minor`
`patch`
`pre`

and again, all parameters should default to true so:

* `/releases.rss` = everything
* `/releases.rss?major=true` also everything (defaults other parameters to true)
* `/releases.rss?major=true&minor=false&patch=true&pre=false` only major and patch versions.

finestructure:
Noooooo 🤣

finestructure:
The way I've done the parameters now is:

- none -> all
- default is false
- true sets that flag only

I think it'd be weird to have `major=true` and have it do nothing. If we want to do that, I think we'll want it the other way around, default all to true and use `major=false` to suppress.

daveverwer:
Yea, sorry! I think what I'm actually going to do is subscribe to 5 different feeds.

* Newly added packages
* New major releases
* New minor releases
* New patch releases
* New pre releases

I want to be in a different frame of mind when looking at each of those lists, major I'm going to read through carefully and potentially visit the package page. Minor I'll skim through. Patch I'll skim quickly. Pre I'll actually just look for potentially interesting packages that are starting out a beta cycle.

daveverwer:
> The way I've done the parameters now is:
> 
> * none -> all
> * default is false
> * true sets that flag only
> 
> I think it'd be weird to have `major=true` and have it do nothing. If we want to do that, I think we'll want it the other way around, default all to true and use `major=false` to suppress.

That works for me!

finestructure:
I'll deploy this revision to staging in a moment. Turned out to be quite a hefty change in the end 😅

daveverwer:
I've just been through all of my test feed subscriptions from yesterday and everything looks perfect. I'll merge and deploy.

finestructure:
Yay! 🎉

PR #424, opened 2020-06-20
Always build the sha

finestructure:
The deployment just now (1.1.5) didn't go through because the tag pipeline ended up coming before the build - and therefore the re-tagging failed.

This ensure that we're always building the revision so that tags can't fail if they are scheduled before the corresponding `build-sha` has completed. While it does a duplicate build for that tag revision it's just safer.

I've tr-triggered the deployment for 1.1.5 and it should go through shortly.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -141,4 +141,5 @@ db-backup (manual):     # PROD: allow manual backup any time
   rules:
     - if: '$ENV == "prod"'
       when: manual
+  allow_failure: true
   <<: *backup-common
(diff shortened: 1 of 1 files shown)


daveverwer:
I'll let you merge and re-tag this.

finestructure:
I'll just merge, no need for another deploy really.

finestructure:
Needed another tweak: staging deploys also tried to access the image before it was built. I've added a 10 min retry loop to ensure it doesn't fail if the build isn't done yet.

PR #428, opened 2020-06-20
Split license lozenge colours up to be more granular

daveverwer:
There are situations where having GPL licensed code is not as serious a problem as it is when shipping package code to the App Store. 

There's no issue open for this, but I felt that while the colour of the lozenge should definitely be red for "No License" or "Unknown License" situations, for GPL it should be orange.


Sources/App/Models/License.swift
```diff
@@ -112,20 +112,26 @@ enum License: String, Codable, Equatable {
...
-                 .lgpl_3_0: return false
-            case .other,
-                 .none: return false
-            default: return true
+                 .lgpl_3_0: return .incompatibleWithAppStore
+            default: return .compatibleWithAppStore
         }
     }
+
+    enum Kind: String {
```
daveverwer, line 130: Feels a bit weird having a type called `Kind`, but `License.LicenseKind` also felt weird.

finestructure: We could use ``enum `Type`: String {``

finestructure: I can fix that up if you're busy with the metada threads, just le me know :)

PR #438, opened 2020-06-21
Add release cadence estimate to package history

### LINKED ISSUE #371: Include a cadence for releases
How about a bit of extra information at the end of the releases clause:

Currently:

* In development for 5 years, with 1,448 commits and 81 releases.

How about including a cadence, like this:

* In development for 5 years, with 1,448 commits and 81 releases. The average time between releases is approximately 10 days.

Sherlouk:
As per issue #371 I've added an approximate time between releases.

Wondering though if the cadence should be averaged between first commit and now, or first commit and latest release. Both have their merits IMO.

Looked like an easy issue though and wanted to take a look into the project - if you'd rather pass on it then no probs!

Closes #371


daveverwer:
What a surprise this is! Thanks James!

I'll check this out locally over the next couple of days, I'd like to see how it looks with real data to see if it's useful information.

Sherlouk:
Aye not a problem - if you decide this isn't worthwhile then feel free to close absolutely no worries, was a great excuse to get everything downloaded and setup so I could start playing around!

Thanks for taking a look Dave

finestructure:
Great job digging through the whole stack and setting this up, James! 🎉

daveverwer:
Alright, I just had a play with this locally using a clone of the production database dump and it's working well. That said, I'm not sure it's beneficial to the goal of the project, to let people make better decisions about their dependencies. That's not your fault at all, it's mine! 👍 It's worth discussing though.

Here are some results:

## Alamofire

This is always my go-to package to test with as it’s extremely well maintained. This clause works reasonably here, the last release was about a month ago, and it confirms that they release about every month.

localhost_8080_Alamofire_Alamofire

## HorizonCalendar

A new package, rapidly iterating at the moment. 3 releases in 4 days! It correctly gets the average at about a day, but is this useful? I’m not sure.

localhost_8080_airbnb_HorizonCalendar

## SofarKit

An abandoned package. The description says [WIP], the latest release was 2 years ago and it was v0.1.0.

localhost_8080_dreymonde_SofarKit

The clause for this package is actually quite misleading. Look at the releases:

github com_dreymonde_sofarkit_releases

4 releases in 5 days, then nothing for almost two years. Yes, 5 months is an average, but that’s not a point in its favour I don’t think.

---

I tested it with packages that have no releases and it does not include the clause, which is good. If we do go ahead with it though, I’d also want to see the clause omitted if there has only been one release.

---

The other change I think we should make if we go ahead with it is to move the clause. Having it in that top bullet point made sense when I came up with the idea, but having it be so much more prominent than the actual release data feels completely wrong. If we go ahead, I think it should go as a new sentence, just below the list of releases. It would make more sense there.

I’d be genuinely interested in your thoughts on this. Let’s discuss the pros and cons.

daveverwer:
I also tested it with packages that have no releases and it does not include the clause, which is good. If we do go ahead with it though, I’d also want to see the clause omitted if there has only been one release.

Sherlouk:
So yeah I understand wanting to hide it if they've only had one release as that's not really a big enough sample size.

If it came out as its own bullet point, we could do something a bit different to better handle deprecated libraries. If we did something like:
> The average time between releases is approximately 5 days. The last release was about 3 months ago.

The difference here is the time between releases would be between the first commit's date (or even the first release's date?) and the latest release's date and work out the average from that. A user should be able to then understand if the latest release is a lot longer than their typical cadence there is likely something wrong - and unless the library has had some sort of hiatus the cadence wouldn't ever change for a deprecated library which seems to make sense to me.

I also think the wording on this is difficult because the "approximately about 1 month" example for Alamofire does not read well...

I do think it's worth asking the question of what are we trying to tell the user? It's not like I'm looking at a cadence and adding next weeks potential release to my calendar.. it seems more that we're trying to tell users if either a package is no longer being maintained or if a package is being rapidly developed (usually representative of an new and unstable library, a big no no for some). Do we think a sentence is the best way of showing this, as opposed to maybe a badge/icon?

I do think it's worth asking the question of what is this data telling the user? It's not like we're trying to predict the future here but more so:
* I want to know if a package is no longer being maintained (not super clear currently)
* I want to know if a package is slow to adopt new swift versions (won't be solved by any of this)
* I want to know if a package is rapidly changing (typical of a new library)

Could we display this data in a more visual way?

daveverwer:
Moving this to discussion on the issue itself #371 as that feels more permanent than a PR

daveverwer:
In terms of this PR. Is it worth putting this in its own sentence down under the releases, while we talk about something a little better, or just closing this?

My gut feeling says that this is more misleading when it goes wrong than it is beneficial when it goes right, so I'd vote for closing this PR unmerged.

Sherlouk:
Closing until we can agree an approach as per above comments

PR #448, opened 2020-06-22
Package update

finestructure:
This drops the dependency of having the vapor toolbox installed to run the Makefile targets. The toolbox is convenient but unnecessary for what we do.

Also bump the packages to some new release versions.


Sherlouk:
Aha this is awesome @finestructure - it was something I had to mess with to get things working yesterday. It's worth noting there are still references to `vapor-beta` in the `scripts/` directory -- can those be replaced safely too?

finestructure:
Ah yes, you‘re right! Those need to change as well - thanks :)

PR #449, opened 2020-06-22
Reconcile versions

### LINKED ISSUE #422: Reconcile versions
We’re currently deleting and recreating all of a package’s versions on each analysis pass for sake of simplicity.

This is not terribly efficient but also not a problem, at least currently.

However, going forward if/when we perform more work per version we’ll want to reconcile versions like we do for packages, such that we diff for removed/added versions - and also check that no versions have moved (re-tagging, default branch updates).

The invariant to look for is probably: reference, commit. I don’t think we need to look at the commit date.

So, we’d look for deleted versions, added versions and any versions where (reference, commit) has changed. Those we then delete and re-analyse.

This is a pre-requisite for the builder.

finestructure:
Fixes #422

### DIFF
--- a/Sources/App/Commands/Analyzer.swift
+++ b/Sources/App/Commands/Analyzer.swift
@@ -170,7 +170,9 @@ func reconcileVersions(application: Application,
     let ops = checkouts.map { checkout -> EventLoopFuture<(Package, [Version])> in
         switch checkout {
             case .success(let pkg):
-                return reconcileVersions(application: application,
+                return reconcileVersions(client: application.client,
+                                         logger: application.logger,
+                                         threadPool: application.threadPool,
                                          transaction: transaction,
                                          package: pkg)
                     .map { (pkg, $0) }
@@ -182,16 +184,18 @@ func reconcileVersions(application: Application,
 }
 
 
-func reconcileVersions(application: Application,
+func reconcileVersions(client: Client,
+                       logger: Logger,
+                       threadPool: NIOThreadPool,
                        transaction: Database,
                        package: Package) -> EventLoopFuture<[Version]> {
     guard let cacheDir = Current.fileManager.cacheDirectoryPath(for: package) else {
-        return application.eventLoopGroup.next().future(error:
+        return transaction.eventLoop.future(error:
             AppError.invalidPackageCachePath(package.id, package.url)
         )
     }
     guard let pkgId = package.id else {
-        return application.eventLoopGroup.next().future(error:
+        return transaction.eventLoop.future(error:
             AppError.genericError(nil, "PANIC: package id nil for package \(package.url)")
         )
     }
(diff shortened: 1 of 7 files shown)


Sources/App/Commands/Analyzer.swift
```diff
@@ -213,19 +213,33 @@ func reconcileVersions(application: Application,
...
                                commit: revInfo.commit,
                                commitDate: revInfo.date) }
 
-    let delete = Version.query(on: transaction)
-        .filter(\.$package.$id == pkgId)
-        .delete()
-    let insert = versions.flatMap { versions in versions.create(on: transaction).map { versions }  }
-    return delete.flatMap { insert }
+    let newReconciliation = true
+    if newReconciliation {
```
finestructure, line 225: I've feature flagged this for local testing but this may actually be something we might want to consider having, since this is a rather critical part of the analysis stage. We could leave it in for a while and then remove the code in a couple of days. (I wouldn't want to leave this in for too long.)

On the other hand we can also simply revert the PR. This is not something that's hard to roll back even if there was a merge conflict.

I do like how minimal the actual change to the processing is.

Sources/App/Commands/Analyzer.swift
```diff
@@ -238,7 +252,9 @@ func getManifests(logger: Logger,
             let errors = m.compactMap { $0.getError() }
                 .map { AppError.genericError(pkg.id, "getManifests failed: \($0.localizedDescription)") }
             errors.forEach { logger.report(error: $0) }
-            guard !successes.isEmpty else { return .failure(AppError.noValidVersions(pkg.id, pkg.url)) }
+            if !versions.isEmpty && successes.isEmpty {
+                return .failure(AppError.noValidVersions(pkg.id, pkg.url))
+            }
```
finestructure, line 257: Our tests nicely flagged that previously we blanked assumed getManifests to fail if there were no successful manifest reads. Because we can how call this for a package with no versions (i.e. there were no additions), we also need to ensure that there was something to look for in the first place and only then error out.

Sources/App/Commands/Analyzer.swift
finestructure: I think I'm leaning towards removing that flag and letting it run on staging for half a day.

finestructure:
`test_processing_pipeline` is deadlocking on Linux for some reason (not on macOS). Investigating.

finestructure:
And rightly so, because I'm not using `transaction` but `application.db`. I even considered making `application.db` unavailable when I added the transaction to prevent exactly this mistake from happening.

Sources/App/Commands/Analyzer.swift
daveverwer: Yep, let's take it out and run it on staging for a few hours.

finestructure:
This should be good to merge for dev deployment now, btw

daveverwer:
I just took a reading of staging performance (screenshots on the desktop), let's see how it affects performance.

To be clear, I don't think it's the database operations that are taxing the CPUs here but the constant need to re-process the versions. Reconciliation will mean we'll have much less to do, not just less records being inserted.

PR #451, opened 2020-06-22
Add build

### LINKED ISSUE #443: Add `Build` model
- status
- swift_version
- logs
- platform

finestructure:
Merge after #449 

Fixes #443

### DIFF
--- a/Sources/App/Models/Build.swift
+++ b/Sources/App/Models/Build.swift
@@ -1,0 +1,63 @@
+import Fluent
+import Vapor
+
+
+final class Build: Model, Content {
+    static let schema = "builds"
+
+    typealias Id = UUID
+
+    // managed fields
+
+    @ID(key: .id)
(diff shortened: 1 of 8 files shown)


Sources/App/Models/Build.swift
```diff
@@ -0,0 +1,63 @@
...
+    @Parent(key: "version_id")
+    var version: Version
+
+    // data fields
+
+    @Field(key: "logs")
+    var logs: String?
+
+    @Field(key: "platform")
+    var platform: Platform?
```
daveverwer, line 32: This probably shouldn't be the current `Platform` enum, but a new one for builds.

finestructure: Fixed! As to the actual values of the enum - that's probably going to change a bit but it's a simple, non-schema affecting change.

PR #455, opened 2020-06-23
Builder auth

### LINKED ISSUE #445: Add token auth

finestructure:
Fixes #445

### DIFF
--- a/Sources/App/Models/User.swift
+++ b/Sources/App/Models/User.swift
@@ -1,0 +1,19 @@
+import Vapor
+
+
+struct User: Authenticatable {
+    var name: String
+
+    static var builder: Self { .init(name: "builder") }
+
+    struct TokenAuthenticator: BearerAuthenticator {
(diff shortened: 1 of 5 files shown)


Sources/App/Models/User.swift
```diff
@@ -0,0 +1,19 @@
...
+struct User: Authenticatable {
+    var name: String
+
+    static var builder: Self { .init(name: "builder") }
+
+    struct TokenAuthenticator: BearerAuthenticator {
+        func authenticate(bearer: BearerAuthorization, for request: Request) -> EventLoopFuture<Void> {
+            if
+                let builderToken = Current.builderToken(),
+                bearer.token == builderToken {
```
finestructure, line 13: Plucking it from an env variable is the simplest way, although it wouldn't be that much more complicated to authenticate against a database table. However, then we start storing config in both the env variables and the db and for now it's probably better to just stick with the env.

Plus it saves us the db lookup and we don't need to add another table to keep the tokens.

If we need more api keys/users we can transition to a more scalable mechanism. Does that make sense?

PR #456, opened 2020-06-24
Swift 5.3

finestructure:
Xcode 12 / macOS 11 compatibility


Sources/App/Core/Rollbar.swift
```diff
@@ -49,15 +49,15 @@ extension Rollbar {
...
-            let level: Level
-            let language = "swift"
-            let framework = "vapor"
-            let uuid = UUID().uuidString
+            var environment: String
+            var body: Body
+            var level: Level
+            var language = "swift"
+            var framework = "vapor"
+            var uuid = UUID().uuidString
```
daveverwer, line 57: What changed in Xcode/Swift to make this necessary? (Just curious more than anything!)

finestructure: That was actually a useful warning (it didn't error, just warned): The issue it flagged was that we declared a type `Decodable` but then had `let language = "swift"` - which makes it impossible for decoding to actually set `language`.

PR #471, opened 2020-06-25
Enable CSS for Snapshot Images

Sherlouk:
Okay so I'm not thrilled with the code here, but it's not as bad as the workarounds I was trying yesterday 😂 

Changes within this MR:
* Update Swift Tools Version to v5.3 (Requires Xcode 12, so we can't merge this anyway)
* Remove leading slash on CSS file (I've tested this locally and it still works in browser as well as for tests)
* Added the public folder as resources within the test target (Available using `Bundle.module` a feature of Swift 5.3)
* Added a symbolic link to the Public/ directory within the tests target (this is because resources are relative to the target, not the repository)
* Updated how we load the HTML into the WKWebView
* Updated all snapshot tests inside of WebpageSnapshotTests


Sherlouk:
@finestructure This MR is far from perfect but I'm interested in your thoughts as it's clear you've looked into this extensively!

daveverwer:
Great work! Thank you James

You're right that this one is for @finestructure but a couple of quick questions from me:

* I just want to double-check this won't change any paths to the stylesheet when serving pages, only in tests?
* Would it be possible to also get the icon font working using a similar technique?

The downside is that I now have to make snapshots every time I tweak any CSS property. 🙄

Sherlouk:
> The downside is that I now have to make snapshots every time I tweak any CSS property. 🙄

Aha, I'm not going to apologise for snapshot tests doing what they're meant 😉  We use them at work and it can definitely be frustrating if you completely forget about them but you soon get into the rhythm of knowing when to turn record mode on and when to leave it off!

> I just want to double-check this won't change any paths to the stylesheet when serving pages, only in tests?

So this does (currently, I suppose there's no reason why we couldn't inject a variable to add/remove the leading slash) change the path when serving pages but I've yet to witness this failing in reality. 

I did check running the project with these changes and it worked fine in localhost but perhaps I've overlooked something?

> Would it be possible to also get the icon font working using a similar technique?

So much like the CSS file the issue here is just the leading slashes - if we remove that from the CSS file it does actually work! Likewise if we remove the leading slash from the logo URL that appears!

Does that make sense? @daveverwer (also if I'm being completely stupid and have missed something terrible then please do give me pointers and I'll investigate further!)

daveverwer:
> Aha, I'm not going to apologise for snapshot tests doing what they're meant to 😉 

You're right, and I'm fully behind markup snapshots. They're super important. 

I must admit I'm less convinced on the benefit of image snapshots with CSS loaded. The chances of some other part of the system breaking the CSS *without a markup change* are slim to none. Unless I'm missing something?

daveverwer:
> So much like the CSS file the issue here is just the leading slashes - if we remove that from the CSS file it does actually work! Likewise if we remove the leading slash from the logo URL that appears!

Yea this is the change that worried me. If it works, it works, but I'm a little nervous about this. I'm actually not sure what the expected behaviour of a relative URL is in CSS. If there was a way to make this change apply only in the tests I'd be much happier.
> 
> Does that make sense? @daveverwer (also if I'm being completely stupid and have missed something terrible then please do give me pointers and I'll investigate further!)

Sherlouk:
I think image snapshots are definitely more helpful once you're past a certain scale (but also often painful if you wait until you're at that scale to start making them!) - it can really help to capture any site-wide changes as you mention but I also think they can really be helpful as part of the review phase as it doesn't force me to download and run everything to see what things look like after CSS/DOM changes.

It's also not always immediately obvious what a DOM change corresponds to in terms of visual output. The same CSS/DOM can look entirely different on a small mobile device vs a full screen desktop (I'm fully aware we don't have that setup yet, but there's no reason why we couldn't!)



> Yea this is the change that worried me

It's a great spot but I think it's worth us doing some more investigations before we entirely rule it out. We can, of course, make it a test only difference but if the change causes absolutely no problems then we've just introduced unnecessary complexity. I'm definitely uneducated about the impact of this change so I'll be sure to do more testing!

finestructure:
Hi @Sherlouk , it’s great to see those pages rendering properly! I’ll have a quick skim now and try and leave some remarks if I have any. I’m quite busy working on something else at the moment so I can’t put a lot of time into it right now. However, that fact that it does render out the pages makes me hopeful that we can get it to work in way that doesn’t require too many compromises!

Just generally re: image based snapshot: We’ve got a few knobs we can turn to ease the pain and I think it’d be great to have a render of images, fonts, and styles. We have no other (automatic) coverage on it otherwise.

In another project I’ve got it set up so that I can run the tests with a `RECORD=1` env variable to re-record everything. And if we set precision to something less than 100% we can take some of the edge off. For instance right now macOS 11 is rendering some things differently (often it’s changes in aliasing) and that should take care of that.

Sherlouk:
Absolutely no rush Sven, appreciate you looking at it regardless of how long it takes 😄 

Agreed we can definitely ease some of the pain -- partially through learning but also yeah we have an environment variable too!

I definitely think as briefly mentioned above we should duplicate the image snapshots to do a mobile size and a web size, that will increase the value from them even more so!

Tests/AppTests/Util+ext.swift
```diff
@@ -90,7 +90,23 @@ extension Snapshotting where Value == HTML, Format == NSImage {
...
+                
+                // Save HTML file at root of public directory
+                do {
+                    try html.write(to: htmlURL, atomically: true, encoding: .utf8)
+                } catch {
+                    print("💥 Failed to write index.html: \(error)")
+                }
+                
+                // Load the HTML file into the web view with access to public directory
+                webView.loadFileURL(htmlURL, allowingReadAccessTo: url)
```
finestructure, line 105: Ouf, having to write out the files is a bummer. I wonder if the resource location would have any impact on `loadHTMLString`.

Sherlouk: Unfortunately I don't believe so - I did a hell of a lot of investigations with this and a lot of StackOverflow comments pointed me in the direction of this method.

It seems to all be to do with the file structure and where the resources are relative to the HTML. WebKit (for security reasons) really doesn't like having file:// resources in a different directory - it needs to be a child of the same parent.

This is ultimately what the whole solution pivots on

Sources/App/Views/PublicPage.swift
```diff
@@ -42,7 +42,7 @@ class PublicPage {
             ),
             .link(
                 .rel(.stylesheet),
-                .href("/stylesheets/main.min.css?\(resourceReloadQueryString())")
+                .href("stylesheets/main.min.css?\(resourceReloadQueryString())")
```
finestructure, line 45: I think we should use `SiteURL` here in a similar way we do for images and that should give us the control to DI a variant for testing.

I experimented with that briefly when I was trying to explore if absolute file paths would help render the css. If we use `SiteURL` we can make this path be anything we want in the tests.

Sherlouk: This seems like a sensible way to address Dave's concerns too!

Tests/AppTests/WebpageSnapshotTests.swift
```diff
@@ -13,6 +13,15 @@ class WebpageSnapshotTests: XCTestCase {
...
         Current.date = { Date(timeIntervalSince1970: 0) }
     }
+    
+    override func tearDownWithError() throws {
+        // Remove any HTML files in bundle
+        let fileManager = Foundation.FileManager.default
+        let indexURL = baseURL().appendingPathComponent("index.html")
+        if fileManager.fileExists(atPath: indexURL.path) {
+            try fileManager.removeItem(at: indexURL)
+        }
```
finestructure, line 23: Might be worth sticking these in `/tmp` to avoid stuff being left behind in case of test crashes. I should have something like that in another project.

Sherlouk: This may be possible, but it would also mean that we have to copy the entire `Public` directory to the same tmp directory at the start of the test which could be intensive - especially if we use a different directory for every test.

I got around that duplication, in essence, by copying the files as test resources and using the bundle. I could move this script in with the snapshot extension to guarantee it's done before a new snapshot is created which guarantees success no matter if a previous run crashed?

Of course if you ever clear your derived data it's all cleared up too with the bundle being stored there.

Sherlouk:
I should clarify what the "fix" is here btw, essentially there's some security measures that are built into WebKit when loading file:// URLs that are not in the scope of the HTML file.

This is valid:
```
Directory/index.html
Directory/Public/main.css
```

This is invalid:
```
Directory/index.html
Public/main.css
```

Equally because `loadHTMLString` doesn't have a URL, it doesn't know where it's allowed to load relative URLs from and as such can't load any file URLs (that I've seen).

With the `loadFileURL(allowingReadAccessTo:)` we're able to specify the URL for the file and then it's sources directory. If these are not next to each other then WebKit actually just hangs and the tests timeout.

So for everything to work, we need to create a file (I've called it "index.html") and put it at the root of "/Public" which allows for the same relative URLs used by the website.

Tests/AppTests/WebpageSnapshotTests.swift
Sherlouk: I’m going to look into this a little more to see if we can copy the public directory just once.  Will update you on whether this works in the morning 👍

Sherlouk:
Superseded by #475 which doesn't have an Xcode 12 dependency and addresses many of the points raised in this thread. Keeping this branch around in case we need to refer back to it.

PR #475, opened 2020-06-25
Enable CSS for Snapshot Images [2: Electric Boogaloo]

Sherlouk:
Really sorry for the double draft PR! This one ultimately "fixes" the problem the same way but approaches it completely differently using a temporary directory and none of the Xcode 12 dependencies the other branch has.

I've also used absoluteURLs instead of releativeURLs which should give us a bit more piece of mind when it comes to the production installation as I can more easily mock the URL here. Though if we have a staging instance somewhere we should definitely give it a run there to see how it copes!

I've not yet fixed the icon font (the URL is inside the CSS which makes it a bit more complicated... not entirely sure how to solve that just yet). I've also taken a slightly different tact and now take 2 images (one for desktop and one for mobile at different sizes - this should enable us to test our CSS size queries and get an idea of any differences).

In order to allow different URLs to be used for the DOM dump and the images I've turned each page creation variable into a block - so we generate every page twice but with different URLs. Without this you end up with the temporary/unique URL being shown in the text output which is of course not desirable and would break quite quickly.

This feels like we're much closer to producing something that _could_ be merged... but I want to make sure I'm on the right track with this -- we can always iterate to fix things like icon paths in a future MR?

Let me know what you think and if I'm on the right track here... likewise (as I know Dave had some concerns) if I'm wasting my time here then no problemo just let me know and we can close this down 😄


daveverwer:
I'd like to see this merged, but we may just need a few days to get some other things out of the way before @finestructure can give this the attention it deserves. We're pushing for a goal right now, and it's a little all-consuming!

I feel much better about the absolute URL over the ambiguous relative URL though. 👍 It's not ideal to use an absolute URL, but it's a better solution than the alternative.

Let's hold on this until the middle of next week, when we should come up for air 👍

daveverwer:
After saying I wasn't in favour of CSS testing, I now have a PR in progress where this would be useful. I'm converting my CSS to SCSS now that we're using CodeKit to compile the JS. 😂

Sherlouk:
I'm sure there's a German word for that 😅

daveverwer:
I'm going to hold on my change until we get this in place.

Sherlouk:
Moved to #495

PR #486, opened 2020-06-27
Add package score as HTML comment to package listing

### LINKED ISSUE #440: Include the score in an HTML comment
I was investigating some suspicious ordering in the search results tonight and it would be useful to have the score inserted as an HTML comment somewhere in the package page so it can be more easily checked.

Sherlouk:
Closes #440 

Score is set in an un-executed part of the tests hence it shows as unknown there. Tested locally with a real package and working fine.

Feel free to close unmerged if the need/desire has gone.

### DIFF
--- a/Sources/App/Views/PackageController/PackageShow+Model.swift
+++ b/Sources/App/Views/PackageController/PackageShow+Model.swift
@@ -15,6 +15,7 @@ extension PackageShow {
         var summary: String
         var title: String
         var url: String
+        var score: Int?
 
         struct History: Equatable {
             var since: String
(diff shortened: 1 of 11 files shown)


daveverwer:
The desire is definitely still there! Thank you, James.

We don't need to test it with a real score, but maybe we should at least populate the view model with a score so we can see it coming through in the test?

Let's also not label what the number is quite so boldly. I'm a little nervous that people will think we're judging packages (which we are) and I'm not ready to put that scoring mechanism in the light yet. It's being used to sort, but before the spotlight comes down on how we're judging a package, I'd like to do more work on it. I know people can read these comments (Hi, if you're reading! 👋) and there's nothing bad to hide, I just want it to be subtle untul we're ready.

The positioning of the comment is 👍, but let's just make it as simple as `<!-- 63 -->`.

Thanks James, I'd love to get this merged!

Sherlouk:
Thanks for the feedback Dave, I'll crack on and get those changes made 👍

daveverwer:
@finestructure if you could give this a review too please, I'm happy with it. 👍

finestructure:
Looks great, thanks James!

PR #487, opened 2020-06-27
Handle Shorthand Emojis in Package Summary

### LINKED ISSUE #276: Render :shortcode: emojis in Package summaries
Screenshot 2020-06-05 at 11 57 04

Sherlouk:
Closes #276 

Extension can be used anywhere, only using it on the summary for now.


Sources/App/Views/PackageController/PackageShow+View.swift
```diff
@@ -43,7 +43,7 @@ enum PackageShow {
                 .hr(),
                 .p(
                     .class("description"),
-                    .text(model.summary)
+                    .text(model.summary.replaceShorthandEmojis())
```
finestructure, line 46: Correct me if I'm wrong but does this not decode a 22k line json file and then loop through all 22k of them to find matches on each package page request? If true that'll probably massively impact our response times.

I think we should

- have a performance test to check expected behaviour
- lazy load the list into a static, caching the list
- make the list a dict [":tada:": "🎉"] for faster lookup
- flip the lookup around: pick out emoji patterns `:\(w+):` in the summary and then match them against the list

finestructure: (": tada :" has extra spaces because GH otherwise helpfully replaces it 😆 )

Sherlouk: Unless static methods behave differently on Vapor setups it should only be decoding the file once per instance startup but it does loop through <2000 models looking for any emojis.

I'll add a performance test, use a static RegEx for matching strings and cache into a dictionary for improved times.

finestructure: A static variable would but not a function! If you place that in a test and break inside the method that should demonstrate it.

Sherlouk: Current time for a single paragraph is 0.043s - will share updated times after the above changes :+1:

Sherlouk: Hopefully addressed your concerns and ideas in the latest commit :+1: Let me know what you think

Thanks for the feedback!

finestructure: Great, thanks for following up, James. Since it's that fast it's no problem at all to prime it on first occurrence (plus we smoke test after deploy - so it'll be primed automatically).

I was concerned that the 22k JSON would be slower. We've seen some requests where large payloads cause delays but I guess it's more dramatic there because it's an encode-decode sequence from the db.

PR #488, opened 2020-06-28
Wire up build status

finestructure:


Sources/App/Views/PackageController/PackageShow+Model.swift
```diff
@@ -16,43 +18,6 @@ extension PackageShow {
...
-            var link: Link
-            var swiftVersions: [String]
-            var platforms: [Platform]
-        }
-
-        struct LanguagePlatformInfo: Equatable {
-            var stable: Version?
-            var beta: Version?
-            var latest: Version?
-        }
```
finestructure, line 55: I’m moved these to another file but the whole `PackageShow` + model + view part needs some love. I’ve rushed this initially and again for the builds and it feels messy. Not sure if much can be refactored - this kind of data transformation code is just naturally messy - but it could (should!) be more organised.

PR #490, opened 2020-06-28
Package name build trigger

finestructure:
This adds POST /api/packages/{owner}/{repo}/trigger-builds and cleans up some of the warts in the PackageShow.Model arena.

I’ve also figured out how to dependency inject the test client, so that we now have coverage on the outbound Gitlab requests being sent.

### DIFF
--- a/Sources/App/Controllers/API/API+PackageController.swift
+++ b/Sources/App/Controllers/API/API+PackageController.swift
@@ -33,6 +33,34 @@ extension API {
                 .flatMap { $0.delete(on: req.db) }
                 .transform(to: .ok)
         }
+        
+        func trigger(req: Request) throws -> EventLoopFuture<HTTPStatus> {
+            guard
+                let owner = req.parameters.get("owner"),
+                let repository = req.parameters.get("repository")
+                else {
+                    return req.eventLoop.future(error: Abort(.notFound))
+            }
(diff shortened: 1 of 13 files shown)


Sources/App/Controllers/API/API+PackageController.swift
```diff
@@ -3,37 +3,65 @@ import Vapor
 
 
 extension API {
-
+
```
daveverwer, line 6: Did you turn off the trim whitespace setting in Xcode? I've not seen it insert whitespace like this in your PRs before.

daveverwer:
I didn't merge in case you want to fix the whitespace, but feel free to merge when you're ready!

Sources/App/Controllers/API/API+PackageController.swift
finestructure: I re-indented with the new Xcode-12 formatting. I've just checked and "Automatically trim ..." is set. I think this wasn't properly trimmed previously! All good :)

daveverwer: Interesting. In Xcode 12 beta with these settings:

Screenshot 2020-06-28 at 19 43 04@2x

When I save this file, 91051cc happens.

finestructure: Ah, I've only had the first one ticked. I'll tick the other one as well then.

PR #491, opened 2020-06-28
Whitespaaaaaaaaaace

daveverwer:
The great whitespacening.


docker-compose.yml
```diff
@@ -5,10 +5,10 @@ version: '3.7'
 
 
 x-shared: &shared
```
daveverwer, line 7: This is the only file I'm not sure about, will need to be tested in deployment.

Sherlouk:
Is it worth noting what settings this has been generated with to ensure future commits are the same?

docker-compose.yml
Sherlouk: When I run the new version of this file and the old version of it through http://www.yamllint.com/ they're fundamentally different - some of the indentation changes has actually changed the structure it generates.

For example the environment for x-shared only has one field with the new setup but had many with the old one. Seems to have lost that nested structure.

Recommend we revert before merge.

daveverwer:
> Is it worth noting what settings this has been generated with to ensure future commits are the same? Based on [this](https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/pull/490#discussion_r446685719)

Yes, these are the settings

Screenshot 2020-06-28 at 19 43 04@2x

daveverwer:
This is merged, you can see the added package [here](https://github.com/SwiftPackageIndex/PackageList/blob/main/packages.json#L369) but GitHub was having a wobble when I pushed the change and so it's not showing as merged.

PR #495, opened 2020-06-29
Snapshot with css 2

finestructure:
This builds on and supercesed #475 - thanks a lot for the great groundwork, @Sherlouk !

This is mainly some cleanup of the original code that I needed to do anyway and your help of making the snapshots work prompted me into getting to it :)


Sherlouk:
This is pretty awesome, thanks for carrying the torch forward Sven! 

Great to see a bug has already been found and fixed too 😅

daveverwer:
There is an issue... I ran the tests locally and the snapshots failed. Looking at the two screenshots, you can see the problem...

The snapshot in source control:

test_HomeIndexView desktop

The snapshot from my machine:

test_HomeIndexView desktop_new

finestructure:
You know what you've got to do 🤣

Sherlouk:
Don't tell me you use your computer in light mode, Dave!? 😂 

Good spot! Likely just need to inject some code to mock out `@media (prefers-color-scheme: dark)` -- if no body else does I may be able to experiment with it tomorrow.

daveverwer:
Let's make the snapshots light mode if we get a choice. There's not as much contrast in dark mode and at least for me, it's harder to see what's wrong (I have a problem with one of my eyes that makes it much harder to differentiate). Also, as I'll be doing almost all the styling changes, I get to choose 🤩

daveverwer:
So I made a change to have the absolute URL only necessary in testing mode. (ef4eaea) Feels like a testing requirement shouldn't really affect the production (or development) site.

However, it seems like the tests don't run in `.testing`? It seems they run in `.development`. Ideally, we should revert fde3648 if we can figure out why the `.testing` env isn't being used.

Tests/AppTests/WebpageSnapshotTests.swift
```diff
@@ -3,136 +3,152 @@
...
-        assertSnapshot(matching: page.render(indentedBy: .spaces(2)), as: .lines)
+        assertSnapshot(matching: page, as: .html)
 
         // Snapshot renders slightly differently on macOS 11 (swift 5.3) - exclude it for now
         #if os(macOS) && swift(<5.3)
         if !isRunningInCI {
-            assertSnapshot(matching: page, as: .image(size: defaultSize, baseURL: baseURL()))
+            configs.forEach {
+                assertSnapshot(matching: page,
+                               as: .image(size: $0.size, baseURL: rootDir), named: $0.name)
```
Sherlouk, line 48: rootDir doesn't exist - We use `TempWebRoot.baseURL` everywhere else.

```suggestion
                               as: .image(size: $0.size, baseURL: TempWebRoot.baseURL), named: $0.name)
```

Sherlouk:
Proposal to force light mode:

Add new value to `AppEnvironment` (or some other injectable variable) which is 'disableDarkMode' defaulted to false.
In the tests we will set this to `true`.
We pull out any `@media (prefers-color-scheme: dark) { ... }` statements into a separate "dark.css" file.
In `PublicPage` we then include something a bit like this:
```
.if(Current.disableDarkMode == false, .link(
    .rel(.stylesheet),
    .href(SiteURL.stylesheets("dark.min.css").absoluteURL())
)),
```

Of course in production it would always be included but in our tests we would remove any code which turns it into dark mode - always forcing it to light mode.

From what I can tell there is no real system in WebKit/WKWebView to force a style.

daveverwer:
> Proposal to force light mode:

I wish there was an emoji halfway between 👍 and 👎 . I get it, but that's a lot to get the tests running consistently.

There is another way to do it, but it'd need the CSS rewriting a bit. Instead of using the media query in the CSS, we could use a class on body or some other attribute. The downside here is that I think it would make dark mode support dependent on JS as we'd have to trigger that attribute in real dark mode.

Sherlouk:
Aye I had seen solutions which used JS to add styles and such but I had the same concern as you about requiring JS. Where possible I'd rather avoid that... This felt like the least impactful but agreed a bit much just for tests. Open to other ideas but it does seem our options are limited with WebKit.

Tests/AppTests/Util+Snapshotting.swift
```diff
@@ -22,6 +22,10 @@ extension Snapshotting where Value == () -> HTML, Format == String {
 extension Snapshotting where Value == () -> HTML, Format == NSImage {
     public static func image(precision: Float = 1, size: CGSize? = nil, baseURL: URL) -> Snapshotting {
         Current.siteURL = { baseURL.absoluteString }
+
+        // Force light mode
+        NSApplication.shared.appearance = NSAppearance(named: .aqua)
```
Sherlouk, line 27: 😮 I thought web views were unaffected from the application's preferences (as seems to be the case from my testing on iOS) -- is macOS different here?

finestructure:
This passes for me locally, in dark mode. I’ll add a couple of comments regarding the changes.

Sources/App/Core/SiteURL.swift
```diff
@@ -138,21 +142,28 @@ enum SiteURL: Resourceable {
...
     }
 
-    static func absoluteURL(for path: String) -> String {
-        Current.siteURL() + relativeURL(for: path)
+    #if DEBUG
+    // make `var` for debug so we can dependency inject
+    static var relativeURL = _relativeURL
+    #else
+    static let relativeURL = _relativeURL
+    #endif
```
finestructure, line 160: This is our only change to support dependency injection in the tests: allow `DEBUG` builds to change the `relativeURL` implementation. We don’t need to make anything env specific anymore.

finestructure: The one caveat is that `relativeURL()` cannot have method parameters when it’s a closure but since it’s a single, obvious parameter I’d say that’s fine?

I’ve made the same change for `absoluteURL` purely for symmetry, even though it’s still a function.

Sources/App/Core/SiteURL.swift
```diff
@@ -170,15 +181,15 @@ protocol Resourceable {
...
     }
 
     func absoluteURL(parameters: [String: String]) -> String {
-        "\(Current.siteURL())/\(path)\(parameters.queryString())"
+        "\(SiteURL.absoluteURL(path))\(parameters.queryString())"
     }
 
     func relativeURL(anchor: String? = nil) -> String {
-        "/" + path + (anchor.map { "#\($0)" } ?? "")
+        "\(SiteURL.relativeURL(path))" + (anchor.map { "#\($0)" } ?? "")
```
finestructure, line 192: This is just ensuring we route all url methods through the same two methods on `SiteURL`. Behaviour is unchanged (covered by `SiteURLTests`).

Sources/App/Views/PublicPage.swift
```diff
@@ -201,7 +203,7 @@ class PublicPage {
                 .a(
                     .href(SiteURL.home.relativeURL()),
                     .h1(
-                        .img(.src(SiteURL.images("logo.svg").relativeURL())),
+                        .img(.src(SiteURL.images("logo.svg").absoluteURL())),
```
finestructure, line 206: Actually, now that I see this: would could use the same mechanism we use for stylesheets to avoid this change here as well.

finestructure: Done

restfiles/trigger-build-package-single.restfile
```diff
@@ -0,0 +1,27 @@
...
+                    name: unknown
+                    version: test
+                swiftVersion:
+                    major: 5
+                    minor: 2
+                    patch: 4
+
+        validation:
+            status: 200
+        log: json
```
finestructure, line 27: Whoops, I wanted to commit that separately but it might not be worth making that a PR. Let me know!

Tests/AppTests/Util+Snapshotting.swift
finestructure: Seems to work for me! :)

PR #500, opened 2020-07-02
Add git clean, re-clone if checkout still fails

finestructure:
Fixes #498

PR #501, opened 2020-07-02
Fix base image

finestructure:


daveverwer:
I'm seeing some unexpected changes here which I expected to go away when #500 was merged, but they haven't.

finestructure:
Ah yes, I hadn't pointed this at the branch I continued from.

PR #508, opened 2020-07-05
More CSS refactoring.

daveverwer:
I separated these from the previous PR as these do change the snapshot tests. The changes to the snapshots are either minor or an improvement though.


daveverwer:
Weird, I merged main into this branch before pushing so that your changes from this morning would not show, but it still shows them.

daveverwer:
Xcode had a think about everything and it feels much better now... It had got a bit confused.

finestructure:
I love how the snapshots make that an even better review experience than checking it out and running it locally 🎉

PR #511, opened 2020-07-05
I broke two things in the header/footer, this fixes them!

daveverwer:


finestructure:
In what way is it broken? Seemes to fine in the snapshots and on staging?

finestructure:
Oh, I see - the blue header. I saw that in the snapshots and thought it’s a bit low contrast but then I thought surely that´s intentional 😀

daveverwer:
I had broken the footer on mobile too

PR #517, opened 2020-07-09
Package dump 5.3

### LINKED ISSUE #465: Support packages that require Swift 5.3.

finestructure:
Fixes #465 

- [x] add test with 5.3 Package.swift
- [x] deploy to staging and see if it picks up the 5.3 packages (looks for `twostraws/VisualEffects`)
- [ ] monitor for any regressions with other packages (probably unlikely)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,7 +7,7 @@ jobs:
     name: Test
     runs-on: ubuntu-latest
     container:
-      image: swift:5.2.4-bionic
+      image: finestructure/spi-base:0.1.0
     services:
       postgres:
         image: postgres:12.1-alpine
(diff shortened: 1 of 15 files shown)


finestructure:
Right, I know why it’s not getting picked up yet: due to the version consolidation we’re not re-analysing versions here. Package parsing failures are too common to _not_ save a version, especially as old, incompatible packages that never change would be churning if we didn’t.

At the same time, we’ve now frozen the 5.3 packages as failing and won’t re-analyse unless the underlying commit changes.

I can easily fix that for packages we know are 5.3 (by clearing the versions for them manually). The question is: how do we find all packages (or versions, rather) with 5.3 Package.swift files?

finestructure:
Visual Effects is available now on staging: https://staging.swiftpackageindex.com/twostraws/VisualEffects

daveverwer:
Let's get this on production as soon as [PackageList#465](https://github.com/SwiftPackageIndex/PackageList/pull/465) is merged.

PR #521, opened 2020-07-12
Add Build.logUrl field

finestructure:
In preparation to send the build logs from the builders.

NB: schema change, should deploy as 1.4.0

### DIFF
--- a/Sources/App/Migrations/011/UpdateBuildAddLogURL.swift
+++ b/Sources/App/Migrations/011/UpdateBuildAddLogURL.swift
@@ -1,0 +1,16 @@
+import Fluent
+
+
+struct UpdateBuildAddLogURL: Migration {
+    func prepare(on database: Database) -> EventLoopFuture<Void> {
+        database.schema("builds")
+            .field("log_url", .string)
+            .update()
+    }
+
+    func revert(on database: Database) -> EventLoopFuture<Void> {
+        database.schema("builds")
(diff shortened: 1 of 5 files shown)


daveverwer:
So the log itself goes into the (existing) `logs` field? Is this for the URL to the GitLab log?

finestructure:
Yes, or any queueing system or storage system we intend to support/use instead perhaps. The idea is that we store a limited log in `logs` and a reference to the complete logs in `log_url`.

An alternative would be to just make `logs` accept a structured `json`, like so:

```
{
“logs”: “...”,
“url”: “...”
}
```

I can’t really decide which I like better tbh. But I’m pretty sure we’ll want to keep a reference to the Gitlab pipelines somewhere.

daveverwer:
I think for a large text field I'd much rather just have that in a standard `text` type field.

Sounds good! Merging now and I'll watch it through to staging then deploy to prod if all goes well.

daveverwer:
Seems like something is failing repeatedly while deploying to staging:

https://gitlab.com/finestructure/swiftpackageindex-dev/-/jobs/634786986

Starting at about line 70, it can't find a manifest.

finestructure:
Yeah, it‘s waiting on the prod project building. I‘ll give it a prod - heh.

finestructure:
Retry worked and it‘s deployed now

PR #524, opened 2020-07-13
Work in Progress: Builds styling

daveverwer:
Not ready for merging!

There are still several things need doing before this is ready to merge, but I wanted to get the markup and styles in place.

- [x] Replace placeholder markup with real markup based on build results.
- [x] We should add a snapshot test now that this is becoming a real page.

Screenshot 2020-07-14 at 00 00 24@2x

Screenshot 2020-07-13 at 23 59 21@2x


finestructure:
Ok, I’ve wired this up and added the snapshot test. I’ll do the log links separately - good to merge!

daveverwer:
Deploying to production as 1.4.2

PR #543, opened 2020-07-19
Add a link to the builds page from the Package page

### LINKED ISSUE #537: Add a link to the build logs from the package page
Add a link below both the Swift version and platform compatibility matrices to the build log index page for that package.

daveverwer:
Just below the Swift version matrix.

Fixes #537


Sources/App/Views/PackageController/PackageShow+Model.swift
```diff
@@ -324,6 +337,13 @@ extension PackageShow.Model {
             .h3("Swift Version Compatibility"),
             .ul(
                 .forEach(rows) { swiftVersionCompatibilityListItem($0) }
+            ),
+            .p(
+                .class("right"),
+                .a(
+                    .href(SiteURL.package(.value(repositoryOwner), .value(repositoryName), .builds).relativeURL()),
+                    "Full build results"
+                )
```
finestructure, line 346: I just realised that this will show up on production. Do we want to guard that for now?

daveverwer: I think this is protected by the `guard` at the top of this function (Line 330) isn't it?

finestructure: Ah of course, we've not published any of that so far!

PR #559, opened 2020-07-25
Platforms matrix

### LINKED ISSUE #535: Create platform matrix compatibility view model
The data required by this view model is matrix of platforms (iOS, macOS, tvOS, and watchOS) against the (up to) three significant versions (default branch, latest stable, and latest beta).

Each status in the matrix can be either:

* Supported
* Not supported
* Unknown

In order for a platform to show as supported, *any* build with *any* version of Swift compiled for that platform needs to pass.

**Note:** watchOS is not being processed, but the view model should support it so we can show that we're not currently processing it by showing Unknown for everything in that column.

finestructure:
Fixes #535


finestructure:
Alright, I’m done tweaking this now :)

PR #575, opened 2020-08-01
Significant versions

### LINKED ISSUE #574: Mark significant versions
We have started making heavy use of the following [“client side” (i.e. code side vs SQL side) logic](https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/blob/2e0f3f528860eb458791457aa65c96ede41bc738/Sources/App/Views/PackageController/Package%2BModelSupport.swift#L107) to find the (up to) three “significant versions” of a package:

```swift
    func releases() -> (stable: Version?, beta: Version?, latest: Version?) {
        guard let versions = $versions.value else { return (nil, nil, nil) }
        let releases = versions
            .filter { $0.reference?.semVer != nil }
            .sorted { $0.reference!.semVer! < $1.reference!.semVer! }
        let stable = releases.reversed().first { $0.reference?.semVer?.isStable ?? false }
        let beta = releases.reversed().first {
            // pick first version that is a prerelease *and* no older (in terms of SemVer)
            // than stable
            ($0.reference?.semVer?.isPreRelease ?? false)
                && ($0.reference?.semVer ?? SemVer(0, 0, 0)
                        >= stable?.reference?.semVer ?? SemVer(0, 0, 0))
        }
        let latest = defaultVersion()
        return (stable, beta, latest)
    }
```

As part of looking into “build reconciliation” in issue #561 it’s becoming increasingly awkward having to pull down all versions of a package and then identify the three ones we’re interested in. Since these significant versions never change on their own, we could simply set a flag on the version record as part of the analysis stage. This would make downstream selections _a lot_ easier.

finestructure:
Fixes #574

### DIFF
--- a/Sources/App/Views/PackageController/Package+ModelSupport.swift
+++ b/Sources/App/Views/PackageController/Package+ModelSupport.swift
@@ -28,7 +28,9 @@ extension Package {
             .first()
             .unwrap(or: Abort(.notFound))
     }
-    
+
+    // FIXME: inline into Analyzer.updateLatestVersions and use latest property instead
+    @available(*, deprecated)
     func defaultVersion() -> Version? {
         guard
             let versions = $versions.value,
@@ -103,7 +105,9 @@ extension Package {
             executables: version.products.filter(\.isExecutable).count
         )
     }
-    
+
+    // FIXME: inline into Analyzer.updateLatestVersions and use latest property instead
+    @available(*, deprecated)
     func releases() -> (stable: Version?, beta: Version?, latest: Version?) {
         guard let versions = $versions.value else { return (nil, nil, nil) }
         let releases = versions
--- a/Sources/App/Commands/Analyzer.swift
+++ b/Sources/App/Commands/Analyzer.swift
@@ -78,6 +78,7 @@ func analyze(application: Application, packages: EventLoopFuture<[Package]>) -> 
             return versions
                 .map { getManifests(logger: application.logger, versions: $0) }
                 .flatMap { updateVersionsAndProducts(on: tx, results: $0) }
+                .flatMap { updateLatestVersions(on: tx, results: $0) }
         }
     }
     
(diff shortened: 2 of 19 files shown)


Sources/App/Views/PackageController/Package+ModelSupport.swift
```diff
@@ -28,7 +28,9 @@ extension Package {
             .first()
             .unwrap(or: Abort(.notFound))
     }
-    
+
+    // FIXME: inline into Analyzer.updateLatestVersions and use latest property instead
+    @available(*, deprecated)
     func defaultVersion() -> Version? {
```
finestructure, line 34: I’ll make this change in a follow-up PR

Sources/App/Views/PackageController/Package+ModelSupport.swift
```diff
@@ -103,7 +105,9 @@ extension Package {
             executables: version.products.filter(\.isExecutable).count
         )
     }
-    
+
+    // FIXME: inline into Analyzer.updateLatestVersions and use latest property instead
+    @available(*, deprecated)
     func releases() -> (stable: Version?, beta: Version?, latest: Version?) {
```
finestructure, line 111: I’ll make this change in a follow-up PR

Sources/App/Views/PackageController/Package+ModelSupport.swift
finestructure: Also note: we’ll need to deploy this PR here and let it run a full processing cycle on all records before I replace these methods with the shortcuts using the new field (or just live with the brief window of not showing the proper versions until everything’s caught up).

daveverwer: Is it worth doing some kind of one-time manual verification after this processing has completed just to make sure that the new versions line up with the old? Is there a SQL query you have handy to select these three versions without the three fields that we could export as a CSV and compare against a similar query that runs from the tags after processing?

Might not be worth it, but it could be a good "belt and braces" test.

finestructure: Yes, I was going to run a check on staging. I've already checked that we are in fact updating all packages, i.e. we can be sure everything gets analysed within a couple of hours and there are no stale packages.

I wasn't going to do a 1:1 comparison, just a check that there are no packages with all versions having `latest = NULL`. That should suffice.

daveverwer: OK, so we shouldn't push the deployment through to production until that's done then 👍

It's deploying to dev now, then we can let it do a full processing and inspect the results there. Sounds good.

finestructure: Yeah, it's probably best to pause twice: once to check the `latest` field is set and then, after deploying to prod, to wait and let the processing convert them all before we deploy the follow-up PR that uses the new field.

PR #576, opened 2020-08-01
Fix releases deprecation

finestructure:
This removes the deprecations introduced in #575 and clean up some cases where we used `Kind` types in the view model that can now be replaced by the canonical `App.Version.Kind`.


finestructure:
Leaving in Draft until we’ve validated #575 does it’s job correctly and processing has completed on prod.

finestructure:
Both dev and prod are done processing.

PR #584, opened 2020-08-03
Builder candidate selection & trigger command

finestructure:
- adds `trigger-builds` command, allowing triggering via candidate selection or package id

```
❯ swift run Run trigger-builds --help

Usage: .build/x86_64-apple-macosx/debug/Run trigger-builds [--limit,-l] [--id]

Trigger package builds

Options:
  limit     id
```
Note that the `limit` flag operates on the package level, i.e. `--limit 1` can still trigger dozens of builds.

- candidate selection by identifiying packages with build gaps in general
- and subsequently identifying which builds exactly need to be triggered
- creates build record stubs with `status = pending` to record a build has been enqueued and prevent its re-triggering

NB: there’s some additional change noise in the PR, because I also unified `SwiftVersion` and `SwiftVersionCompatibility`. The main reason was to make it clearer what the canonical source of our platforms and swift versions is.

## Deployment

This is safe to deploy - it only exposes a new command that we can use to trigger but there's no automated running of it yet.


Sources/App/Commands/BuildTrigger.swift
```diff
@@ -0,0 +1,162 @@
...
+        .column(packageId)
+        .from(versions)
+        .join("builds", method: .left, on: "builds.version_id=versions.id")
+        .where(isNotNull(latest))
+        .groupBy(packageId)
+        .groupBy(latest)
+        .groupBy(SQLColumn(updatedAt, table: versions))
+        .having(buildUnderCount)
+        .orderBy(SQLColumn(updatedAt, table: versions))
+        .limit(limit)
```
finestructure, line 113: I’m tempted to replace this with the corresponding raw SQL query. I’m not sure the type safety buys us much here and it’s not a dynamic query like search where we profit from automatic quoting.

The raw query has the advantage that it’s easy to copy and run it for manual validation.

Sources/App/Views/PackageController/Builds/BuildShow+View.swift
```diff
@@ -75,20 +75,25 @@ private extension Build.Status {
...
+            case .pending: return "Build Pending"
         }
     }
 
     var cssClass: String {
         switch self {
             case .ok: return "green"
             case .failed: return "red"
+            // FIXME: what class to use?
+            case .pending: return "grey"
```
finestructure, line 87: We have a choice of creating the build stubs with `status = NULL` or `status = 'pending'` but either way we'd need to handle the extra state here. Is there an existing class I could/should use?

Sources/App/Views/PackageController/Builds/BuildShow+View.swift
```diff
@@ -75,20 +75,25 @@ private extension Build.Status {
...
         }
     }
 
     var cssIcon: String {
         switch self {
             case .ok: return "matrix_succeeded"
             case .failed: return "matrix_failed"
+            // FIXME: what icon to use?
+            case .pending: return "matrix_pending"
```
finestructure, line 96: Dto

daveverwer: Yep, this one certainly needs handling. I've set it to the clock icon we already have in 534ef6e. I might change it up later but that icon will do to get this PR merged.

Sources/App/Views/PackageController/Builds/BuildShow+View.swift
daveverwer: I've added this class for now, but do we really want to let people get to this page for a pending build? I guess they can see the build command that's going to be executed, but it might be worth making it the same as the "unknown" state where this page isn't accessible.

Either way, I've added the CSS for this so it'll work if we decide to keep this use case in. Let's chat about it in the morning.

PR #585, opened 2020-08-04
Fix docker login on spi-dev

finestructure:
The retry change in the `before_script` stage seems to have broken login on `spi-dev` (and prod, but we haven't tagged since then).

This reverts that change.


finestructure:
I'll sudo merge this so I can test the trigger

PR #592, opened 2020-08-04
Added getStatusCount

finestructure:
This adds a pipeline pending check before triggering any builds. Note that it's only looking at pending builds, without running. That's just to avoid having to make too many requests to Gitlab. If we ever have so many runners that this will get us in trouble with the total we can lower the limit or just make that extra request :)

There's another state that counts towards the limit of 500: `created`. This seems to be a very short lived transitional state that pops up briefly before a pipeline goes to pending. I don't think it's worth querying for those in addition because we have enough margin even we should undercount pending a little if we catch some in created state.

Also, because we won't run the trigger in a loop tighter than a minute, this would only be a problem if a pipeline was in pending state for longer than a minute (since the last trigger command), which I don't think can be the case at all.


Sources/App/Core/Constants.swift
```diff
@@ -9,6 +9,8 @@ import Vapor
 
 
 enum Constants {
+    static let defaultGitlabPipelineLimit = 200
```
finestructure, line 12: I've set the fallback limit (if we don't have one set via secrets) a bit more conservative. Plan is still to go with 300 via the env variables.

PR #593, opened 2020-08-04
Override switch

### LINKED ISSUE #587: More feedback from `trigger-builds` command
```
ssh spi-dev
# attach to one of the containers, for instance the analyser
docker exec -it spi_analyze_1 bash
# then inside the container run the command we will later run via docker-compose
root@6bbeb922eb4e:/run# ./Run trigger-builds --limit 1
Triggering builds (limit: 1) …
```

At least log the package id that's being processed, plus what versions and platforms are being triggered (in aggregate, don't spam it for all builds)

finestructure:
Continued from #592 (merge after)

Also fixes #587


daveverwer:
Not sure if this is ready to go as it's still draft but all looks 👍 to me.

finestructure:
Was just in draft so the other one will go first!

PR #599, opened 2020-08-05
Add trigger downscaling

### LINKED ISSUE #598: Add trigger command to compose file
- [x] test prod trigger
- [x] trigger command
- [x] downscaling factor to later reduce processing rate

The way I envision this working is that now that all the pieces are in play, we add

```
  trigger-builds:
    image: registry.gitlab.com/finestructure/swiftpackageindex:${VERSION}
    <<: *shared
    depends_on:
      - migrate
    entrypoint: ["/bin/bash"]
    command: ["-c", "--",
      "trap : TERM INT; while true; do ./Run trigger-builds --env ${ENV} --limit ${BUILD_TRIGGER_LIMIT:-1}; sleep ${BUILD_TRIGGER_SLEEP:-60}; done"
    ]
    restart: always
```

to the compose file to kick off builds.

This will trigger available builds every 60s if the gitlab queue has capacity.

The downscaling factor is a number between 0 and 1 that we set to say, 0.1 or 0.05, on staging and to 1 on prod. It scales the processing rate to 10% or 5%, meaning that staging behaves exactly like production except that processing is 10x or 20x slower (or lower capacity). We will still go through all the packages but take a lot longer.

If we need specific packages fast tracked on staging, we can use the package id based triggering mechanism via the command with a parameter to ignore downscaling (which I'll add). Or we use the existing trigger mechanism via the script and rester, which is also still available.

finestructure:
Fixes #598 

With this we're good to go.

I've set ALLOW_BUILD_TRIGGERS to true on staging and false on prod for now and BUILD_TRIGGER_DOWNSCALING to 0.2 on staging.

Once we deploy this, staging should start picking up builds at a 20% rate. Since this is kicking off every minute we should quickly see builds going in and also the downscaling taking effect.

Production wouldn't yet process any builds due to the override being off.

When we're happy with that, I'd propose to make the following changes and redeploy to make it take effect:

staging:

BUILD_TRIGGER_DOWNSCALING: 0.05

prod:

ALLOW_BUILD_TRIGGERS: true

That'll throttle staging back and open up processing on production.

### DIFF
--- a/Sources/App/Commands/BuildTrigger.swift
+++ b/Sources/App/Commands/BuildTrigger.swift
@@ -48,6 +48,10 @@ func triggerBuilds(on database: Database,
         logger.info("Build trigger override switch OFF - no builds are being triggered")
         return database.eventLoop.future()
     }
+    guard Current.random(0...1) <= Current.buildTriggerDownscaling() else {
+        logger.info("Build trigger downscaling in effect - skipping builds")
+        return database.eventLoop.future()
+    }
     return Current.getStatusCount(client, .pending)
         .flatMap { count -> EventLoopFuture<Void> in
             // check if we have capacity to schedule more builds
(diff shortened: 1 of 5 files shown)


daveverwer:
So from skimming the code it looks like the downscaling processes a random 5% (or whatever threshold) of builds rather than a consistent set? That'll probably be OK as we'll have data to test with. The only difficulty will be knowing *which* packages have data.

PR #604, opened 2020-08-07
Change unique index on builds to ignore patch version

### LINKED ISSUE #590: Upsert builds ignoring patch version
We are currently upserting build records by matching the entire swift version, including patch revision. In order to avoid keeping multiple patch variants in the database, we should probably upsert over major.minor instead to avoid this scenario:

Screenshot 2020-08-04 at 14 39 16

This leads to the pending status still being shown even though a failed status was reported but for the actual builder swift version that was used to compile:

Screenshot 2020-08-04 at 14 41 25

finestructure:
Fixes #590 

I've already applied an equivalent temporary index

```
CREATE UNIQUE INDEX "temp_1"
            ON builds (
                version_id,
                platform,
                (swift_version->'major'),
                (swift_version->'minor')
            );
```

on both dev and prod dbs to ensure that the migration will succeed.

Once this is merged and deployed I'll remove those indexes again.

⚠️ Schema change, deploy as 1.8.0 ⚠️

### DIFF
--- a/Sources/App/Migrations/015/UpdateBuildUniqueIndex1.swift
+++ b/Sources/App/Migrations/015/UpdateBuildUniqueIndex1.swift
@@ -35,7 +35,7 @@ struct UpdateBuildUniqueIndex1: Migration {
             .unique(on: "version_id", "platform", "swift_version")
             .update()
             .flatMap {
-                db.raw(#"DROP INDEX "\#(newIndexName)""#).run()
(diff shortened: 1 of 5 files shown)


daveverwer:
I'll send through to production as 1.8.0.

finestructure:
Migration applied without issued and I've dropped `temp_1` on both dbs.

PR #607, opened 2020-08-08
Linux builds

### LINKED ISSUE #606: Add support for Linux builds
- [x] add docker build command
- [x] think through job distribution/pick up
- [x] set up docker runner
- [x] manually trigger on staging
- [x] think through deployment with runner backlog
- [x] update server platform list

finestructure:
Fixes #606 

What will happen when we merge this?

- Once deployed to dev (automatic), linux processing will start for staging (albeit at a very slow pace, so we will likely not observe any of it for quite a while)
- Once we tag and deploy on prod we will start linux processing for real and in volume

~~NB: Apparently the platform section on the package page is not driven by the `allActive` property. (The snapshots for PackageShow are unchanged.) This is obviously purely a display issue. We can either hold off on this PR while we fix this to make it a part of it or go ahead and start processing while we update the page.~~ (now fixed)


Sources/App/Models/Build+Platform.swift
```diff
@@ -54,7 +54,7 @@ extension Build {
 
         /// Currently supported build platforms
         static var allActive: [Self] {
-            [.ios, .macosSpm, .macosXcodebuild, .tvos, .watchos]
+            [.ios, .macosSpm, .macosXcodebuild, .linux, .tvos, .watchos]
```
finestructure, line 57: This determines the platform order in the build matrix. I was thinking behind iOS and macOS but before tvOS & watchOS is a good place for it to appear.

finestructure:
This is ready to go but leaving it in Draft so we can coordinate merge and deploy.

Sources/App/Views/PackageController/PackageShow+Types.swift
```diff
@@ -149,22 +152,24 @@ extension PackageShow.Model {
...
              watchosStatus: BuildStatus) {
             self.ios = .init(parameter: .ios, status: iosStatus)
+            self.linux = .init(parameter: .linux, status: linuxStatus)
             self.macos = .init(parameter: .macos, status: macosStatus)
             self.tvos = .init(parameter: .tvos, status: tvosStatus)
             self.watchos = .init(parameter: .watchos, status: watchosStatus)
         }
 
-        var cells: [BuildResult<PlatformCompatibility>] { [ios, macos, tvos, watchos] }
+        var cells: [BuildResult<PlatformCompatibility>] { [ios, macos, linux, tvos, watchos] }
```
finestructure, line 172: This determines the order of the table columns on the package show page. Ideally those two were more tightly coupled or had the same source but that's not a trivial task due to different aggregation. Probably not worth losing sleep over unless we see a huge influx of platforms.

daveverwer: Yea I think this is fine, and I think I agree with the order. The only other option would be to put Linux on the right-hand side as it's quite distinct. I'm happy with it here though.

finestructure:
Off we go then :)

Issue #566, opened 2020-07-29
Split the builds matrix into sections based on Swift compiler version


daveverwer:
The full builds index is huge, and also has spacing issues with the header column as we found out in #563. We should split it into several matrices, by Swift version.

We should use an `h3` with the Swift version, and then just have the platform/variant in the first column of each matrix. This will help with showing the full Swift version, but it'll also break up that *huge* matrix.

daveverwer:
We could also roll up and indicate the build status for the Swift version using the lozenge we're adding in #564

daveverwer:
> We could also roll up and indicate the build status for the Swift version using the lozenge we're adding in #564

Thinking about it... It's already rolled up into the Package page, and this only rolls up Swift version status, not platform status. It's weird. Let's not do it.

PR #611, opened 2020-08-10
Arm builds

### LINKED ISSUE #605: Add support for Apple Silicon builds
- [x] set up runner on DTK
- [x] add job to builder CI config
- [x] test and confirm ARM builds on DTK
- [x] add cases (xcode, spm) to supported platforms
- [ ] tweak concurrency on `spi-dtk-builder`

finestructure:
This adds triggering and display of ARM builds.

The table doesn't look great with the long label, in particular on mobile but I guess we can tweak that in another pass.

Leaving this in draft so I can merge it when I'm around to monitor the runners.

Fixes #605


daveverwer:
The table display issues should be mitigated by #566

daveverwer:
> The table display issues should be mitigated by #566

Ignore that, I see the issue now. It's not in that view, it's in the Package show page. I'll have a think about what we can do with that.

PR #620, opened 2020-08-12
Group versions on Build Index View

### LINKED ISSUE #566: Split the builds matrix into sections based on Swift compiler version

finestructure:
Fixes #566 

I've tried both with and without the leading version in the individual items:

test_BuildIndex desktop

test_BuildIndex desktop 2

and I _think_ I like it better _with_ the version still showing in each row.

In particular, I think we'll want to keep it there at least in the mobile variant where it's easy to have a whole section on screen without any `h3` showing:

test_BuildIndex mobile

Also, I think we should tweak the padding of the `h3`s a bit. It should be closer to the table below, in particular for desktop.


daveverwer:
> In particular, I think we'll want to keep it there at least in the mobile variant where it's easy to have a whole section on screen without any `h3` showing:

> Also, I think we should tweak the padding of the `h3`s a bit. It should be closer to the table below, in particular for desktop.

Both great points and I'd like to have a play with this too. We can either get this merged and I can do the tweaks as an additional PR to not hold up the deployment of this which is 99% done. I'll approve this change, and let you make the decision on whether to merge and let me patch on top, or whether we keep this open and I work on this PR.

finestructure:
Great, let's get this in then and tweak it afterwards!

PR #623, opened 2020-08-12
Build badges

### LINKED ISSUE #619: Build badges
Just recording that here as a reminder

Screenshot 2020-08-12 at 09 00 45

finestructure:
Merge after #620

Fixes #619

### DIFF
--- a/Sources/App/Controllers/API/Package+Badge.swift
+++ b/Sources/App/Controllers/API/Package+Badge.swift
@@ -98,27 +98,45 @@ extension Package {
 }
 
 
-private func _badgeMessage(platforms: [Build.Platform]) -> String {
-    platforms.map {
-        switch $0 {
-            case .ios:
-                return "iOS"
-            case .macosSpm, .macosXcodebuild, .macosSpmArm, .macosXcodebuildArm:
-                return "macOS"
-            case .tvos:
-                return "tvOS"
-            case .watchos:
-                return "watchOS"
-            case .linux:
-                return "Linux"
+func _badgeMessage(platforms: [Build.Platform]) -> String {
+    struct Value: Hashable {
+        var index: Int
+        var value: String
+        init(_ index: Int, _ value: String) {
+            self.index = index
+            self.value = value
         }
     }
+    return Array(
+        Set(
+            platforms
+                .map { p -> Value in
+                    switch p {
+                        case .ios:
+                            return .init(0, "iOS")
+                        case .macosSpm, .macosXcodebuild, .macosSpmArm, .macosXcodebuildArm:
+                            return .init(1, "macOS")
+                        case .linux:
+                            return .init(2, "Linux")
+                        case .tvos:
+                            return .init(3, "tvOS")
+                        case .watchos:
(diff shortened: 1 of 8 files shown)


finestructure:
Tested against local server:

- https://05cb462bf161.ngrok.io/api/packages/Alamofire/Alamofire/badge?type=swift-versions
- https://05cb462bf161.ngrok.io/api/packages/Alamofire/Alamofire/badge?type=platforms

```markdown
![](https://img.shields.io/endpoint?url=https%3A%2F%2F05cb462bf161.ngrok.io%2Fapi%2Fpackages%2FAlamofire%2FAlamofire%2Fbadge%3Ftype%3Dswift-versions)

![](https://img.shields.io/endpoint?url=https%3A%2F%2F05cb462bf161.ngrok.io%2Fapi%2Fpackages%2FAlamofire%2FAlamofire%2Fbadge%3Ftype%3Dplatforms)

# Alamofire
```


Screenshot 2020-08-12 at 18 31 21

Sherlouk:
In the screenshot above we have macOS listed twice under platform compatibility - did you fix this?

finestructure:
Ah, of course - good spot :)

It's been Dave's idea, all credit to him! :)

Sources/App/Controllers/API/Package+Badge.swift
```diff
@@ -0,0 +1,142 @@
...
+            .compactMap(latestVersion(for:))
+        return versions.reduce(into: []) {
+            $0.append(contentsOf: $1.$builds.value ?? [])
+        }
+    }
+
+}
+
+
+func _badgeMessage(platforms: [Build.Platform]) -> String {
```
finestructure, line 101: This function feels a bit clumsy but I couldn't come up with a better way to make the list unique and preserve ordering. Once tuples gain Hashable autoconformance we can at least drop the `Value` helper type.

daveverwer:
This is amazing work Sven! Taking advantage of their endpoint API is the perfect way to implement it.

We need to think about how we position this and promote it to package authors, but the hard work looks to be done! We want to be careful about making it simple to get right initially and check whether there's anything in here that we might want to abstract. Basically, I want to chat through the idea of proxying these requests through the SPI server so that if the shields API ever changes, we're not left with thousands of broken badges.

finestructure:
Yeah, let's catch up, but let me just add one thing regarding proxying. We're very much insulated from shields.io changes by using their API endpoint with a versioned payload. I think a lot would need to happen (so much that we'd be impacted even if we proxied) for that to break.

But there's a bigger reason why I think not proxying is better for us. I haven't timed the query or the overall cost for the badge API call yet but it's clear just from the db access part that it's not insiginficant. If badges get used a lot, we'd be hitting our service a lot when Github pages are viewed and I don't think we'd want to do that without some sort of caching on our end.

The beauty here is that we can set the caching at the shield.io level (I've set it to 6 hours) and not worry about any of that. That means we'll see at most one hit per package per 6 hours and don't have to worry about tuning that api call further or shielding it with a cache.

PR #632, opened 2020-08-15
Copy badge

### LINKED ISSUE #630: Add a "Copy Badge Markdown" link to the package page.
Temporarily, just to the left of the "Full Build Results" link, but may move before launch.

finestructure:
Fixes #630 

This needs moving 😆  The `appendChild()` makes it get tacked onto the end instead of after the actual `p` element with the markdown.

Screenshot 2020-08-15 at 18 27 12


Sources/App/Views/PackageController/PackageShow+Model.swift
```diff
@@ -350,6 +350,15 @@ extension PackageShow.Model {
...
             ),
+            .p(
+                .id(
+                    // TODO: remove check when we go live
+                    ((try? Environment.detect()) ?? .development) == .production
+                        ? "ignored"
+                        : "swift_version_badge"),
+                .text(badgeMarkdown(for: .swiftVersions)),
+                .attribute(named: "hidden")
+            ),
```
finestructure, line 361: I ended up sticking it in `p` tag, because the markdown wouldn't go in the `href` (got prefixed with the package url - I guess it has to be a url or something).

This makes the button show up in a new line. But even as an `<a href=...` the button was still misplaced - it came after the "Full build logs" link, because it's attached like so: `element.parentNode.appendChild(copyButtonElement)`

Probably an easy fix, and I may dig around a bit but I wanted to push what we've got for now.

finestructure:
I've managed to get these two right-aligned variants:

Screenshot 2020-08-16 at 11 31 29

The padding with the matrix is a bit tighter than before, because the `p` is now `div` but I think for the stacked variant that's probably preferable.

Of the two, I think the stacked is clearer in that the badge copy is something different than the full build results.

It also avoids the vertical alignment issue of the button (but I'm sure that can be fixed somehow).

finestructure:
One other thing to consider (at some point, probably not now) is that in the case where the results _don't_ collapse across significant versions is that the badge is currently a union of results. I.e. for the following example it'd report supporting all versions except 4.2:

Screenshot 2020-08-16 at 11 39 26

We may want to support specifying the significant version in the future.

daveverwer:
> One other thing to consider (at some point, probably not now) is that in the case where the results _don't_ collapse across significant versions the badge is currently a union of results.

I think this is the correct decision, we're not hoping to replace our matrix with this badge, just guide people to it.

PR #644, opened 2020-08-17
Per package id trigger limit check

### LINKED ISSUE #643: Check build pipeline limit per package
We're currently only checking the pipeline depth per trigger. Instead we should check per package that we trigger builds for.

If we raise the package limit (currently 1) on the candidate selection query we could otherwise trigger for multiple packages and go over our hard limit.

finestructure:
Fixes #643


finestructure:
Good to go now :)

PR #650, opened 2020-08-18
Package update

### LINKED ISSUE #648: Vapor 4.15.2 breaks build trigger syntax
We're a little behind with our dependency updates and it turns out that Vapor 4.15.2 (current is 4.28.0) broke 5 build trigger tests.

The issue is this change: https://github.com/vapor/vapor/pull/2432

> Percent-encoded square brackets are now parsed as array or dictionary delimiters in query strings. For example, both of the following query strings are now equivalent:
>
> ```
> page[offset]=0&page[limit]=50
> page%5Boffset%5D=0&page%5Blimit%5D=50
> ```

Now this could just be a matter of adjusting the test cases but it could also mean that we'd be sending

```
variables%5BBUILD_PLATFORM%5D=ios
```

instead of

```
variables[BUILD_PLATFORM]=ios
```

and have our build triggers rejected by Gitlab. Needs testing.

We also need to bump our dependencies regularly to make the changeset smaller.

finestructure:
Fixes #648 

The reason I actually looked into this was that if we go with option 2 for the admin auth (JWT), we need to add that package and in the course of that we'd bump revisions on all the others (unless we pin them).

The admin API I think is going to be pretty important shortly after launch, because I suspect we may get a few requests to re-run builds and I don't want to fiddle around directly in the db too much.

Also, if it was _only_ about the admin API I wouldn't have bothered but we've gone quite long without package updates, so there's that as well.

Everything is back in working order now but there's a change to our outgoing build trigger request format. We used to send `variables%5BFOO%5D=bar` but now we're sending `variables[FOO]=bar`. Ironically, I _thought_ we've been sending the latter all along.

In any case, Gitlab seems to support both. Here's a trigger I ran with the `postTrigger` method now using the `PostDTO` struct to package up the data instead of `[String: String]` (which is percent encoded):

Screenshot 2020-08-18 at 11 51 08

Still working fine.

Now we still may want to hold off on merging this right now but I wanted to mention why I wrapped that up at this time and that it's influenced by how we're dealing with admin auth.

### DIFF
--- a/Package.resolved
+++ b/Package.resolved
@@ -6,8 +6,8 @@
         "repositoryURL": "https://github.com/swift-server/async-http-client.git",
         "state": {
           "branch": null,
-          "revision": "037b70291941fe43de668066eb6fb802c5e181d2",
-          "version": "1.1.1"
+          "revision": "c9a9bf061d713c91ee9974fa8a6afe413acfd0e9",
+          "version": "1.2.0"
         }
       },
       {
@@ -24,8 +24,8 @@
         "repositoryURL": "https://github.com/vapor/console-kit.git",
         "state": {
           "branch": null,
-          "revision": "9b5842b47be1a3164a42811613dce09bf5bf1f91",
-          "version": "4.1.3"
+          "revision": "7cf8185ad62d50ae9777ce78bcfde8f5c9f900e2",
+          "version": "4.2.1"
         }
       },
       {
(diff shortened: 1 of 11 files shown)


daveverwer (approved):

finestructure:
It triggered fine on staging, will now tag and push to prod

https://gitlab.com/finestructure/swiftpackageindex-builder/-/jobs/693401565

PR #654, opened 2020-08-19
Avoid empty badge message

finestructure:
Fixes #651 

Merge after #653

Issue #621, opened 2020-08-12
Admin API


finestructure:
Add an admin interface to schedule packages/versions for rebuilding (by wiping their existing build records)

Things we'll need:

- [ ] API admin auth mechanism
- [ ] API to delete builds for a given version
- [ ] API to delete builds for a given package
- [ ] API to trigger missing builds for a package
  - [ ] including override flag to bypass pipeline capacity check
- [ ] API client to perform these actions

daveverwer:
Is the plan still to do this through the authenticated API and wrapper client app?

finestructure:
Yes, exactly!

finestructure:
Regarding then admin auth mechanism, we've got a few options here

1. add an admin token just like we do for the builders
2. create a users table where we configure admins and add a login endpoint that issues JWTs
3. integrate with GH

#1 is simple - we just create another `TokenAuthenticator` and inject the token value via secrets per env. While it's simple, this doesn't scale very well in two dimensions: if we need further scopes we'll end up creating more and more token types with more and more injected secrets.

Also, we have only one token per scope. For the builders that's fine but for admins it's not great in the long run.

#2 is a bit more complex than #1 but not terribly so. The main problem is that we're stepping into the realm of PII (although arguably for a very small and consenting "crowd") and keeping data that isn't derived.

#3 I don't know how to do (yet). Would take some research. We'd also need to record what accounts are admins somewhere (presumably in the db), so there's also the aspect of non-derived data.

finestructure:
I think I'm leaning towards option 2 at the moment, mainly because we'd be sticking what is essentially per user config in the secrets and it's already cluttered there. We will soon also need a prometheus monitoring endpoint and that will require another token in the secrets. It's getting crowded.

Option 1 would be ok as well, but I think we should only do it that way if we intend to keep it. If it's transitional, I'd rather make the extra effort and go for option 2.

Option 3 I'm not super happy with mainly because I can't judge how complex that is right now _and_ it ties us to a vendor.

daveverwer:
> Option 3 I'm not super happy with mainly because I can't judge how complex that is right now _and_ it ties us to a vendor.

Of all of the "Sign in with ..." GitHub is the only one I've ever used. I trust them, and they're not going anywhere so I don't fear being tied to *that* vendor very much. It's definitely more work though.

[...]

PR #655, opened 2020-08-19
Delete builds command

finestructure:
Merge after #654 

I realised last night that there's a quicker intermediate route to adding some safe admin functionality for re-building (step towards #621 ).

It is very simple and safe to run a command attached to a container - I've been doing this to manually trigger the popular packages. Rather than build all of an auth layer, an admin api, and a controlling app (or relying on http triggers via rester or so), we can just use a `delete-builds` command from that same container to get us started.

So in order to delete all builds for a package, run

```
# attach to trigger builds container
docker exec -it spi_trigger_builds_1 bash
# run deletion
./Run delete-builds --env dev --package-id 7eabed2b-96dd-4a1a-9d78-5fe4b5cb1d62
```

or

```
./Run delete-builds --env dev --package-id 7eabed2b-96dd-4a1a-9d78-5fe4b5cb1d62 --latest default_branch
```

- or `--latest release`, `--latest pre_release`.

Or just a specific version's builds

```
./Run delete-builds --env dev --version-id 5ca2c9b1-8fc1-446f-ae32-c4f88ebb6db2
```

This should get us going while we build the other things out.


daveverwer (approved):

daveverwer:
I really love the idea that this could remove the need to think about authentication for the API so quickly. I think having to do that change is less than ideal at the moment when we could be working to improve the user-facing side of the site.

I'm all in on this 👍

PR #684, opened 2020-08-26
Tweak downscaling boundary

finestructure:
Merge after #683


daveverwer (approved):
The most subtle of tweaks! I'm sure there's a reason for this 😂

finestructure:
Just that 0 should mean no triggers - but random could roll a 0 of course :)

Noticed it when adding the same thing in the builder where we actually use 0 as an off-switch.

PR #690, opened 2020-08-31
Db backup

### LINKED ISSUE #663: New db backup strategy
- backup is too heavy now - takes 10mins and causes outage
- 270 MB compressed file

Need a new strategy.

Would sending it uncompressed work and then compressing it on the receiving end? Or is the dump server-side itself the issue?

We could copy the db files directly from the container and then either have another instance perform the dump or use the files directly. Probably the former.

finestructure:
Fixes #663 

This uses a runner to dump the db file system into a tar file and then starts up a new db process with that tar file as its db and exports it via pg_dump. This avoids the downtime on the prod db and seems to work file despite the raw file based copy.

What it currently doesn't do is ship the backup somewhere online - it simply copies the dump file to ~/Desktop/DB-Backups on prod.

NB: These dump files can be used locally exactly the same way we used to do it before we launched builds.

We should move these not only to keep them safe but also to avoid cluttering up prod. Just not sure where that should be at the moment.

### DIFF
--- a/scripts/convert_to_db_dump.sh
+++ b/scripts/convert_to_db_dump.sh
@@ -11,12 +11,12 @@ DUMPFILE=$(basename $TARFILE).dump
 
 tar xvfz $TARFILE
 
-docker run --rm --name backup-db -d -v "$PWD/db_data:/var/lib/postgresql/data" -p 9432:5432 postgres:12.1-alpine
+/usr/local/bin/docker run --rm --name backup-db -d -v "$PWD/db_data:/var/lib/postgresql/data" -p 9432:5432 postgres:12.1-alpine
 
 echo "Exporting ..."
 
 RETRIES=20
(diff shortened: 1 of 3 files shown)


scripts/convert_to_db_dump.sh
```diff
@@ -0,0 +1,34 @@
...
+[[ -n $DB_BACKUP_DIR ]]      || (echo "DB_BACKUP_DIR not set" && exit 1)
+[[ -n $ENV ]]                || (echo "ENV not set" && exit 1)
+
+echo "Unpacking ..."
+
+tar xfz $TARFILE
+
+echo "Launching backup db ..."
+
+docker run --rm --name backup-db -d -v "$PWD/db_data:/var/lib/postgresql/data" -p 9432:5432 postgres:12.1-alpine
```
finestructure, line 18: This makes the backup db briefly available on port 9432 (with a strong password of course), until the backup job is complete. We probably avoid this if we ran the pg_dump in a container as well and had them just talk on the container network.

finestructure: Fixed, just couldn't live with it 🤣

daveverwer (approved):

PR #695, opened 2020-09-01
Delete local tags, making room for upstream tag moves

### LINKED ISSUE #693: Tag move not being detected
As discussed in #692, when packages have changes force pushed into their repositories, the default branch rebuilds without issues. 👍 This is good!

However, if the force pushed changes include re-targeting of a tagged version, we do not invalidate those builds.

Even though it's going to be unusual to have a force push re-target a tagged version, we should probably invalidate builds when this happens.

finestructure:
Fixes #693


daveverwer (approved):

daveverwer:
That's a nice quick fix for it! 👍

PR #697, opened 2020-09-01
Fix beta version sorting, add tests

### LINKED ISSUE #188: Don't include beta versions unless they are semantically greater than the stable version.
In the SwiftPM Library, I only consider a beta version to be significant if it's a greater semantic version than the latest stable version (with the .pre part removed).

For example, in the SwiftPM Library:

Screenshot 2020-05-27 at 20 54 51@2x

In the SPI:

Screenshot 2020-05-27 at 20 55 53@2x

Before considering a beta version significant, we need to verify that it's semantically greater than the latest stable.

finestructure:
Fixes #188 

I've also fixed this sorting issue in https://github.com/SwiftPackageIndex/SemanticVersion and will push a release there and then switch the server to use it in a follow-up PR.

NB: this is not purely a display issue - all packages with this issue need to come around for re-analysis to get the fix.

After running analysis for Arena locally the versions have collasped as expected:

Screenshot 2020-09-01 at 11 48 06


finestructure:
Converting to draft for now so we don't merge and deploy to dev nor prod until the open issue around 2.0.4/2.0.5 is resolved (#698)

PR #703, opened 2020-09-07
Re-generate snapshots (macOS 11 Beta 20A5364e, Xcode 12b6 12A8189n)

finestructure:
This is a weird one. I'm pretty sure the snapshots passed for me locally last week. What I did over the weekend was move my Big Sur install from the external SSD to the internal one by reinstalling the Big Sur Beta and running Migration Assistant.

Now for some reason the snapshots run with this miniscule diff in the "re" or "Alamofire". Did the reinstall on the internal SSD for some reason mess with font rendering?

Do the snapshots pass for you on Catalina/Big Sur on this branch?

Screenshot 2020-09-07 at 07 11 15


finestructure:
I'll boot back into the external SSD Big Sure install I was using all along later to see if there's a delta.

daveverwer:
I just tested these and the snapshots on `main` pass, and the ones on this branch fail in `test_PackageShowView`, `test_PackageShowView_emoji_summary`, `test_PackageShowView_unknown_license`, `test_PackageShowView_incompatible_license`, `test_PackageShowView_no_authors_activity`, and `test_PackageShowView_single_row_tables`.

finestructure:
Right, I've rebooted into the external SSD and the same 6 tests are failing there as well. So it's just that I didn't run them on Big Sur beta 6 before moving to the internal SSD. That makes a lot more sense.

I'm presuming you tested on Catalina.

I'll just live with the fact that they won't pass on BS until we switch over completely and use Ceres in case I need to regenerate snapshots.

TBH, I was surprised we didn't see any snapshot deltas earlier in the beta cycle. It's always been happening on macOS major version bumps (sometimes on minor ones as well) when little aliasing details changed.

I might give the precision parameter a quick try. `SnapshotTesting` allows you to give the tests a little leeway.

daveverwer:
> I'm presuming you tested on Catalina.

Yep, latest Catalina with Xcode 12 beta 6.

finestructure:
Using `precision: 0.999` makes the tests pass - but they take 18s instead of < 1.5s. That's not worth the convenience.

finestructure:
That's _each_ - so that'd be almost 2mins just for those 6 tests.

PR #707, opened 2020-09-08
Fix double digit releases

### LINKED ISSUE #706: Release sorting fails for double digit builds
Screenshot 2020-09-07 at 17 08 02

finestructure:
Fixes #706 

Requires an analysis pass to fix up the pre-releases (because we save the latest significant releases during analysis).

Here's today's prod dump running locally after new analysis:

Screenshot 2020-09-08 at 08 31 21

### DIFF
--- a/Sources/App/Models/Package.swift
+++ b/Sources/App/Models/Package.swift
@@ -100,7 +100,8 @@ extension Package {
     static func findPreRelease(_ versions: [Version], after release: Reference?) -> Version? {
         versions
             .filter { $0.reference?.semVer != nil }
-            .sorted { $0.reference!.semVer! > $1.reference!.semVer! }
+            .filter { $0.commitDate != nil }
+            .sorted { $0.commitDate! > $1.commitDate! }
             .first {
                 // pick first version that is a prerelease *and* no older (in terms of SemVer)
                 // than the latest release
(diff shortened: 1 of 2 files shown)


Sources/App/Models/Package.swift
```diff
@@ -100,13 +100,14 @@ extension Package {
     static func findPreRelease(_ versions: [Version], after release: Reference?) -> Version? {
         versions
             .filter { $0.reference?.semVer != nil }
-            .sorted { $0.reference!.semVer! > $1.reference!.semVer! }
+            .filter { $0.commitDate != nil }
+            .sorted { $0.commitDate! > $1.commitDate! }
```
finestructure, line 104: Only doing this for `findPreRelease`, because `findRelease` is correctly sorting by semantic version `(Int, Int, Int)`, which is arguably a stronger criterion than commit date.

Where this would fall down is if someone did this:

|     |
|-b2  |
|     |
|     |-b3
|     |
|-b1 /
|   /
|  /
| /
|/

🤷‍♂️

daveverwer: Yea I think that situation is going to be very rare. I can't think of a good reason anyone would do this (which doesn't mean people won't do it, but it does mean we're allowed to care a little less about it 😂)

daveverwer (approved):
It's not ideal to base this off date, but given how flexible people can be with the pre part, I think it's the only choice. 👍

daveverwer:
Deploying as 2.0.12.

PR #719, opened 2020-09-25
Show s3 logs

finestructure:
Logs on `dev` and `prod` are now all available on S3 and should display just as before with these changes.

Now live on staging:

https://staging.swiftpackageindex.com/builds/765B4D30-471A-4B41-AC9D-EAD5AC4521F9

Screenshot 2020-09-25 at 11 00 01

### DIFF
--- a/Sources/App/Controllers/BuildController.swift
+++ b/Sources/App/Controllers/BuildController.swift
@@ -5,32 +5,14 @@ import Vapor
 
 struct BuildController {
 
-    @available(*, deprecated)
-    func _show(req: Request) throws -> EventLoopFuture<HTML> {
+    func show(req: Request) throws -> EventLoopFuture<HTML> {
         guard let id = req.parameters.get("id"),
               let buildId = UUID.init(uuidString: id)
         else { return req.eventLoop.future(error: Abort(.notFound)) }
 
         return Build.query(on: req.db, buildId: buildId)
-            .map(BuildShow.Model.init(build:))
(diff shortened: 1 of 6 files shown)


Sources/App/Controllers/BuildController.swift
```diff
@@ -11,7 +11,11 @@ struct BuildController {
         else { return req.eventLoop.future(error: Abort(.notFound)) }
 
         return Build.query(on: req.db, buildId: buildId)
-            .map(BuildShow.Model.init(build:))
+            .flatMap { build in
+                Build.fetchLogs(client: req.client, logUrl: build.logUrl)
+                    .map { (build, $0) }
+            }
+            .map(BuildShow.Model.init(build:logs:))
```
finestructure, line 18: I've tested this locally and the performance is fine.

Sources/App/Models/Build.swift
```diff
@@ -32,6 +32,7 @@ final class Build: Model, Content {
     @Field(key: "job_url")
     var jobUrl: String?
 
+    @available(*, deprecated)
```
finestructure, line 35: Marking the `logs` field deprecated for now. Will ditch the whole column in a follow up PR once we're happy with the result.

daveverwer (approved):
Just the one change to those information messages. If you're happy with my change, this is good to go!

finestructure:
Excellent - I'll merge and deploy and follow up with the clean up steps.

PR #720, opened 2020-09-25
Remove `migrate-logs` command

finestructure:
Did its job, off we go.


daveverwer (approved):
So long Soto, we barely got to know you. 😢

PR #721, opened 2020-09-25
Remove logs column

finestructure:
This completes the S3 migration on the server side.

I've run the migration locally against a prod copy and it's nearly instant - so there's no disruption to expect. We may need to run a vacuum to free up space on the db, although that might be automatic. We can just check the backup size in the morning.

⚠️ Schema change: tag as 2.3.0 when the time comes ⚠️


finestructure:
Perhaps we may want to hold off on merging this until we've heard back re: the AWS open source funding.

daveverwer (approved):
Approving this so it's ready to go, but let's wait to hear about the AWS thing.

PR #736, opened 2020-10-05
Invalid byte sequence

finestructure:
Errors have started again immediately after merging the nightly that adds back `swift-atomics`. This should fix it until we remove the logs from the db.


finestructure:
I'll self-merge this when CI has passed so we can avoid racking up too many rollbar items against our quota.

PR #747, opened 2020-10-12
Funding links

daveverwer:


Public/deploy/robots.txt
```diff
@@ -1,6 +1,6 @@
 # All robots allowed
 User-agent: *
-Disallow:
+Disallow: /builds/*
```
daveverwer, line 3: It must have been pure coincidence that the search engines finished their index as we made this change, because it didn't go live 😂

finestructure: It didn't? But I saw the diff?

finestructure: d8ea75724e261f091a416273ef57f25e646de887

finestructure: Oooh, it's not in `deploy`

finestructure (approved):

Public/deploy/robots.txt
daveverwer: Yep 😬

daveverwer: I checked it on the live site too, it's not there.

PR #765, opened 2020-10-31
Copy dumpfile, not the intermediate tarfile

finestructure:


daveverwer (approved):

daveverwer:
Oops just noticed the checks hadn't finished before I hit merge! 

This isn't tested, so it won't be an issue here. Sorry though!

PR #767, opened 2020-10-31
fetch readme

finestructure:
Fetch and save readme url from Github. It's wired through to the PackageShow.Model, ready for use in pages.

⚠️ Schema change ⚠️


daveverwer (approved):
Let's let this deploy to dev and propagate data for a bit before pushing to prod.

PR #775, opened 2020-11-09
Monitoring

### LINKED ISSUE #754: Build monitoring
The last couple of days we're been hovering at our max build threshold longer than usual. *)

This effectively looks like we're barely able to keep up with builds at the moment. I believe this isn't quite the case, however. For one, we've had issues with prod that made us lag and it looks like yesterday has been unusually busy with builds.

There are too things that are concerning however:

- The CI stats view on Gitlab doesn't seem to be working anymore for us, probably because we have too many builds now. That means we do not know how busy our build system is, other than me having the build monitor app running and eyeballing the status. That's not great.

- We seem to be having more builds pending in Gitlab than are pending in the db. Right now, it's 307 vs 76 (both prod and dev combined). I believe the reason this is happening is that we've got builds older than 4 hours in the queue and the backend has come around and deleted the corresponding pending builds in the db. That's a problem, because they will now be rescheduled and we do double work.

  I removed Ceres as a linux builder on Monday and by the looks of it the delta of 200 is due to a Linux backlog. That would make sense, because this morning dev wasn't picking up Linux builds and it was down to a single Linux queue on prod to do all the processing. Also, we've probably had some backlog there in particular, due to the downtime on Monday.

  Ceres is back at it now but I think there's a few things we need to address soon:

- we need build monitoring, in particular
  - builds per day
  - ideally also per platform
  - if possible also duration
- we may need to increase the build trimming grace period from 4 hours
- we may need a little more headroom with build capacity soon (but best to get a view first)

*) by that I mean the 300 pending in the queue beyond which we don't schedule new builds. I.e. the closest we get to our hard limit of 500 is when we're at 299 and schedule a package with the maximum number of 96 builds

Screenshot 2020-10-22 at 11 18 55

finestructure:
Basic monitoring setup with Prometheus and Grafana (which we already had running):

Screenshot 2020-11-09 at 16 09 51

I've published Grafana on port 3000 but we could obv also keep the port closed and use the SSH tunnel to attach. Not sure if that's worth the hassle though, it is a password secured instance (nor does it contain anything sensitive).

What we may want to do is configure Cloudflare to proxy port 3000. Right now, you have to use its IP + ":3000" to get access.

Fixes #754


docker-compose.yml
```diff
@@ -115,27 +114,42 @@ services:
     restart: always
 
   grafana:
-    image: grafana/grafana:7.0.0
+    image: grafana/grafana:7.3.1
     environment:
       GF_SECURITY_ADMIN_PASSWORD: admin
       GF_USERS_ALLOW_SIGN_UP: 'false'
```
finestructure, line 120: This is only the initial launch password to get in. I've alread connected and changed it.

finestructure:
Back to draft, there are a couple of tweaks I want to get in.

daveverwer (approved):

Issue #410, opened 2020-06-15
Display README on package page


daveverwer:
Depends on #409

emma-k-alexandra:
Looking forward to this addition. README and docs are probably the most important factor to me when choosing a package.

Sherlouk:
Is it valuable to report some broken READMEs or would you rather wait until more improvements are made?

Another one here which is a bit funky: https://staging.swiftpackageindex.com/appcompany/swift-sky

The top image spans much wider than the site's layout (so we would need to enforce a max-width or something?) there's also a button at the bottom which is a URL with a HTML image inside of it which doesn't get handled.

Source for comparison: https://github.com/appcompany/swift-sky/blob/master/README.md

daveverwer:
Max width is, of course, something we'll handle. No styling at all has been done yet, so that's expected. Image stuff too I'm fully expecting to need styling. 

That said, please do! I need to know what's broken so I can build a list of what needs handling.

Sherlouk:
https://staging.swiftpackageindex.com/alamofire/alamofire#communication

* Anchor links do not work
* Checkmarks will probably want a style at some point
* There seems to be an issue with detecting the closing of the code block (See Alamofire and you'll quickly see what I mean!)

https://staging.swiftpackageindex.com/moya/moya

* Relative URLs do not work (both links and image URLs etc)
* Fails soon as it gets to the table (doesn't render anything below that point)

daveverwer:
Implemented

PR #779, opened 2020-11-12
Show readme

finestructure:
Largely unstyled display of Markdown-rendered README.

- use `?readme=0` to switch it off
- gated to show only on `development` env either way

Screenshot 2020-11-12 at 12 45 11

Our H1 displays smaller that our H2, I think, which makes the titles look a bit weird right now.

Speed seems fine running it via my coworking network connection.


daveverwer (approved):
> Our H1 displays smaller that our H2, I think, which makes the titles look a bit weird right now.

Yea, when we roll this out we'll have a `readme.scss` at the same time to namespace some rules for styling inside the readme. There will be lots of edge cases too!

This looks great 👍 but, I'm not sure it "Fixes" #410 yet as I'd like to merge this without marking that as closed. It needs styling, not gating to staging, etc... I'll merge this now and then re-open that issue.

finestructure:
Part of #410

Sherlouk:
Will it take some time for the URLs to get updated after changed to raw/download URL? Package below isn't rendering correctly just yet but might be queued for an update!

https://staging.swiftpackageindex.com/sindresorhus/CircularProgress
https://raw.githubusercontent.com/sindresorhus/CircularProgress/master/readme.md

image

finestructure:
Well spotted, that's probably an issue with the Markdown renderer. We probably need to use one that's fully CommonMark compliant. If the url wasn't updated it'd show raw html instead (and the recent packages will definitely be updated as well).

Sherlouk:
Interesting point yeah - I suppose we need to take into account the "GitHub Flavoured" version too (not sure what the delta is between CommonMark and GitHub's version though)

daveverwer:
> We probably need to use one that's fully CommonMark compliant. 

If only there was a website where we could search all available Swift packages to see if there was a CommonMark compatible package. Maybe it could also show how well maintained it was, and possibly even show the READM.... Wait... 😂

daveverwer:
In all seriousness though, I don't see one which supports GitHub Flavoured Markdown in the index. There are several that wrap cmark though.

Sherlouk:
I know Ryan Nystrom started working on a GitHub Flavoured package for GitHawk but not sure it had 100% coverage either 🤔 

I think he mentioned their new app uses HTML for rendering it in the main GitHub so might be worth reaching out and seeing if he can point you in a good direction?

daveverwer:
Yea I found that. It's a couple of years out of date, doesn't support SPM yet, and is marked as WIP. But if we decide we do need to support GFM, that's where we'll start.

https://github.com/GitHawkApp/cmark-gfm-swift

PR #781, opened 2020-11-13
RSS feed fixes

### LINKED ISSUE #774: Add whitespace after title in RSS feed
Not sure if this displays similarly in other RSS readers but we could tack on a ` ` after the version number to separate these.

516382A9-5454-413C-9CDD-C48C8EF8C29F copy

finestructure:
Fixes redundant `Package - version` display and links to release url.

Screenshot 2020-11-13 at 10 45 37

Fixes #774 

⚠️ Schema change ⚠️


Sources/App/Core/RSS.swift
```diff
@@ -86,26 +80,20 @@ extension RecentPackage {
...
             .link(link),
             .pubDate(releasedAt, timeZone: .utc),
             .description(
-                .p(
-                    .a(
-                        .href(link),
-                        .text(packageName)
-                    ),
-                    .element(named: "small", text: " – \(version)")
-                ),
```
finestructure, line 104: This was the redundant part of the RSS item. The link is already in the `.link(link)` item above (and it pointed to the SPI package).

Sources/App/Models/Package.swift
```diff
@@ -140,6 +142,36 @@ extension Package {
...
+        switch (hostingProvider, reference) {
+            case let (.github, .tag(_, tagName)):
+                return "\(url)/releases/tag/\(tagName)"
+            case let (.github, .branch(branchName)):
+                return "\(url)/tree/\(branchName)"
+            case let (.gitlab, .tag(_, tagName)):
+                return "\(url)/-/tags/\(tagName)"
+            case let (.gitlab, .branch(branchName)):
+                return "\(url)/-/tree/\(branchName)"
+        }
```
finestructure, line 156: This is more for illustration for if/when we add other providers. I wanted to avoid hard-coding the release urls in the RSS feed itself. The Gitlab ones are the actual urls it uses.

finestructure:
Looks like Reeder can't be fixed to keep whitespace between elements (at least not if it's a simple `" "`) but the version string pushes that out of the view at least 😅 

Screenshot 2020-11-16 at 11 17 43

daveverwer (approved):

Issue #260, opened 2020-06-03
Add owner package list


finestructure:
Now that we've got `/owner/repo` link support it would be nice if we presented all packages beloging to a given owner under `/owner`.

finestructure:
Fixed via #782 - thanks, James!

PR #782, opened 2020-11-13
Create Author's Packages Page

Sherlouk:
Issue: #260 

Added a new page available at `url.com/:owner` which shows the packages from that user in order of score (most popular at the top). It currently just shows the title and description with a link to each package but could be expanded to include more data should it be desired.

image


Sources/App/Views/Author/AuthorShow+View.swift
```diff
@@ -0,0 +1,39 @@
...
+        let model: Model
+
+        init(path: String, model: Model) {
+            self.model = model
+            super.init(path: path)
+        }
+
+        override func content() -> Node<HTML.BodyContext> {
+            .div(
+                .class("author recent"), // 'recent' is temporary while I figure out how to regenerate the CSS file from SCSS
```
Sherlouk, line 17: Looks like you need CodeKit3 (paid app) in order to generate the CSS from SCSS? If someone else could give us a hand here I think the SCSS is right!

Sources/App/Views/Author/AuthorShow+View.swift
```diff
@@ -0,0 +1,39 @@
...
+                .div(
+                    .h2(.text("Showing \(model.count) packages from '\(model.owner)'"))
```
Sherlouk, line 19: Probably aught to pluralise this count so it says '1 package' or '2 packages'

daveverwer:
I did some styling work on this. I'd upload screenshots, but GitHub is not letting images be uploaded right now for me. It looks amazing though... haha 

@finestructure Both James and I are on Catalina still, so can you check this branch out and generate snapshots please?

daveverwer (approved):
This is good from my side. @finestructure if you're happy with the code then please merge.

daveverwer:
One interesting thing on this is that there's no navigation to this page, people will just guess the URL. That's fine, but we'll want to have some way to get here eventually. I have some ideas on this, along with some other navigation issues we have in the package right now, this is fine until I address those navigation issues. Tracked as #783.

Sherlouk:
Some images 😄 

image
image

finestructure:
Oh nice, that looks great! I'll generate some snapshots, you laggards 😆

finestructure:
I think we'll need to push this branch into the SPI repo to trigger the CI action. I'll do that now.

Sherlouk:
Thanks Sven! 🙌

Issue #523, opened 2020-07-12
Post package releases to the @PackageFirehose account


daveverwer:
We have two Twitter accounts for tweeting package updates and site updates. The accounts are:

* [@SwiftPackages](https://twitter.com/SwiftPackages) - Which will tweet edited highlights of package releases, and site news and updates.
* [@PackageFirehose](https://twitter.com/PackageFirehose) - Which will tweet every release of every package as the index discovers it.

This issue deals *only* with the @PackageFirehose account, automatically tweeting package releases.

The tweets should be driven by the same mechanism that drives the “all releases” RSS feed, here: https://swiftpackageindex.com/releases.rss.

Every time we discover a new package release, whether major, minor, patch, or pre-release, we should post a tweet in the following format:

```
[Package name] just released [Package version] - [Package summary] https://swiftpackageindex.com/[Repository owner]/[Repository name]
```

Maybe @finestructure can chime in on the best approach to where the code that posts the tweets could live.

daveverwer:
The @PackageFirehose Developer account just got approved 👍 I'll send you the credentials @Sherlouk

Sherlouk:
Happy to close this one down @daveverwer? or would you rather wait until the bug fixes are sorted and it's shipped into production?

daveverwer:
Yep, we can close this. Thanks for your work on it @Sherlouk!

PR #785, opened 2020-11-14
WIP: Twitter Firehose for New Versions

Sherlouk:
Relates to #523 

### Added TwitterVapor Dependency

Okay so I was considering writing my own Twitter OAuth setup but the amount of effort it takes to create the signature felt really overly complicated so opted for a lightweight dependency instead.

I suppose this is up to you both whether or not you'd like to add a dep for this or go full internal code.

### New Environment Variables

Add the following values to your .env files - if you don't set them it will just not setup Twitter (which is probably preferred for staging/etc anyways). We might want to setup a PackageFirehoseStaging Twitter account or something?

```
TWITTER_CONSUMER_KEY=xxx
TWITTER_CONSUMER_SECRET=xxx
TWITTER_USER_KEY=xxx
TWITTER_USER_SECRET=xxx
```

### Tweet Contents

Here's an example of what it looks like in Twitter:

image

> PredicateKit just released v0.1.1 - Direct hit PredicateKit allows Swift developers to write expressive and type-safe predicates for CoreData using key-paths, comparisons and logical operators, literal values, and functions. 
>
> http://localhost:8080/ftchirou/PredicateKit

### Unsure Parts

Okay so the biggest thing I'm unsure about (and has not been tested) is the analyser part. First off the code is just grim - I wasn't really sure of a good way to add a non-blocking component to the end of the version save.

I've also not been able to test the analyser flow. I've ran the `make analyze` command but it fails because on line 67 (the start of the method I added to) it expects an array of packages which is being sent through as an empty array.

This is also very likely not the best place to have this logic.

---

Open forum really... had a bit of time so thought I might as well take a look (and cry over Twitter APIs a little)


finestructure:
Haven't looked through the changes in detail yet but just a quick note: I had started a branch `on-version-change` or something similar that provides a hook for exactly a twitter feed. It might be useful to rebase onto that one and adopting that hook.

I'll have a look in detail asap :)

Sherlouk:
Awesome thanks Sven - I'll take a look at that and see what I can do 🙌🏼

Sherlouk:
So taken a look Sven and (at least with how it's setup at the moment) I'd also need to route through the package (so we can access the summary) and the application (where the Twitter credentials are stored).

In fact, we don't currently even use the versions as discovered in the reconcileVersions function -- this was partially intentional as I didn't think we'd want to post multiple updates at the exact same time for the exact same package. I felt it was more relevant to just inform users of the latest version update for each unique package.

I could of course calculate that using the array of versions provided, but currently I just pull it off the package itself. Notably though this doesn't happen until after the current hook point (I placed mine just after that latest version update step!)

I'll give you some time to take a look around and we can figure out how I should re-implement it based on your thoughts!

daveverwer:
Just to throw a spanner in the works here, it might be worth splitting this task into two steps. The point at which we detect a new version is not necessarily the best time to tweet it. If we're detecting several versions at the same time, is the Twitter account suddenly going to tweet a hundred releases? (probably not a hundred, but you get the idea).

I wonder if it might be worth decoupling the posting of the tweet and the detection of a new version, so we could throttle/space out the tweets.

Just an idea, and I'm not saying we should do it, but I thought it'd be worth mentioning.

Sherlouk:
I suppose an important part of that comes down to what triggers a version update - logic which I'm currently unfamiliar with!

Is it likely that we'd get massive peaks of important just because that's how the ingest is setup?

If that's the case then yeah we should wire up some sort of queue - suppose the slight issue with that is balancing it so we're not spamming the API but tight enough that the queue doesn't leave a huge de-sync between the actual release and the tweet (and eventually, potentially, emails and push notifications)

finestructure:
The recent releases RSS feed shows that release frequency isn't an issue, at least at the moment:

Screenshot 2020-11-15 at 16 44 05

Since we're batching the analysis by 25 packages every 2 mins (I think), the most we'd ever see is 50 releases (if *every* package pushed a release and a beta!). I don't think that's terrible even if it did happen (as unlikely as that is - and it _is_ a firehose account after all :) ) but we could also further reduce that by running on a smaller batch every minute instead.

So I think we've got enough knobs to turn here that'll allow us to keep the scheduling simple.

Sherlouk:
I suppose one consideration to have is the rate limit on the Twitter post itself which is currently 300 per 3-hour window. Again, not an issue at the moment but it _could_ be in your example!

Package.resolved
```diff
@@ -226,6 +235,15 @@
...
       },
+      {
+        "package": "TwitterVapor",
+        "repositoryURL": "https://github.com/handya/TwitterVapor.git",
+        "state": {
+          "branch": null,
+          "revision": "a33b9e5d832a02e269c67d54a9fabdc8d2bde855",
+          "version": "1.1.0"
+        }
+      },
```
finestructure, line 246: I've had a look through these two packages and I'd be ok adding them. While they don't have any tests, they're quite small and understandable. I'm confident we could easily fork and maintain this ourselves should there be issues - plus losing the Twitter firehose isn't really critical.

I think even if we rolled our own here it'd probably be a good idea to make these separate packages, so from that point of view it's fine. Might add some tests and upstream them.

finestructure: PS: We should definitely add them to some package index for discovery 😆

Sources/App/Commands/Analyzer.swift
```diff
@@ -97,6 +97,18 @@ func analyze(application: Application, packages: [Package]) -> EventLoopFuture<V
                 .map { getManifests(logger: application.logger, versions: $0) }
                 .flatMap { updateVersionsAndProducts(on: tx, packages: $0) }
                 .flatMap { updateLatestVersions(on: tx, packages: $0) }
+                .flatMap { input in
+                    let ops: [EventLoopFuture<Void>] = input.compactMap {
+                        guard let package = try? $0.get() else {
+                            return nil
+                        }
```
finestructure, line 104: This is a bit problematic, because it'll strip off and discard errors. I look into rebasing this onto my `onVersionChange` thingie and extend it to pass in the package or something.

Package.resolved
finestructure: On second thought, I think we should drop `TwitterVapor`. It requires us to route though `Application` when all we really need is a `Client` if we send the post ourselves (which isn't all that complicated). We've got the client in `reconcileVersions` where we'd fire off the request and the credentials will be accessible via `Current`.

I'll take a look later today/tomorrow!

daveverwer:
Closing in favour of #791. Thanks @Sherlouk!

PR #786, opened 2020-11-14
Add Tools Version to Version Schema

### LINKED ISSUE #729: Parse out the Swift tools-version during analysis
It would be useful to get some statistics on tools version but we don't currently parse it out, let's run `swift package tools-version` during analysis and store it in the database.

Sherlouk:
Closes #729 unless there was an intention to display this somewhere or reveal it in the Prometheus metrics?

⚠️ Schema change ⚠️

### DIFF
--- a/Sources/App/Models/Manifest.swift
+++ b/Sources/App/Models/Manifest.swift
@@ -37,10 +37,17 @@ struct Manifest: Decodable, Equatable {
         var name: String
         var type: `Type`
     }
+    struct ToolsVersion: Decodable, Equatable {
+        enum CodingKeys: String, CodingKey {
+            case version = "_version"
+        }
+        var version: String
+    }
     var name: String
     var platforms: [Platform]?
     var products: [Product]
     var swiftLanguageVersions: [String]?
+    var toolsVersion: ToolsVersion
 }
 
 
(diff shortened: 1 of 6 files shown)


daveverwer (approved):
I'm looking forward to seeing the results of this!

Sources/App/Models/Manifest.swift
```diff
@@ -37,10 +37,17 @@ struct Manifest: Decodable, Equatable {
         var name: String
         var type: `Type`
     }
+    struct ToolsVersion: Decodable, Equatable {
+        enum CodingKeys: String, CodingKey {
+            case version = "_version"
```
finestructure, line 42: I'll need to check but I think we've got a fixture from an actual manifest that we should extend to test for this property name. I believe the existing tests that you modified injected the tools version rather than use actual output, so it's not testing the property name.

finestructure (changes requested):
I'll look into extending that test and will push the branch to SPI for CI

Sherlouk:
If you want me to look at increasing test coverage just let me know 👍

finestructure:
Nothing big really, I just wanted to make sure we record a "live" fixture with the `_version` key somewhere and decode it. Would be even better if we ran an actual `swift package dump-package` and decoded that so that we're alerted when the key changes from `_version` (if it ever will) but I didn't want to bother with that now.

I've also made `toolsVersion` optional, because we had a failure in `test_processing_pipeline` where we didn't mock that key in. While it is strictly possible to get a package dump without `toolsVersion` that's for Swift 3, which we don't support. So we wouldn't need to make that option but it's just simpler for the test.

I've open a PR with your changes and my little tweaks now - thanks again for picking this up!

PR #791, opened 2020-11-17
Twitter firehose

finestructure:
Supercedes #785 

Thanks @Sherlouk for laying the groundwork!

This is a tweet from an actual package version:

Screenshot 2020-11-17 at 14 26 02

We'll probably want to adjust the copy and url a bit - maybe sprinkle in an emoji? (Random, from a set?). We've got the following entities available without further adjustment: `Package`, `Repository`, `Version`.


Sources/App/Core/AppEnvironment.swift
```diff
@@ -75,7 +77,17 @@ extension AppEnvironment {
                 .get("ROLLBAR_LOG_LEVEL")
                 .flatMap(AppError.Level.init(rawValue:)) ?? .critical },
         shell: .live,
-        siteURL: { Environment.get("SITE_URL") ?? "http://localhost:8080" }
+        siteURL: { Environment.get("SITE_URL") ?? "http://localhost:8080" },
+        twitterCredentials: {
+            guard let apiKey = Environment.get("TWITTER_API_KEY"),
+                  let apiKeySecret = Environment.get("TWITTER_API_SECRET"),
+                  let accessToken = Environment.get("TWITTER_ACCESS_TOKEN_KEY"),
+                  let accessTokenSecret = Environment.get("TWITTER_ACCESS_TOKEN_SECRET")
```
finestructure, line 85: I've added the Blahblaher account's keys and secrets to `DEV` - `PROD` has no keys yet.

Sherlouk:
I've noticed some of the URLs (including the one in the image) is to GitHub. Is this correct? I think the initial issue said we should link them to the SPI page.

finestructure:
Yes, you're right, I wired through the package url for simplicity, expecting we'll tweak this and either link to the package or the version (or both?).

And good point about the summary length. We'll want to calculate the whole length and trim the summary if it runs too long 👍

finestructure:
Ok, I've fixed up the url and there was also a mixup with name/owner.

Screenshot 2020-11-17 at 17 13 23

finestructure:
Screenshot 2020-11-17 at 17 16 17

Sources/App/Commands/Analyzer.swift
```diff
@@ -537,3 +538,24 @@ func updateLatestVersions(on database: Database,
...
+/// - Parameters:
+///   - client: `Client` object for http requests
+///   - transaction: database transaction
+///   - versions: array of newly discovered versions
+/// - Returns: future wrapping operation(s)
+func onNewVersions(client: Client,
+                   transaction: Database,
+                   versions: [Version]) -> EventLoopFuture<Void> {
+    let posts = versions
+        .filter { $0.reference?.isTag ?? false }
```
finestructure, line 554: NB: this is redundant, because 

```swift
let semVer = version.reference?.semVer
```

in `firehostPost` will return `nil` unless it's a tag but this avoids some extra lookups and makes that clearer.

However, we may actually _want_ to post default branch changes - it's a firehose after all (not that they're all that frequent though). We could do that with a hashtag `#latest` or something so people can mute it. Or even add those in general, `#beta` & `#stable`?

Sources/App/Core/Twitter.swift
```diff
@@ -0,0 +1,101 @@
...
+
+    static func firehostPost(db: Database, for version: Version) -> EventLoopFuture<String?> {
+        version.fetchPackage(db)
+            .flatMap { pkg in
+                pkg.fetchRepository(db).map { (pkg, $0) }
+            }
+            .map { pkg, repo in
+                guard let repoName = repo?.name,
+                      let owner = repo?.owner,
+                      let semVer = version.reference?.semVer
```
finestructure, line 68: This effectively requires `reference` to be a `tag` (see above)

finestructure:
Screenshot 2020-11-18 at 08 46 05

finestructure:
Ok, I think apart from tweaks to the copy I think I've addressed all outstanding issues:

- error logging
- trimming tweets if summary is too long
- link to SPI package

Should be good to go now!

Sources/App/Core/AppEnvironment.swift
```diff
@@ -26,6 +26,8 @@ struct AppEnvironment {
     var rollbarLogLevel: () -> AppError.Level
     var shell: Shell
     var siteURL: () -> String
+    var twitterCredentials: () -> Twitter.Credentials?
+    var twitterPostTweet: (_ client: Client, _ tweet: String) -> EventLoopFuture<Void>
```
finestructure, line 30: We may also want to add an off switch to disable tweeting without having to deploy a new release. Or we just yank the credentials - but switching it on again is then more fiddly, because we need to look them up and enter them again, which is error prone.

Sherlouk:
Looks really good Sven! Awesome stuff 🙌

Sources/App/Core/AppEnvironment.swift
daveverwer: Yea, I think this is a great idea. Let's do that before we push it live so we can switch it off with an environment change, or some other switch.

daveverwer (approved):
I think the copy and layout is great, no changes from me on that side of things. This is fantastic work, and thank you for picking it up @Sherlouk!

Giving an approval now, but let's stick the kill switch in before we merge it.

finestructure:
Ok, so I went to set up the PROD app and it seems to be a newer mechanism which only has "API key", "API secret", "Bearer token" credentials, not "API key", "API secret", "Access token", "Access token secret" like I got from the DEV account.

We should register a DEV app to have the same setup as PROD before merging this so the envs align and we'll certainly need to make changes in the POST request.

finestructure:
Ah no - false alarm. The access token and access token secret are created in yet another step. The UI is different but it seems we can generate the same set of credentials with the same names. Not sure what the "Bearer token" is for but I'll give this a spin with the prod credentials.

finestructure:
Ok, we're set. Tested locally and variables set up on PROD.

Screenshot 2020-11-18 at 11 32 27

Issue #794, opened 2020-11-18
Suppress old revisions in firehose tweets for new packages


finestructure:
(or see if we can skip new packages entirely - might not be possible)

<img width="592" alt="Screenshot 2020-11-18 at 12 36 34" src="https://user-images.githubusercontent.com/65520/99525931-cf9ff080-299a-11eb-8b17-33f280741b20.png">

Sherlouk:
Yah I can see this being a bit confusing as we add more and more packages. I don't think we necessarily want to skip the package though - but instead change the message?

Rather than "owner just released package v1.0.0 - summary" it could be something more to do with "owner's package was just added - summary" or something.

I don't think the firehose has to necessarily be limited to new versions! New packages are just as valid and might pique the interest of onlookers.

Not sure the easiest/best way to detect a new package though! Check the created date on the package reference? Check how old the latest version is?

PR #796, opened 2020-11-19
Refactor function signatures

finestructure:
Refactoring things to make #794 possible. It also brings `onNewVersion` to the top of the pipeline rather than burying it inside `reconcileVersions` which wasn't great in hindsight.

Left to do:

- [x] add tests for `Array.whenAllComplete`
- [x] add/update docstrings for added/updated analyzer functions

### DIFF
--- a/Sources/App/Commands/Analyzer.swift
+++ b/Sources/App/Commands/Analyzer.swift
@@ -106,7 +106,7 @@ func analyze(application: Application, packages: [Package]) -> EventLoopFuture<V
     }
     
     let statusOps = packageResults
-        .map(extractPackages)
+        .map(\.packages)
         .flatMap { updatePackages(application: application,
                                   results: $0,
                                   stage: .analysis) }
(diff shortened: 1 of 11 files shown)


Sources/App/Commands/Analyzer.swift
```diff
@@ -83,26 +83,33 @@ func analyze(application: Application, packages: [Package]) -> EventLoopFuture<V
...
+                         packages: packages)
+                .flatMap { applyVersionDelta(on: tx, packageDeltas: $0) }
+                .map { getManifests(logger: application.logger, packageAndVersions: $0) }
+                .flatMap { updateVersions(on: tx, packageResults: $0) }
+                .flatMap { createProducts(on: tx, packageResults: $0) }
+                .flatMap { updateLatestVersions(on: tx, packageResults: $0) }
+                .flatMap { onNewVersions(client: application.client,
+                                         logger: application.logger,
+                                         transaction: tx,
+                                         packageResults: $0)}
```
finestructure, line 104: The main point of this exercise (apart from some tidying up): This refactor allows us to place `onNewVersions` at the end of the pipeline, with access to updated properties like `latest`. This will allow us to suppress versions that aren't the latest.

daveverwer (approved):
This is going to be a useful trigger as we move forward.

finestructure:
I'm loving the type safety of Swift to be able to do things like this in a morning without going absolutely crazy. This should be set now - I'll merge and then let it run a bit on staging before rolling it out.

PR #797, opened 2020-11-19
Convert to CommonMark for README display

Sherlouk:
~~Currently using [iwasrobbed/Down](https://github.com/iwasrobbed/Down) but with a view to remove this dependency in favour of just `swift-cmark` without the overhead.~~

~~Switched away from using Ink in favour of the latest [swift-cmark](https://github.com/apple/swift-cmark).~~

Switched away from [swift-cmark](https://github.com/apple/swift-cmark) in favour of [libcmark_gfm](https://github.com/KristopherGBaker/libcmark_gfm) which adds GFM support (tables, task lists, and more) to libcmark.

This resolves many of the display bugs seen in README's but is not yet perfect.

TODO:
- [x] Add tests
- [x] Maybe add environment variable to switch between cmark and ink? (To be confirmed)
- [x] Check a variety of READMEs and identify list of remaining problems.

KNOWN PROBLEMS:
- Anchor links don't work
    * This feels like a really minor feature and not something we should prioritise.
- Relative URLs don't work
  * Need to handle relative URLs for both resources and direct links. These are different but addressed the same?
  * This might require us to do some sort of RegEx to detect relative URLs and then prefix it with an appropriate URL.
- Various Styling Issues
  * Already known before this - I believe it's on Dave's to do list.
  * Need to do some CSS voo-doo to prevent images from exceeding the width of the container.
  * Blockquotes and inline code might want to be emphasised a bit differently!
  * Headers need to have their sizes updated
  * Tables need some sort of dimension!


Sherlouk:
TIL tables aren't part of CommonMark - genuinely thought they were standardised!

For the test I've dropped a markdown file into the Fixtures folder and done snapshot tests for desktop and mobile as well as just the pure HTML for the content. Markdown file isn't exhaustive but got some of the most common stuff in there (and can easily expand)

Sherlouk:
Okay so really we have three versions that could be enabled with trivial changes (it might be worth making this an environment variable?):

* Ink
* CommonMark
* CommonMark + GFM

Currently got the latter one enabled which is a huge improvement over Ink!

Going to open it for review to get some feedback from @finestructure and @daveverwer 😄

Sherlouk:
Staging | Branch
--- | ---
image | image
image | image

finestructure:
This looks great! Thanks, @Sherlouk 

I think the only reason to ship two MD parsers would be mutually exclusive feature sets. Dave and I discussed this and if cmark can handle all the other MD parsing we're doing we'd swap out Ink entirely. So we don't really need a switch to handle both.

I'll take that out then if you don't mind. I also noticed that I seem to be able to push to your branch? And it triggers CI, so I'll just fiddle around a bit this morning :)

Sources/App/Core/MarkdownHTMLConverter.swift
```diff
@@ -0,0 +1,80 @@
...
+        
+        guard let ast = tree else {
+            throw ConversionError.markdownToASTError
+        }
+        
+        return ast
+    }
+    
+    func toHTML() throws -> String {
+        let markdown = self.markdown.replaceShorthandEmojis()
```
finestructure, line 61: Reminder to check this but perhaps you know, @Sherlouk - does the `:emoji:` substitution perhaps come for free with this parser?

Sources/App/Core/MarkdownHTMLConverter.swift
```diff
@@ -0,0 +1,80 @@
...
+        let ast = enableGFM ? try createGFMCommonMarkNode(markdown: markdown) :
+                              try createCommonMarkNode(markdown: markdown)
```
finestructure, line 63: Is it even worth keeping the common one around? Looks like GFM is a straight super set, right?

Sources/App/Core/MarkdownHTMLConverter.swift
```diff
@@ -0,0 +1,80 @@
...
+        
+        guard let ast = tree else {
+            throw ConversionError.markdownToASTError
+        }
+        
+        return ast
+    }
+    
+    private func createGFMCommonMarkNode(markdown: String) throws -> UnsafeMutablePointer<cmark_node> {
+        cmark_gfm_core_extensions_ensure_registered()
```
finestructure, line 31: Do we have a link for the docs of the C-interface? Would be good to link to I think as the initialisation steps aren't really all that obvious 😆

Sources/App/Core/MarkdownHTMLConverter.swift
daveverwer: Yep, we should only have one markdown parser. If we have a reliable GFM one, that's the only one we need.

finestructure:
The other open issues link relative links etc we can address in a follow-up, I'd say. This is already much improved for viewing on staging.

Package.resolved
```diff
@@ -64,6 +64,15 @@
           "version": null
         }
       },
+      {
+        "package": "libcmark_gfm",
+        "repositoryURL": "https://github.com/KristopherGBaker/libcmark_gfm",
```
finestructure, line 69: Note to self: see how this points to the official cmark-GFM repo, i.e. how are we getting updates.

Sources/App/Core/MarkdownHTMLConverter.swift
Sherlouk: I don't believe so - I noticed in the snapshot tests even after I updated to GFM that the emojis weren't being handled!

Sources/App/Core/MarkdownHTMLConverter.swift
Sherlouk: Sorry docs? What are these? 😂

I figured this all out by basically educated guessing and combing through a couple different repos which used swift-cmark to try and get the right flow. Never found solid docs on it all or a 'simple' one-page solution.

Tests/AppTests/MarkdownHTMLConverterTests.swift
```diff
@@ -32,7 +32,7 @@ class MarkdownHTMLConverterTests: XCTestCase {
         try XCTSkipIf((Environment.get("SKIP_SNAPSHOTS") ?? "false") == "true")
         Current.date = { Date(timeIntervalSince1970: 0) }
         TempWebRoot.cleanup()
-        SnapshotTesting.isRecording = true
+        SnapshotTesting.isRecording = false
```
Sherlouk, line 35: Swear I told myself to change this before committing on about 5 different occasions and still forgot 😂  My bad!

Sherlouk:
Let me know if anything is waiting on me btw 👍  More than happy to pick up further improvements (probably going to end up looking at the relative URL issue soon enough!)

Package.resolved
finestructure: So this directly depends on github/cmark-gfm by copying sources into the repo. cmark-gfm is on 0.29.0 and so is the Swift adapter. The latest release of cmark-gfm is from 8 Apr 2019 and there've been only two upstream releases in 2019, so it's not been changing a lot recently.

Should be pretty straightforward to update this dependency ourselves even if the maintainer doesn't, and it doesn't seem likely that there are going to be many changes to begin with.

finestructure (approved):

daveverwer (approved):
Thanks @Sherlouk!

Issue #799, opened 2020-11-20
Consider using package name instead of repo name in tweets


finestructure:
<img width="192" alt="Screenshot 2020-11-20 at 06 24 02" src="https://user-images.githubusercontent.com/65520/99762607-376c4d80-2af9-11eb-8b62-d81096a9fa6c.png">

vs

<img width="460" alt="Screenshot 2020-11-20 at 06 24 12" src="https://user-images.githubusercontent.com/65520/99762683-5834a300-2af9-11eb-9a9a-673740756444.png">

If it wasn't for "Optimizely" being mentioned in the summary it'd be a pretty obscure tweet.

daveverwer:
Yea we should absolutely use the package name rather than the repo name, I actually thought we were already!

PR #800, opened 2020-11-20
Better tweets: only new versions and announce new packages

### LINKED ISSUE #794: Suppress old revisions in firehose tweets for new packages

finestructure:
Fixes #794  (only tweet about latest versions) and also tweets differently for new packages entirely (also fixing #794, in a different way):

Screenshot 2020-11-20 at 08 36 11

### DIFF
--- a/Sources/App/Core/Twitter.swift
+++ b/Sources/App/Core/Twitter.swift
@@ -80,12 +80,12 @@ extension Twitter {
                       url: url)
     }
 
-    static func versionUpdateMessage(repositoryOwner: String,
-                                     repositoryName: String,
+    static func versionUpdateMessage(packageName: String,
+                                     repositoryOwner: String,
                                      url: String,
                                      version: SemanticVersion,
                                      summary: String?) -> String {
-        createMessage(preamble: "\(repositoryOwner) just released \(repositoryName) v\(version)",
+        createMessage(preamble: "\(repositoryOwner) just released \(packageName) v\(version)",
                       summary: summary,
                       url: url)
     }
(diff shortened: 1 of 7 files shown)


Sources/App/Core/Twitter.swift
```diff
@@ -73,22 +71,47 @@ extension Twitter {
...
+                      summary: summary,
+                      url: url)
+    }
+
+    static func versionUpdateMessage(repositoryOwner: String,
+                                     repositoryName: String,
+                                     url: String,
+                                     version: SemanticVersion,
+                                     summary: String?) -> String {
+        createMessage(preamble: "\(repositoryOwner) just released \(repositoryName) v\(version)",
```
finestructure, line 88: As I mentioned in #799, I think we want `packageName` here as well, instead of `repositoryName`. Agree to change that as part of this PR?

daveverwer: Yep!

daveverwer (approved):
This is great, let's get the package name in there and get it merged.

Tests/AppTests/PackageTests.swift
```diff
@@ -770,6 +770,48 @@ final class PackageTests: AppTestCase {
...
+        XCTAssertEqual(try Package.query(on: app.db).count().wait(), 1)
+
+        // MUT & validate
+        do {
+            let pkg = try XCTUnwrap(Package.query(on: app.db).first().wait())
+            XCTAssertTrue(pkg.isNew)
+        }
+
+        // run analysis to progress package through pipeline
+        try ingest(application: app, limit: 10).wait()
```
finestructure, line 799: Good catch, @daveverwer -- look what I wrote in the comment vs what I wrote in the code 🚀

finestructure: Now outdated, it said

```swift
// run analysis to progress package through pipeline
try ingest(application: app, limit: 10).wait()
```

PR #803, opened 2020-11-22
Fix new package tweets

### LINKED ISSUE #802: Package.isNew not behaving as expected
Screenshot 2020-11-21 at 09 42 55

finestructure:
Fixes #802

### DIFF
--- a/Sources/App/Models/Package.swift
+++ b/Sources/App/Models/Package.swift
@@ -78,7 +78,7 @@ extension Package {
         case analysis
     }
 
-    var isNew: Bool { processingStage != .analysis }
+    var isNew: Bool { status == .new }
 
 }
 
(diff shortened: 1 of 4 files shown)


Sources/App/Models/Package.swift
```diff
@@ -78,7 +78,7 @@ extension Package {
         case analysis
     }
 
-    var isNew: Bool { processingStage != .analysis }
+    var isNew: Bool { status == .new }
```
finestructure, line 81: I was done adding the new field and wiring it up when I discovered this in `updatePackage` - where a I was going to reset `isNew`:

```swift
                        if stage == .ingestion && pkg.status == .new {
                            // newly ingested package: leave status == .new for fast-track
                            // analysis
                        } else {
```

🤦 

Still want to double check everything tomorrow but I think this will be all that's needed. Tests all pass, so that's a good sign

daveverwer: Thanks, past-Sven! 😅

daveverwer (approved):

PR #805, opened 2020-11-23
loki 2.0

finestructure:
I've already test-deployed this to dev. This should give us a (configurable) 7-day (168h) history of logs, across restarts (as shown in the screenshot).

I've added a couple of notes, because when I previously observed that it doesn't keep history, I believe the main issue was that I didn't notice the default 1000 line limit on logs. So it might have just been an artifact of how I used the log viewer, to a degree. What's certainly new is the log history of 168h (but see below).

The top diagram shows that we're keeping logs across restarts (and you can see nicely how the "log peaks" align with the increased output from the analyser every 5mins).

What we still need to test properly is the log history. There's a lot going on in the config and I'm not 100% I set the correct values in all the places.

Screenshot 2020-11-23 at 14 42 03

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -137,9 +137,6 @@ services:
     build: ./infrastructure/loki
     command: -config.file=/loki.yaml
     restart: always
-    # TODO: remove
-    ports:
-      - 3100:3100
 
   promtail:
     build: ./infrastructure/promtail
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -118,6 +118,8 @@ services:
       - "2222:22"
     restart: always
 
+  # logging & monitoring visualiser
+
   grafana:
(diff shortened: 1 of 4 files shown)


daveverwer (approved):

finestructure:
Right, looks like I'm not the only one who tripped over that 😆  https://github.com/grafana/loki/issues/1362#issuecomment-561696853

HristoA:
Hi.
Something that can help is to set under `Data source`->`Maximum lines` to 5000 lines or high (this can NOT be overwrites from Grafana Explorer  `Line limit` field). In this way will see a larger piece of logs in selected time frame. Keep in mind that if you goo to high or unlimited with this value you will lose stability  and is possible Loki to start crash if you have too much log lines for selected time frame.

PR #817, opened 2020-11-25
Store isArchived on repository and show on package page

### LINKED ISSUE #552: Show if a package is archived
The repo API tells us if a repo has been marked as archived/read-only by the repo owner.

https://github.com/Ch0uti/ChouTi is a known archived repo in our index.

In a discussion with Dave we believe it's best that we don't completely exclude these from the index, but is important that we highlight this information to users of the site to make a more informed decision.

Changes:
- [ ] Record a `archived` field in the repositories table and populate it from the GitHub API call.
- [ ] Reduce the package's search score (-5) if it's archived (?? - I've made this requirement up)
- [ ] Display whether a repository is archived on the Package page.

This does likely link up to the "extra package metadata" conversations which are ongoing (#435) as people may wish to label their repo as archived from there too. I believe this is an acceptable interim solution which can be evolved to include that at a later date.

Sherlouk:
GraphQL query already contained isArchived so pulled it into the Repository model and then routed through the PackageShow model.

Closes #552 

(Also updated the directory of one of the migrations just to have a leading zero - consistency and all that...!)

I feel like archived is not a word any more


daveverwer (approved):
I'll let Sven review the code, but the copy looks great! (The copy is exactly what I asked for, so it'd definitely be a YTA moment if I said it wasn't) 😅

finestructure:
LGTM - thanks, James! Will merge after CI passes.

daveverwer:
Thanks @Sherlouk! This is a great little feature.

PR #821, opened 2020-11-26
fix isArchived

### LINKED ISSUE #820: Archived clause not taking effect?
According to the `dev` db, this package is archived: https://staging.swiftpackageindex.com/vapor/template-kit (and it _is_: https://github.com/vapor/template-kit

Yet the page doesn't reflect this:

Screenshot 2020-11-26 at 11 52 24

finestructure:
Fixes #820


daveverwer (approved):
I think that we could have made it work with the `text` field. 

We could have made it so that "Yes, this repository is archived" or "Out of date, don't use" equals true and "No, this is a valid and current project" means false. All we'd need is a bit of NLP and ML to figure out the intent of the text in the field and it'd be fine. 

But if you wanna be all "computer science" about it, sure 😂

finestructure:
https://twitter.com/causal_agent/status/1328747512816144390

PR #825, opened 2020-11-27
Readme Styling Improvements

daveverwer:
Progress towards #410

This PR fixes all known issues with README rendering apart from #824. The next step is to do some larger-scale manual testing to find more bugs and edge cases.

This PR includes the following:

* CSS improvements for images, tables, blockquotes, etc...
* Syntax highlighting for code snippets.
* A client-side fix for relative images.

The highlighting uses `highlight.js` and I have stripped it down to only load support for a few languages. Loading all languages takes the page initialisation time up by over 450ms, and the subset only adds about 55ms. I determined the set of languages to support by downloading *all* README files and finding all code blocks, and sorting by most used + some common sense.

PR #827, opened 2020-11-27
Update Markdown tests to include new CSS styling

Sherlouk:
New CSS rules introduced in #825 require a certain DOM layout. This change updates the markdown tests to match that enabling the CSS rules to apply to the generated snapshots meaning we can visually confirm the styling on each markdown element.


daveverwer:
Thank you James! Great catch.

The interesting thing with this is that some of the styles (the heading styles especially) are very specifically different inside a readme than from the markdown for the docs pages.

I wonder if it's worth two tests for this, one with the `markdown` class on the `body` tag which is what we do for the markdown docs pages, and one for the readme, as you have here.

Sherlouk:
So we do technically have a test covering that already but it only covers the privacy.md which doesn't really have a broad range of markdown!

I'll add another test though which uses the same markdown test file but using the actual MarkdownPage 😄

Sherlouk:
This is also a fun way to compare the markdown support between the two pages 😅 

Is there any reason we shouldn't migrate it over to using the same GFM engine as READMEs?

daveverwer:
> Is there any reason we shouldn't migrate it over to using the same GFM engine as READMEs?

Honestly, I thought the PR that introduced the GFM Markdown parser switched the whole thing out for all pages. I'd certainly be in favour of only having one Markdown parser for the whole project. I'll merge this and raise an issue for that separately.

Sherlouk:
The main issue (I 100% didn't just find this out now as I tried to switch it out) is that the MarkdownPage uses the 'metadata' feature of Ink which pulls out some information right at the top of the file.

We'd need to implement this ourselves first.

daveverwer:
Ah right, and that's key to that page working and isn't something we should spend time implementing

Two Markdown parsers it is!

daveverwer (approved):
LGTM, @finestructure can you take a look too please?

Sources/App/Views/MarkdownPage.swift
```diff
@@ -26,6 +26,12 @@ class MarkdownPage: PublicPage {
         super.init(path: path)
     }
     
+    init(path: String, html: String) {
+        self.metadata = [:]
+        self.html = html
+        super.init(path: path)
+    }
```
finestructure, line 33: IIUC we only need this init for the tests, right? If so, we should place it in an extension in the tests. (If that’s possible, I recall that `init` extensions have some rules attached.)

finestructure:
The be honest, I don’t quite follow what we’re doing here 😅 I’m going to need a little explainer!

What does the new test do in addition/differently? I _think_ it’s about the readme being wrapped in an `article` such that headers etc can have distinct css rules, right? I think snapshotting whole pages with those different parsers is confusing for onlookers - you don’t really understand what’s being used where and why if we don’t use it in the same places, so to speak.

Maybe I’m misunderstanding but it seems to me a realistic snapshot would be:

```
all markdown elements in “outer style”
<article>
all markdown elements in “inner style”
</article>
```

That way they’re together and the purposes of comparison is clear?

Again, I might be missing the point of this!

Sherlouk:
Sorry Sven some context is definitely missing here - had a brief chat about it with Dave as part of the previous merge request.

So there's two tests that have been changed here, the first one is for the GFM markdown converter - we already had a snapshot test checking that it was correctly converting the markdown into HTML which was being displayed. The modification in this MR adds the necessary class etc so that the recent styling improvements for READMEs (where the GFM markdown converter is used) can be seen -- without it there are no tests covering the markdown styling recently merged.

The second test covers the MarkdownPage (used for some fixed page docs like privacy and FAQs) and is running the same test markdown file but through that alternative (Ink) markdown converter. This is because not only does it use a different markdown converter but also the CSS is applied slightly differently (using different classes) so is feasible a CSS rule could affect only one of the two markdown pages (FAQs/etc, and READMEs).

Hope that clarifies a bit?

Sources/App/Views/MarkdownPage.swift
Sherlouk: I believe the issue here is that for it to be in an extension we would need to use a `convenience init` which can only use other initialisers in the class (so would need to pass the `markdownFilename` in!)

This would also mean that we need to make html mutable so I can update it within the init since there's no other way to (in)directly modify it.

Would that still be better than this, do you think?

finestructure: Yeah, that's exactly what I thought might prevent making it an extension and the minor convenience - heh - of having it scoped to the tests doesn't really warrant messing with it.

finestructure:
Aah, I see 👍

finestructure (approved):

finestructure:
Thanks for the clarification and the additional tests, James!

Sherlouk:
No worries! 
Always more than happy to justify my changes (otherwise what's the point in making the change in the first place!)

Thanks for merging it in

Issue #421, opened 2020-06-20
RSS releases - show release notes


finestructure:
This would be a bigger task, as we’re not currently collecting release notes (if there are any, even), but it’d be particularly nice for the RSS feed to show you what changed at a glance inline in the feed.

daveverwer:
If we need to get the release date to fix #414 we may be able to grab the release notes in the same API call.

finestructure:
We have release notes in `versions.release_notes` in the db now, so this should be a fairly straightforward task now.

daveverwer:
Technically, yes. It does need some careful placing into the package page though. Still a good issue to tackle, but the design aspect of this task isn't trivial.

finestructure:
This is for the RSS feed though, not the package page.

daveverwer:
Oops, so it is. 😵‍💫

elfenlaid:
I'll have a look at it if everyone is okay

daveverwer:
Sounds good to me! I also had a thought of how we can fit this into the current web design too, but that's for a different ticket. 

One thought I did have is that we are currently getting the release notes as plain text, meaning we'll have to process them through a Markdown parser. That's no problem, however, we'll potentially run into the same issues we had with README files where people are using GitHub flavoured markdown and relying on GitHub syntax highlighting etc... 

We found that for the GitHub README call, we can request rendered HTML instead of grabbing the plain text README. I wonder if that's possible for the release notes too? That would make us much more compatible with and also make the implementation of web release notes much easier in the future. 

The API documentation for how to fetch README updates [is here](https://docs.github.com/en/rest/reference/repos#get-a-repository-readme) and it's specifically the part where they reference 

If that's not possible, we may just want to use GitHub's Markdown API for rendering what we already have rather than using Ink for the same reasons I mentioned above. https://docs.github.com/en/rest/reference/markdown

daveverwer:
Oops didn't mean to close it! 😬

elfenlaid:
> We found that for the GitHub README call, we can request rendered HTML instead of grabbing the plain text README. I wonder if that's possible for the release notes too? That would make us much more compatible with and also make the implementation of web release notes much easier in the future.

I think I've found the right API:

<pre><code>
$ curl https://api.github.com/repos/SnapKit/SnapKit/releases/tags/5.0.1 \
  -H "Accept: application/<strong>vnd.github.html+json</strong>"

{
[...]

Issue #414, opened 2020-06-17
Support GitHub release information


daveverwer:
This isn't a bug but is worth a discussion.

I saw @freak4pc [tweet about the release of Prism 0.4.0](https://twitter.com/freak4pc/status/1272921165238018048) yesterday and thought it would be a good little real-world test to watch that release progress into the index. I checked a couple of times yesterday, and again this morning and it never made it to the list recent releases.

We know this as we talked about it the other day, but our version/release date is based on the date of the commit, not the date when the tag was added. The release was [tagged yesterday](https://github.com/GettEngineering/Prism/releases/tag/0.4.0), but the tag was added to a commit made [15 days ago](https://github.com/GettEngineering/Prism/commit/a60191d12c19d6b3a7ab68be82545c67cf8fdbc2). By the time we saw the release, the release would never have made the list of recent packages because it only shows the last 7 releases, ordered by the commit time.

The solution to this is [annotated tags](https://www.git-scm.com/book/en/v2/Git-Internals-Git-References#_tags) which can then be queried with `git cat-file`, but annotated tags are rare. The Prism release was not an annotated tag and looking through several popular repositories I couldn't find a single one using annotated tags.

So, I don't think there's an easy fix to this, without relying on the GitHub API to query for release information. This is less than ideal as it requires yet more API calls where we are already struggling with our API limits, and makes it harder to support repositories hosted outside GitHub.

freak4pc:
I don't know the specific code but isn't this API helpful for this case? 
https://developer.github.com/v3/repos/releases/#list-releases

daveverwer:
> So, I don't think there's an easy fix to this, without relying on the GitHub API to query for release information. This is less than ideal as it requires yet more API calls where we are already struggling with our API limits, and makes it harder to support repositories hosted outside GitHub.

Yep, 100%, but as I mentioned here we're already running up against our API limits pretty hard without querying for release information. We could potentially grab it via a GraphQL aggregate query (which we're discussing in #284) but it would still make supporting new repository hosting platforms harder.

freak4pc:
Oh, sorry, I missed that last sentence. I wonder if GitHub would be open to extending your limits given this is a community-wide project. Not sure who'd be the right contact for that, but I heard of some projects that get this.

finestructure:
It looks like even _with_ this extra request we'd only capture the same info we have now unless projects use annotated tags, right?

[...]

PR #829, opened 2020-11-30
fetch releases

finestructure:
Ground work for #421, #414

⚠️ Schema change ⚠️

### DIFF
--- a/Sources/App/configure.swift
+++ b/Sources/App/configure.swift
@@ -104,6 +104,9 @@ public func configure(_ app: Application) throws {
     do {  // Migration 022 - add is_archived to repositories
         app.migrations.add(UpdateRepositoryAddIsArchived())
     }
+    do {
+        app.migrations.add(UpdateRepositoryAddReleaseNotes())
+    }
 
     app.commands.use(AnalyzeCommand(), as: "analyze")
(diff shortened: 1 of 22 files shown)


finestructure:
Good to go now!

finestructure:
Good to go now!

Sources/App/configure.swift
```diff
@@ -104,6 +104,10 @@ public func configure(_ app: Application) throws {
     do {  // Migration 022 - add is_archived to repositories
         app.migrations.add(UpdateRepositoryAddIsArchived())
     }
+    do {
```
Sherlouk, line 107: nit: do we want to continue the pattern of the migration number and comment here?

daveverwer (commented):
So I see that we're now storing a publishedAt date which is the release date rather than the date of the commit that the release is made from, that's good. I don't see anything that acts on that data yet though (for the Twitter/RSS feeds, for example), is that correct?

I don't think we should act on that data yet, we need to think about how best to approach the problem, I just want to make sure I understand the change.

daveverwer (approved):

finestructure:
Yes, that's correct. Just wiring it up and bringing in the schema change while I was touching that entity anyway.

PR #831, opened 2020-11-30
Deal with empty and blank summaries

### LINKED ISSUE #819: Trailing hyphen in Firehose tweets
https://twitter.com/PackageFirehose/status/1331644360761413638

We do have code to try and avoid this:
https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/blob/main/Sources/App/Core/Twitter.swift#L63-L69

My best guess (without debugging) is that summary exists but is an empty string (or purely contains whitespace). We might need to do something like

```
guard let summary = summary?.trimmingCharacters(in: .whitespacesAndNewlines), 
      summary.isEmpty == false 
else {
  return ""
}
```
(might not compile but you get the picture)

finestructure:
Merge after #829 

Fixes #819

### DIFF
--- a/Sources/App/Core/Twitter.swift
+++ b/Sources/App/Core/Twitter.swift
@@ -61,7 +61,8 @@ extension Twitter {
         let separator = " \(separator) "
         let availableLength = tweetMaxLength - preamble.count - separator.count - link.count
         let description: String = {
-            guard let summary = summary else { return "" }
+            guard let summary = summary?.trimmingCharacters(in: .whitespacesAndNewlines),
+                  !summary.isEmpty else { return "" }
             let ellipsis = "…"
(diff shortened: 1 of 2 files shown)


Sources/App/Core/Twitter.swift
```diff
@@ -106,12 +107,12 @@ extension Twitter {
                     ? newPackageMessage(packageName: packageName,
                                         repositoryOwner: owner,
                                         url: url,
-                                        summary: repo?.summary ?? "")
```
Sherlouk, line 109: 👀 

about that 😂

Sherlouk (approved):

daveverwer (approved):

PR #836, opened 2020-12-02
Alert setup (via Telegram)

finestructure:
Screenshot 2020-12-02 at 16 36 35

Configured for dev. Remaining prod setup:

- [x] set up telegram bot
- [x] set up token variable
- [x] set up chat id variable

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -125,6 +125,7 @@ services:
     environment:
       GF_SECURITY_ADMIN_PASSWORD: admin
       GF_USERS_ALLOW_SIGN_UP: 'false'
+      GF_SERVER_ROOT_URL: "${GF_SERVER_ROOT_URL}"
     volumes:
       - grafana_data:/var/lib/grafana
     restart: always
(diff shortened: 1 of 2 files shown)


docker-compose.yml
```diff
@@ -125,6 +125,7 @@ services:
     environment:
       GF_SECURITY_ADMIN_PASSWORD: admin
       GF_USERS_ALLOW_SIGN_UP: 'false'
+      GF_SERVER_ROOT_URL: "${GF_SERVER_ROOT_URL}"
```
finestructure, line 128: This allows us to link to dashboards or log views with filter set ups etc. For example the shutdown issue: http://208.52.185.184:3000/goto/mX1lvEAMz

finestructure: (we may want to set up some domain aliases to avoid having to use the raw IPs)

finestructure:
This is good to go, just leaving it in draft as a reminder for me to set up the prod channel if we're happy with using Telegram.

finestructure:
I should add that we can also set this up with Slack instead. I have no real preference tbh except that it would require us to join yet another workspace 😅

daveverwer (approved):

docker-compose.yml
daveverwer: Added aliases on Cloudflare.

PR #838, opened 2020-12-03
improved errors

finestructure:
This should both report more detailed errors up to Rollbar _and_ log a distinct error to the console which we can alert on whenever we're showing an error page to users:

```
[ NOTICE ] Server starting on http://127.0.0.1:8080
[ INFO ] GET /api/search [request-id: 93F433B6-3C8F-4256-BA5B-EECE78264D8E]
[ CRITICAL ] ErrorPage.View: Generic error: test error (id: 759ACFE5-93DB-4CFC-B1CD-F2E804270896)
```

This should take of a good number of the obscure errors in Rollbar but we'll have to see how many or if there are others that are reported as `AbortError` (which obscures the details).


daveverwer (approved):
Ths is great 👍

PR #846, opened 2020-12-06
Switch to using URL instead of NSString to determine base URL

### LINKED ISSUE #844: Relative URLs in README incorrect format
Images that rely on relative URLs are currently not working, at least not on [Moya](https://staging.swiftpackageindex.com/moya/moya).

The relative URL it has detected (via `data-readme-base-url`) is "https:/raw.githubusercontent.com/Moya/Moya/master" which is not a valid URL due to the scheme only having one slash instead of two. 

This means that the webpage tries to load `https://staging.swiftpackageindex.com/raw.githubusercontent.com/Moya/Moya/master/web/diagram.png` which is obviously not going to work.

https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/blob/main/Sources/App/Views/PackageController/PackageShow%2BModel.swift#L100
Interestingly we seem to use `repository.readmeUrl` - so the fact this is incorrect might be concerning?

Follow-on from #410

Sherlouk:
Closes #844 

**The test does not work**
It's crapping out when trying to run `pkg.productCounts()` due to:
>Fatal error: Children relation not eager loaded, use $ prefix to access: Children<Version, Product>(for: [version_id]): file FluentKit/Children.swift, line 33
2020-12-06 20:32:07.626463+0000 xctest[18804:5885046] Fatal error: Children relation not eager loaded, use $ prefix to access: Children<Version, Product>(for: [version_id]): file FluentKit/Children.swift, line 33

If someone wouldn't mind pointing me in the right direction that'd be much appreciated 🙏  I've locally run a test using the new approach and it seems to be a bit happier 👍  

There is one minor change that the new URL will have a trailing slash, so I've updated the JS to handle this.


finestructure:
I know why that's happening - I'll fix it tomorrow, no worries!

daveverwer (approved):
Front end change is all good from my side.

The removal of /deploy/ is strange, but doing a build fixed it. I don't know why it removed it for you @Sherlouk! CodeKit does weird stuff though, so don't worry about it.

finestructure (approved):

PR #868, opened 2021-01-06
Refactored secret comments into a standard place.

daveverwer:
Refactored the secret comments we have in some pages to be in a consistent place that's out of the way for the design changes, and easier to view in the web inspector.

### DIFF
--- a/Sources/App/Views/PublicPage.swift
+++ b/Sources/App/Views/PublicPage.swift
@@ -180,6 +180,7 @@ class PublicPage {
     final func body() -> Node<HTML.DocumentContext> {
         .body(
             .class(bodyClass() ?? ""),
+            bodyComments(),
             analyticsBody(),
             stagingBanner(),
             header(),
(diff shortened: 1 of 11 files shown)


Sources/App/Views/PublicPage.swift
```diff
@@ -191,6 +192,12 @@ class PublicPage {
             stagingBanner()
         )
     }
+
+    /// Any page level HTML comments for hidden metadata.
+    /// - Returns: An element, or `group` of elements.
+    func bodyComments() -> Node<HTML.BodyContext> {
+        .text("")
```
finestructure, line 199: could use `.empty` here (doesn't seem to change snapshots)

finestructure (approved):

Sources/App/Views/PublicPage.swift
daveverwer: Oh, empty is so much better! We use `.text("")` in a few places so I'll standardise in a separate PR.

PR #880, opened 2021-01-12
Generate package collection

finestructure:
WIP

PR #881, opened 2021-01-13
Improved Package Collection creation, using SPM models

finestructure:
This is an improved version of #880 which uses the verbatim `JSONPackageCollectionModel.V1` files from SPM.

It does however require a couple of changes in the upstream to make the model independent from SPM `PackageModel` types: https://github.com/finestructure/swift-package-manager/pull/1/files

We can go ahead and merge this as is. There's a draft PR #882 which is rebased on top of this branch that's adopting the model from an external package, in case that would be accepted.

This PR only includes the two `generate` methods to create package collections but no actual API. That's to follow.

The output generated from the snapshot tests passes validation with `package-collection-validate` (from https://github.com/apple/swift-package-collection-generator.git)


finestructure:
This is good to go now.

daveverwer (approved):

PR #888, opened 2021-01-20
Package page design changes

daveverwer:
From:

Screen Shot 2021-01-27 at 11 41 49@2x

To:

Screen Shot 2021-01-27 at 11 43 14@2x

This is a major update to the design, along with a significant clean-up of the front-end markup and CSS/images too. The idea was to bring the README further up the page, which it does, although not by much in the end! It does give us a menu area on the right-hand side which will expand into more menu options as we add features, it also makes the display of what significant versions we are dealing with much more obvious.

All comments welcome!


Sources/App/Views/PackageController/PackageShow+Model.swift
```diff
@@ -114,246 +114,209 @@ extension PackageShow {
 
 
 extension PackageShow.Model {
-    func authorsClause() -> Node<HTML.BodyContext>? {
```
finestructure, line 117: It's great that these are non-optional now! I think we had those because I didn't know about `.empty` early on.

Sources/App/Views/Utils.swift
```diff
@@ -3,11 +3,11 @@ import Plot
 
 // MARK: - Pluralisation
 
-func pluralizedCount(_ count: Int, singular: String, plural: String? = nil) -> String {
+func pluralizedCount(_ count: Int, singular: String, plural: String? = nil, capitalized: Bool = false) -> String {
```
finestructure, line 6: I'd suggest instead of an additional parameter we let the caller do

```
pluralizedCount(products.libraries, singular: "library", plural: "libraries").capitalized
```

using `String.capitalized`.

finestructure: It ignores numbers, so it's the same thing.

Tests/AppTests/PackageShowModelTests.swift
```diff
@@ -68,118 +75,80 @@ class PackageShowModelTests: AppTestCase {
...
-                       [[\.stable], [\.beta], [\.latest]])
-        XCTAssertEqual(lpInfoGroups(.init(stable: v1, beta: v2, latest: v2)),
-                       [[\.stable], [\.beta, \.latest]])
-        XCTAssertEqual(lpInfoGroups(.init(stable: v1, beta: v1, latest: v2)),
-                       [[\.stable, \.beta], [\.latest]])
-        XCTAssertEqual(lpInfoGroups(.init(stable: v2, beta: v1, latest: v2)),
-                       [[\.stable, \.latest], [\.beta]])
-        XCTAssertEqual(lpInfoGroups(.init(stable: v1, beta: v1, latest: v1)),
-                       [[\.stable, \.beta, \.latest]])
-    }
```
finestructure, line 108: note to self - is there a corresponding test for compatibility grouping?

finestructure: Just me being confused!

finestructure (approved):

PR #889, opened 2021-01-21
Version re-analysis

### LINKED ISSUE #887: Process to re-analyse versions
This is the umbrella issue to tackle re-analysis of existing versions.

In our current processing, we diff versions we find against db content, adding new ones and deleting removed ones.

While efficient, this means any analysis changes made are only reflected in newly created versions.

Currently, there are three data model and analysis changes that suffer from this:

- `tools-version` #801
- `targets` #872
- `products.type` #878 (not yet implemented)

At a high level, the process to be implemented would work as follows:

- find package candidates with versions last updated before a certain cut-off date
- feed these packages into the normal analaysis pipeline
- change version diff not to return new/deleted but existing versions
- filter those versions on "updated < cut-off date"
- get manifests for these versions
- update version fields
- update products *)
- update targets *)

The particular steps re-use the currently implemented analysis steps. This should all be refactorable such that the main difference between regular analysis and this "refresh pass" is

- candidate selection
- which versions are picked

*) "update products" and "update targets" are currently "create products" and "create targets", because they are only ever called with new versions. We would change this to "update" and to a simple reconciliation, to be used from both.

finestructure:
This provides a command `re-analyze-versions` to update existing versions.

NB: this command assumes that packages have been analysed and repository checkouts are present. This should normally be the case but in order to run this locally, you would either need to ensure you have `checkouts` populated (via full analysis) or have run `analysis` for a matching command. I.e.

```
swift run Run analyze --id F6999391-19C4-427A-B011-D70F5BD8E0C7
swift run Run re-analyze-versions --before 2021-01-01 --id F6999391-19C4-427A-B011-D70F5BD8E0C7
```

Fixes #887


Sources/App/Commands/VersionReAnalyzer.swift
```diff
@@ -0,0 +1,236 @@
...
+        let limit = signature.limit ?? defaultLimit
+
+        let client = context.application.client
+        let db = context.application.db
+        let logger = Logger(component: "re-analyze-versions")
+        let threadPool = context.application.threadPool
+
+        guard let cutoffDate = signature.before ?? Current.reAnalyzeVersionsBeforeDate() else {
+            logger.info("No cut-off date set, skipping re-analysis")
+            return
```
finestructure, line 29: This is a potential mechanism to later deploy this as an automatic command that runs on a cycle like the others. By setting `RE_ANALYZE_VERSIONS_BEFORE_DATE` to a `YYYY-MM-DD` value, we'll start processing those versions until they all have been updated.

We can leave the job running and simple set the value to `-` (or anything that isn't `YYYY-MM-DD`) or remove the key to make the process a no-op.

Alternatively, we can kick the process off manually whenever we need to without setting up an automatic process and the env variable.

Sources/App/Commands/VersionReAnalyzer.swift
```diff
@@ -0,0 +1,236 @@
...
+                            threadPool: threadPool,
+                            transaction: tx,
+                            packages: packages,
+                            before: cutoffDate)
+            .flatMap { setUpdatedAt(on: tx, packageVersions: $0) }
+            .flatMap { mergeReleaseInfo(on: tx, packageVersions: $0) }
+            .map { getManifests(logger: logger, packageAndVersions: $0) }
+            .flatMap { updateVersions(on: tx, packageResults: $0) }
+            .flatMap { updateProducts(on: tx, packageResults: $0) }
+            .flatMap { updateTargets(on: tx, packageResults: $0) }
```
finestructure, line 159: This is the essential processing pipeline. It effectively shares all logic with the `analysis` pass. There are only a handful of helper functions below which are mainly adapters to make function signatures match.

The only additions are:

- `getExistingVersions` which returns the existing versions instead of the `added, deleted` pair in `analysis`
- `setUpdatedAt` which updates `update_at` for processed versions (analysis only adds versions, so this is not needed there)
- `Package.fetchReAnalysisCandidates` which is the candidate selection specific to re-analysis

daveverwer (approved):

PR #893, opened 2021-01-22
Log more build details

finestructure:
Merge after #889 so we can benefit from its string interpolation of options, which will turn

```
2021-01-22T11:22:28+0100 info codes.vapor.application : Triggering 32 builds for package name: nil, ref: Optional(main), id: Optional(005AF96B-8214-4F35-899E-FCDD434C6EA0)
```

into

```
2021-01-22T11:22:28+0100 info codes.vapor.application : Triggering 32 builds for package name: nil, ref: main, id: 005AF96B-8214-4F35-899E-FCDD434C6EA0
```

### DIFF
--- a/Sources/App/Commands/TriggerBuilds.swift
+++ b/Sources/App/Commands/TriggerBuilds.swift
@@ -162,7 +162,7 @@ func triggerBuildsUnchecked(on database: Database,
                             logger: Logger,
                             triggers: [BuildTriggerInfo]) -> EventLoopFuture<Void> {
     triggers.flatMap { trigger -> [EventLoopFuture<Void>] in
-        logger.info("Triggering \(trigger.pairs.count) builds for version id: \(trigger.versionId)")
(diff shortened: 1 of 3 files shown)


Sources/App/Commands/TriggerBuilds.swift
```diff
@@ -240,10 +240,21 @@ struct BuildPair: Equatable, Hashable {
 struct BuildTriggerInfo: Equatable {
     var versionId: Version.Id
     var pairs: Set<BuildPair>
-
-    init(_ versionId: Version.Id, _ pairs: Set<BuildPair>) {
+    // non-essential fields, used for logging
+    var packageId: Package.Id?
+    var packageName: String?
+    var reference: Reference?
```
finestructure, line 246: I'm not in love having to tack on these properties used only for logging but there's no way to do this that I can see that'd be even worse.

daveverwer (approved):

PR #902, opened 2021-01-28
Build trigger update

### LINKED ISSUE #901: Throttle branch builds
At substantial amount of our build workload is due to default branch changes:

CleanShot 2021-01-28 at 14 21 57@2x

We need to look into ways to throttle this back a bit, ideally with a parameter based "dead time" that we can adjust as needed.

finestructure:
Fixes #901

### DIFF
--- a/Sources/App/Core/Constants.swift
+++ b/Sources/App/Core/Constants.swift
@@ -34,5 +34,7 @@ enum Constants {
     static let searchLimit = 20
     static let searchLimitLeeway = 5
 
+    // build system settings
     static let trimBuildsGracePeriod = 4  // hours
+    static let branchBuildDeadTime = 24 // hours
 }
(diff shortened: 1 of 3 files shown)


Sources/App/Core/Constants.swift
```diff
@@ -34,5 +34,7 @@ enum Constants {
     static let searchLimit = 20
     static let searchLimitLeeway = 5
 
+    // build system settings
     static let trimBuildsGracePeriod = 4  // hours
+    static let branchBuildDeadTime = 24 // hours
```
finestructure, line 39: I've made this a `Constant` instead of an env variable for now. It's just simpler for now and we can always transition it if we find we need more control.

finestructure:
Ok, this should be done in principle but I want to give this another pass myself in the morning with fresher eyes.

Please have a thorough look at the selection logic changes, I often mess up inverted boolean logic! 😬

finestructure:
Ok, I've given this another pass and I think it makes sense. Still, we should hold off merging that until Monday.

daveverwer (approved):

PR #910, opened 2021-02-01
Fix bind parameter interpolation

### LINKED ISSUE #909: Build candidates still reporting old count
The new build candidate query selects 5 candidates as of this moment, while the old one selects 7.

Monitoring still reflects the old count:

CleanShot 2021-02-01 at 09 20 57@2x

Need to investigate what drives monitoring (should be the same source) or why this isn't being picked up yet. 2.19.0 has been live for 25mins now, so it should all be reporting up-to-date figures.

finestructure:
Fixes #909


finestructure:
Self-merging, as discussed

PR #920, opened 2021-02-05
Address build delays

### LINKED ISSUE #916: Default branch build gaps after throttling change
Analysis deletes versions on changes and we lose the associated builds. Since we now delay default branch builds until the next day, we've created a gap in the build matrix and show "?" until the builds kick in.

I believe we've got the following options to address this

1. Roll back change to build trigger
2. Introduce additional tracking fields to support latest version display on package page during cool-down period and keep version with builds
3. Introduce a new “intraday latest branch” version for the purpose of the package page display that is without builds but won’t get built until the cool-down period has expired. Then roll over once that build has happened.
4. Roll back build throttling and introduce a longer version update window during analysis for branch changes

I think there are only two non-hacky ways to do this:

- Option 1 - roll back and pay the build price
- Option 4 - roll back build throttle + longer version update window

I’m leaning towards 4. The way this would work is that during analysis we ignore changes to default branch versions within X hours, say 8 or 12. That way, any changes that happen within that window get collated into a single refresh, such that we’d build at most 2 or 3 times per day while also not falling too far behind with the “last updated” display on the website. It shouldn’t really matter if the latest change (for just the default branch) is 8h or 12h behind.

We'd pick up a few more builds than we do with current throttling (unless we decide to make the window 24h and accept the update metadata on the package page to be more outdated).

The upside is that the throttling query becomes simpler again, albeit at the expense of more complex analysis version handling. OTOH, that's outside of an SQL query and probably easier to test and understand.

finestructure:
Fixes #916

### DIFF
--- a/Sources/App/Commands/Analyze.swift
+++ b/Sources/App/Commands/Analyze.swift
@@ -435,8 +435,8 @@ func throttle(lastestExistingVersion: Version?, incoming: [Version]) -> [Version
         return incoming
     }
 
-    // TODO: distance to latestIncoming to to Current.date() ?
-    // TODO: consider version gaps and what would/should happen if latestExisting was later than latestIncoming, i.e. delta < 0
+    #warning("distance to latestIncoming to to Current.date() ?")
+    #warning("consider version gaps and what would/should happen if latestExisting was later than latestIncoming, i.e. delta < 0")
     let delta = latestIncoming.timeIntervalSinceReferenceDate - latestExisting.timeIntervalSinceReferenceDate
 
     let resultingBranchVersion = delta < Constants.branchVersionRefreshDelay
--- a/Sources/App/Commands/Analyze.swift
+++ b/Sources/App/Commands/Analyze.swift
@@ -407,15 +407,6 @@ func throttleBranchVersions(packageDeltas: [Result<(Package, VersionDelta), Erro
 
 func throttleBranchVersions(_ deltas: VersionDelta,
                             delay: TimeInterval) -> VersionDelta {
-    let rejectRecentBranchVersions: (Version) -> Bool = { version in
-        // if any value is nil leave unchanged (don't reject)
-        if case .some(.branch) = version.reference,
-           let commitDate = version.commitDate,
-           commitDate >= Current.date().addingTimeInterval(-delay) {
-            return false
-        }
-        return true
-    }
     let selectRecentBranchVersions: (Version) -> Bool = { version in
         // if any value is nil leave unchanged (don't select)
         if case .some(.branch) = version.reference,
@@ -424,6 +415,9 @@ func throttleBranchVersions(_ deltas: VersionDelta,
             return true
         }
         return false
+    }
+    let rejectRecentBranchVersions: (Version) -> Bool = {
+        !selectRecentBranchVersions($0)
     }
 
     return .init(
--- a/Sources/App/Commands/Analyze.swift
+++ b/Sources/App/Commands/Analyze.swift
@@ -395,6 +395,52 @@ func diffVersions(client: Client,
 }
 
 
+#warning("change to throttle(lastestExistingVersion: Version?, incoming: [Version]) -> [Version]")
+func diffVersions(existing: [Version], incoming: [Version]) -> VersionDelta {
+    let isBranch: (Version) -> Bool = { $0.reference?.isBranch ?? false }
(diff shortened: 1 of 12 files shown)


Sources/App/Commands/Analyze.swift
```diff
@@ -143,6 +143,9 @@ func analyze(client: Client,
                          threadPool: threadPool,
                          transaction: tx,
                          packages: packages)
+                .map { throttleBranchVersions(
+                    packageDeltas: $0,
+                    delay: Constants.branchVersionRefreshDelay) }
```
finestructure, line 148: This is the core of it: after determining the `VersionDelta` (versions to be added, deleted, kept), we go through the three arrays and remove any recently changes `.branch` versions from `toAdd` and move the existing ones from `toDelete` to `toKeep`, essentially undoing any branch changes detected if they happened within the configured time window.

Sources/App/Commands/Analyze.swift
```diff
@@ -354,21 +364,17 @@ func diffVersions(client: Client,
...
         return transaction.eventLoop.future(error: AppError.invalidPackageCachePath(package.id, package.url))
     }
     guard let pkgId = package.id else {
         return transaction.eventLoop.future(error: AppError.genericError(nil, "PANIC: package id nil for package \(package.url)"))
     }
-    
-    let defaultBranch = Repository.defaultBranch(on: transaction, for: package)
-        .map { b -> [Reference] in
-            if let b = b { return [.branch(b)] } else { return [] }  // drop nil default branch
-        }
```
finestructure, line 370: This was an unnecessary query we ran against the `repositories` table. We were already joining against it. I dropped the whole `Repository.defaultBranch` function.

Sources/App/Commands/Analyze.swift
```diff
@@ -399,16 +406,51 @@ func diffVersions(client: Client,
...
+    let rejectRecentBranchVersions: (Version) -> Bool = {
+        !selectRecentBranchVersions($0)
+    }
+
+    return .init(
+        toAdd: deltas.toAdd.filter(rejectRecentBranchVersions),
+        toDelete: deltas.toDelete.filter(rejectRecentBranchVersions),
+        toKeep: deltas.toKeep + deltas.toDelete.filter(selectRecentBranchVersions)
+    )
+}
```
finestructure, line 439: This implements the "throttling"

Sources/App/Models/Version.swift
```diff
@@ -113,7 +113,13 @@ final class Version: Model, Content {
 
 extension Version: Equatable {
     static func == (lhs: Version, rhs: Version) -> Bool {
-        lhs.id == rhs.id
+        if let id1 = lhs.id, let id2 = rhs.id {
+            return id1 == id2
+        } else {
+            return lhs.commit == rhs.commit
+                && lhs.commitDate == rhs.commitDate
+                && lhs.reference == rhs.reference
```
finestructure, line 121: This was a nasty issue for our test assertions: there were several cases where we simply created versions like so without saving:

```
let v = Version(package: p)
```

i.e. without `save()` and without assigning `id:` which meant we were comparing different versions with `nil` ids...

finestructure: (I don't think it was any existing cases but new ones I just wrote for this PR)

Sources/App/Commands/Analyze.swift
```diff
@@ -371,23 +406,44 @@ func diffVersions(client: Client,
...
+    }
+
+    guard let incomingVersion = incoming.latestBranchVersion,
+          let latestIncoming = incomingVersion.commitDate else {
+        // there's no incoming branch version -> leave incoming alone (which will lead to removal)
+        return incoming
+    }
+
+    #warning("distance to latestIncoming to to Current.date() ?")
+    let delta = latestIncoming.timeIntervalSinceReferenceDate - latestExisting.timeIntervalSinceReferenceDate
```
finestructure, line 438: In practise this should make much of a difference as we're discovering new versions every couple of hours at most. I.e. there won't really be a meaningful difference between when a new version appears and `Current.date()`

I think we should go with `Current.date()`.

finestructure: As discussed, change to `Current.date()`

daveverwer (approved):

Sources/App/Commands/Analyze.swift
finestructure: Fixed

PR #921, opened 2021-02-08
Added an "Information for package authors and maintainers" page

daveverwer:
OK, so this is basically done, with just one thing I'm not happy with.

First up, the link to the page:

Link

The thing I'm unhappy with here is the link text. I toyed with "Information for package maintainers" here, which fits fine but feels very generic and people might think it's information for *any* package maintainer, not the maintainers of *this* package. It's actually "Information for <package name> maintainers", but with some of the lengthy package names in the index, that doesn't work. Best I could come up with is what's pictured here, but I'd love suggestions.

Next, the desktop view of the page itself.

Desktop Maintainers

Finally mobile.

Mobile Maintainers

All comments welcome.


Sources/App/Views/PackageController/MaintainerInfo/MaintainerInfoIndex+View.swift
```diff
@@ -0,0 +1,66 @@
...
+                .p("Copy the Markdown above into your package's README file to show always-up-to-date compatibility status for your package."),
+                .h3("Build Compatibility"),
+                .p(
+                    .text("For information on improving your "),
+                    .a(
+                        .href(SiteURL.package(.value(model.repositoryOwner), .value(model.repositoryName), .builds).relativeURL()),
+                        "package's build results"
+                    ),
+                    .text(", including why you might want to add a "),
+                    .code("_spi.yml"),
```
finestructure, line 55: It's actually `.spi.yml` :)

daveverwer: Good catch. Fixing.

finestructure (approved):

finestructure:
I think this looks great. One thing I wondered seeing the list of links: should we call the link just "Information for Maintainers"? It's a bit shorter and the lengthy link make that list hard to parse. That's true of the "Full Package Compatibility Report" as well. Maybe that's something to think about another day. (How about just "Build Report"?)

Re: the long title for packages with long maintainer names: maybe we could pick up the name from the link and do:

```
Alamofire –
Maintainer Information
```

(with an optional or deliberate line break after `–`

daveverwer:
> Re: the long title for packages with long maintainer names: maybe we could pick up the name from the link and do:
> 
> ```
> Alamofire –
> Maintainer Information
> ```
> 
> (with an optional or deliberate line break after `–`

Unfortunately, there are [some packages](https://swiftpackageindex.com/PerfectlySoft/Perfect-GoogleAnalytics-MeasurementProtocol) in the index that mean we can't use the package name in that menu at all. 😅 If it were just that one, we could just live with it, but there are quite a few with titles so long they'd break the menu.

> I think this looks great. One thing I wondered seeing the list of links: should we call the link just "Information for Maintainers"? It's a bit shorter and the lengthy link make that list hard to parse. That's true of the "Full Package Compatibility Report" as well. Maybe that's something to think about another day. (How about just "Build Report"?)

I think "Information for maintainers" suffers a little from the same problem my original title. Maintainers of what? This package? Any package? Also, even "Information for maintainers" also wraps:

Screen Shot 2021-02-15 at 08 37 03@2x

finestructure:
We could just drop the name from the title if it's too long? It's not like you ended up there entirely by accident. However, this just made me realise we should perhaps include a link back to the package. (You said menu but I think you meant title, right?)

Re: the menu, how about "Maintainer information"? Yes, it's not super clear but a) I don't think anything will be really and b) it's not harmful if you click through and find it's not what you were looking for.

daveverwer:
> We could just drop the name from the title if it's too long? It's not like you ended up there entirely by accident. However, this just made me realise we should perhaps include a link back to the package. (You said menu but I think you meant title, right?)

I added the link back to the page inline in the first sentence. Do you think it needs to be more prominent?

Screen Shot 2021-02-15 at 09 01 21@2x

finestructure:
> I added the link back to the page inline in the first sentence. Do you think it needs to be more prominent?

Oh, indeed. I tend not to look for "navigational" links in text, so I'm a bit blind to this. A link "Back to package" in the top right corner like the "View all builds" backlink from the build details would be clearer but it'll also add clutter. OTOH we're also linking back to the package from the builds page from in inline link.

Probably best to leave it as is for now and consider if we want to do it one way or the other in general.

daveverwer:
Deploying as 2.20.1

PR #938, opened 2021-02-16
Package update

finestructure:


daveverwer (approved):

finestructure:
CleanShot 2021-02-16 at 10 16 11@2x

daveverwer:
What? How is this failing now when it passed as this was integrated?

daveverwer:
Ohh, did a vapor update kill something in it?

finestructure:
No, I think it's aliasing, due to OS version or even architecture. I think this is the first time I've run the test after merge:

CleanShot 2021-02-16 at 10 27 23@2x

finestructure:
Yep, fails on `main` for me as well. I'm on macOS 11.2.1 (20D74), Xcode 12.4 (12D4e). Did you record with the same versions? It'd be curious if those don't line up. Normally that only happens when OS or Xcode versions change.

finestructure:
> What? How is this failing now when it passed as this was integrated?

The snapshots aren't being run on CI, because WKWebView is not available on Linux.

finestructure:
I'll go ahead and merge this as it's not a regression (but we'll want to figure out why we're getting different snapshot test results separately)

daveverwer:
> > What? How is this failing now when it passed as this was integrated?
> 
> The snapshots aren't being run on CI, because WKWebView is not available on Linux.
> 
> Edit: the image snapshots, that is

I knew that, but sometimes I don't think 😂

daveverwer:
> Yep, fails on `main` for me as well. I'm on macOS 11.2.1 (20D74), Xcode 12.4 (12D4e). Did you record with the same versions? It'd be curious if those don't line up. Normally that only happens when OS or Xcode versions change.

All just passed for me with:

* Safari 14.0.3 (16610.4.3.1.4)
* Xcode 12.4 (12D4e)
* macOS 11.2.1 (20D74)

daveverwer:
Strange that our versions of macOS are different. Is that an M1 specific version?

finestructure:
Different? Isn't it exactly the same?

daveverwer:
> Different? Isn't it exactly the same?

You snuck in before I was able to delete my comment. I was mixing up the Xcode build number with the macOS build number at first.

PR #940, opened 2021-02-16
Replace CodeKit with Yarn and Parcel.

daveverwer:
Deployment still to be tested, but everything else works locally.

There are new make commands. To build the front-end files:

`make build-front-end`

To watch/update the front-end files while working on them:

`make serve-front-end`


daveverwer:
nbd 😂

Screen Shot 2021-02-16 at 14 02 05@2x

.gitlab-ci.yml
```diff
@@ -32,8 +32,10 @@ build:
   script: |
     VERSION=${CI_COMMIT_TAG:-$CI_COMMIT_SHA}
     echo 'let appVersion = "'${VERSION}'"' > ./Sources/App/Core/AppVersion.swift
+
+    docker run --rm -it -v $PWD:/host -w /host --entrypoint yarn node:15.8-alpine build
```
finestructure, line 36: need to drop `-it` since we're not running it from a terminal:

```
$ VERSION=${CI_COMMIT_TAG:-$CI_COMMIT_SHA} # collapsed multi-line command
the input device is not a TTY
```

https://gitlab.com/finestructure/swiftpackageindex/-/jobs/1033648182

Makefile
```diff
@@ -78,3 +78,9 @@ db-down-test:
 	docker rm -f spi_test
 
 db-reset: db-down db-up migrate
+
+build-front-end:
+	docker run --rm -it -v $$PWD:/host -w /host --entrypoint yarn node:latest build
+
+serve-front-end:
+	docker run --rm -it -v $$PWD:/host -w /host --entrypoint yarn node:latest serve
```
finestructure, line 86: I think we should use the same pinned `node:15.8-alpine` version here so we see when/if we need to update.

daveverwer: Fixed.

.gitlab-ci.yml
daveverwer: I removed it but didn't replace it with anything. Is that what's needed?

finestructure: Yep, that should do! I’ll verify if the build passed. I don’t think we should merge without confirming the deployment images are building correctly.

finestructure: Right, so this command isn’t quite standalone, is it? We need `parcel`?

```
Digest: sha256:db3f9c88223ba2491e7d2846a18efe46ecb4b60bf2bcf4b53cdcd39e5f6888b5
Status: Downloaded newer image for node:15.8-alpine
yarn run v1.22.5
$ parcel build Resources/Scripts/main.js --experimental-scope-hoisting -d Public
/bin/sh: parcel: not found
error Command failed with exit code 127.
```

https://gitlab.com/finestructure/swiftpackageindex/-/jobs/1034263910

daveverwer: It should be standalone. `parcel` is a dependency in the `package.json` file which yarn should install.

daveverwer: Oh, wait. I think I know what's needed.

finestructure (approved):

PR #941, opened 2021-02-16
Let the Prettier formatter do its thing

daveverwer:
I'm now using Prettier as a JS formatted and it wanted to adjust my formatting.

Doing this as a separate PR as it's only formatting changes which can cloud other PRs with noise. There are no functionality changes here.

Merge after #940


finestructure:
This is now good to go, right?

finestructure (approved):

finestructure:
I've clicked "Update branch" here, which maybe I shouldn't have give there's the other branch also based on these changes (a rebase would be clearer, I think). I can fix that up but I don't want to force push your branches.

Just let me know or update and merge yourself, I'm marked them approved!

PR #944, opened 2021-02-17
Yarn package upgrade

daveverwer:
Should fix the dependabot security alert. We're not using that bit, but still better to do it!

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -4,26 +4,26 @@
 
 "@babel/code-frame@7.12.11":
   version "7.12.11"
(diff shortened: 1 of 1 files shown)


finestructure (approved):

finestructure:
My word, a 5.7k line dependency lock file 🤪

daveverwer:
Yea 😬

Also, it didn't fix the alert because `parcel` is locked on an old version of the vulnerable package. There's a new major version of parcel, which will hopefully move that dependency forward too. We're still not using the vulnerable dependency in any meaningful way though, so it's all good.

PR #945, opened 2021-02-18
Basic search page

finestructure:
Just the page for now, pagination is requiring bigger changes/discussion.

### DIFF
--- a/Sources/App/Core/SiteURL+siteMap.swift
+++ b/Sources/App/Core/SiteURL+siteMap.swift
@@ -54,6 +54,8 @@ extension SiteURL {
                 return .hourly
             case .rssReleases:
                 return .hourly
+            case .search:
+                return .hourly
             case .siteMap:
                 return .weekly
             case .stylesheets:
--- a/Sources/App/Views/Search/Search+View.swift
+++ b/Sources/App/Views/Search/Search+View.swift
@@ -1,0 +1,37 @@
+import Plot
+
+
+extension Search {
+
+    class View: PublicPage {
+
+        let model: Model
+
+        init(path: String, model: Model) {
+            self.model = model
+            super.init(path: path)
+        }
+
+        override func content() -> Node<HTML.BodyContext> {
+            .group(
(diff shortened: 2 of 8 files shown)


Sources/App/Views/Search/Search+View.swift
```diff
@@ -0,0 +1,37 @@
...
+        init(path: String, model: Model) {
+            self.model = model
+            super.init(path: path)
+        }
+
+        override func content() -> Node<HTML.BodyContext> {
+            .group(
+                .h2(.text("Results for \(model.query)")),
+                .ul(
+                    .class("list"),
```
finestructure, line 19: Don't think this class actually exists but `package_list` was rending the two column layout. Just a placeholder.

Sources/App/Core/SiteURL+siteMap.swift
```diff
@@ -54,6 +54,8 @@ extension SiteURL {
                 return .hourly
             case .rssReleases:
                 return .hourly
+            case .search:
+                return .hourly
```
daveverwer, line 58: It doesn't matter very much, but I'm not sure `hourly` is correct here. I was going to suggest `never` as the search page on its own is not something we'd want Google sending people to the search page directly, but the [docs](https://www.sitemaps.org/protocol.html) for `never` say:

> The value "never" should be used to describe archived URLs.

Which isn't true.

It's probably fine as `hourly`, but I thought I'd bring it up.

daveverwer (approved):
Tested locally. Only the comment about the sitemap really.

Sources/App/Core/SiteURL+siteMap.swift
finestructure: 👍  I don't really know what to pick here, I just filled in what I thought the change rate was going to be.

daveverwer: It's just telling Google how often to come back and re-index that page., which should really be `never` as there will never be anything on that page for Google to index. It's not a huge deal

PR #946, opened 2021-02-18
Search pagination

finestructure:
Merge after #945

### DIFF
--- a/Sources/App/Controllers/SearchController.swift
+++ b/Sources/App/Controllers/SearchController.swift
@@ -8,11 +8,10 @@ struct SearchController {
     func show(req: Request) throws -> EventLoopFuture<HTML> {
         let query = req.query[String.self, at: "query"] ?? ""
         let page = req.query[Int.self, at: "page"] ?? 1
-        let pageSize = req.query[Int.self, at: "page-size"] ?? Constants.searchPageSize
         return API.search(database: req.db,
                           query: query,
                           page: page,
-                          pageSize: pageSize)
+                          pageSize: Constants.searchPageSize)
             .map(\.results)
             .map { SearchShow.Model.init(query: query, results: $0) }
             .map { SearchShow.View.init(path: req.url.path, model: $0).document() }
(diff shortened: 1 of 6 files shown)


Sources/App/Controllers/SearchController.swift
```diff
@@ -7,7 +7,12 @@ struct SearchController {
 
     func show(req: Request) throws -> EventLoopFuture<HTML> {
         let query = req.query[String.self, at: "query"] ?? ""
-        return API.search(database: req.db, query: query)
+        let page = req.query[Int.self, at: "page"] ?? 1
+        let pageSize = req.query[Int.self, at: "page-size"] ?? Constants.searchPageSize
```
finestructure, line 11: Not sure about the parameter name `page-size` here, `pageSize`, `page_size`?

Sources/App/Core/Constants.swift
```diff
@@ -31,8 +31,7 @@ enum Constants {
     static let rssFeedMaxItemCount = 100
     static let rssTTL = 60  // minutes
     
-    static let searchLimit = 20
-    static let searchLimitLeeway = 5
```
finestructure, line 35: I've had to drop the "leeway" mechanism as that was too hard to reconcile with paging.

daveverwer:
This is great 😍

Sources/App/Core/Constants.swift
daveverwer: Yea, I'd say this is no longer needed.

Sources/App/Controllers/SearchController.swift
daveverwer: I'd say maybe this doesn't need to be a query string parameter. The page number is certainly necessary, but page size can be fixed in the source. I think once we settle on a value, it won't need to change.

daveverwer (approved):
Is there one more PR after this to get the pagination controls into the basic search view?

finestructure:
Yes, that'll follow

PR #948, opened 2021-02-22
Basic pagination controls

### LINKED ISSUE #635: Add an intermediate search results page
There should be an intermediate search results page that lives on the URL:

`/search?query=[query]&page=[page]`

The rough layout for the page can be seen in #636. 

The "paging" control at the bottom should have previous/next links (if relevant for the page number) and then a list of pages. It's almost certainly not necessary to do the "..." thing as pictured because I can't think that there would ever be more than a few pages of results. So, the numbers at the bottom are simply a list of pages that cover all possible search results. Those link to the search page again with the `page` parameter specified above.

Each page should have 20 results listed (but let's make this easily tweakable as we implement this as we'll want to play with what feels right).

Each search result should include the following:

* Package Name
* Package Description
* Package Identifier (For example `daveverwer/LeftPad`)

finestructure:
Merge after #947 

Fixes #635 

Please run the new `test_SearchShow` on your machine and overwrite with your result if there are differences to avoid us having a mix of things that pass for you but not me. (For me currently all image snapshots except this new one which I recorded are failing due to aliasing differences.)

test_SearchShow desktop


Sources/App/Core/Search.swift
```diff
@@ -105,6 +105,7 @@ enum Search {
             .where(isNotNull(repoName))
             .orderBy(eq(lower(packageName), mergedTerms), .descending)
             .orderBy(score, SQLDirection.descending)
+            .orderBy(packageName, SQLDirection.ascending)
```
finestructure, line 108: The order can change slightly between runs, because ordering by score is not very granular (there are many packages which share a score). Pagination makes that more obvious when we slice into the set and results end up on different pages between requests.

daveverwer: Yep, this is perfect 👍

finestructure:
test_SearchShow desktop

daveverwer (approved):

PR #949, opened 2021-02-22
Log throttling

### LINKED ISSUE #933: Add logging and/or metrics to be able to monitor version throttling
Should have done that from the start and it's something to bear in mind whenever we're doing this sort of thing in the future.

finestructure:
Fixes #933


daveverwer (approved):
This will be useful information 👍

PR #950, opened 2021-02-22
Simplified snapshot testing setup

daveverwer:
Merge after #948 

As discussed, simplification of the setup for snapshot test cases. You may not want the second level of abstraction here, it's easy to remove if we decide it's overkill, but the TempWebRoot stuff only needs configuring for web page snapshot tests.

### DIFF
--- a/Tests/AppTests/MarkdownHTMLConverterTests.swift
+++ b/Tests/AppTests/MarkdownHTMLConverterTests.swift
@@ -6,7 +6,7 @@ import XCTest
 import Plot
 
 
-class MarkdownHTMLConverterTests: XCTestCase {
+class MarkdownHTMLConverterTests: WebpageSnapshotTestCase {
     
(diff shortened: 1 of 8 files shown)


Tests/AppTests/MarkdownHTMLConverterTests.swift
```diff
@@ -6,7 +6,7 @@ import XCTest
 import Plot
 
 
-class MarkdownHTMLConverterTests: XCTestCase {
+class MarkdownHTMLConverterTests: WebpageSnapshotTestCase {
```
finestructure, line 9: One thing to bear in mind is that `AppTestCase` (which `WebpageSnapshotTestCase` extends) recreates the schema and is noticeable slower than `XCTestCase` (~150ms per test on my machine). If we start adding lots of snapshot tests that don't need the db we should probably disentangle that.

finestructure (approved):

PR #955, opened 2021-02-24
Clarify throttle logic

finestructure:
Since I already struggled to understand it after a week or so I've added a comment.

### DIFF
--- a/Tests/AppTests/AnalyzerVersionThrottlingTests.swift
+++ b/Tests/AppTests/AnalyzerVersionThrottlingTests.swift
@@ -265,47 +265,48 @@ class AnalyzerVersionThrottlingTests: AppTestCase {
         }
     }
 
-    func test_throttle_pathological_cases() throws {
-        // Test the pathological case where we have a newer branch revision in
-        // the db than the incoming version.
+    func test_throttle_force_push() throws {
+        // Test the exceptional case where we have a newer branch revision in
+        // the db than the incoming version. This could happen for instance
+        // if an older branch revision is force pushed, effectively removing
+        // the "existing" revision, replacing it with an older "incoming"
+        // revision.
         // setup
         Current.date = { .t0 }
         let pkg = Package(url: "1")
         try pkg.save(on: app.db).wait()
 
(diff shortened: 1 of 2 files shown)


Tests/AppTests/AnalyzerVersionThrottlingTests.swift
```diff
@@ -304,6 +305,7 @@ class AnalyzerVersionThrottlingTests: AppTestCase {
 
             // validate
             XCTAssertEqual(res, [new])
+            // XCTAsserEqual(res, [old]) ??
         }
```
finestructure, line 309: This looks weird - note that `new` is actually the older version and we choose it, because if the old (or rather: existing) version is outside the window we indiscriminately choose the incoming version. There is no check if it's actually newer.

I _think_ this is still the correct thing to do, because it reflects the scenario where a branch is force pushed to an older revision.

daveverwer: This is a complicated situation. The force pushing would need to be within the 24-hour window for that scenario you mentioned to affect this.

But, this test as it's written, I'd expect the assert to be `[old]` and not `[new]`, so it's potentially concerning that this passes as-is.

finestructure: It's actually the other way around - it's outside the 24h window and if we didn't do what we currently do we'd fail to purge the outdated version forever.

In other words, I think we _must_ do what we're currently doing or a force push never get the intended version through.

Should add that as a comment to the test to clarify.

finestructure: I've renamed the variables and the test to make this clearer.

daveverwer: As we discussed, I tripped up on the meaning of the `ex` variable. My mind kept going to `excluded` rather than `existing`. I'd personally rename them to `existing` and `incoming` for extra clarity but I'll leave it to you.

daveverwer (approved):

PR #957, opened 2021-02-26
New search design

### LINKED ISSUE #477: UX - hard to find all results when searching
I wanted to check several logging libraries so I used the index website. 

It took me a significant amount of time to realize that the list of the results when searching for some term is scrollable. At first I was like: "how the hell can I display all results, not just 4?"

I would recommend to add there some visual key - either button "show all results" (with number) or flash scroll indicators or something like that ;)

### LINKED ISSUE #637: Integrate the home page with the intermediate search page
Depends on #635.

Functionality is described in #636.

daveverwer:
Fixes #637
Fixes #477 

### Advantages

* Search from anywhere. The home page's focus is still searching, but people can also search from the results page after a query executes, and even the package pages, FAQ pages, etc.
* The search results are much less cramped. I already prefer it over the small dropdown, but as we get more packages and more information to display per search result, this problem worsens.

### For discussion:

* We could remove the search buttons from being visible. They need to be there for accessibility but do not need to show. The search field in the navigation bar does this.
	* Or, the search button could be an icon or different colours rather than being removed. I could style it nicely as an icon to look like a macOS search field.

### Ideas for making search progressively enhance

* We could add keyboard navigation on the search results page. Pressing up/down should start a selection in the list, and enter could navigate to the selected item.
	* The search field also could be autofocused on the search results page. We'd need to make sure this worked well with the previous keyboard navigation with up/down taking focus out of the text field and into the selection. A little like it does now on the home page right now.
* The search results page could be search-as-you-type. Unfortunately, this is not quite as simple as it first seems. The API only returns one page of results, where the search results page returns paginated results. We could fake it as we have a `hasMoreResults` field in the API, which we could use to add a "Next page" link.

### If we don't like this method at all

* What is currently the search results page could be turned into the home page. Without a query, it could show the current home page content. Then, as you start typing, we could replace that with search results. I don't think this is necessarily the best idea, but it would be the compromise I'd propose if we decide that the home page search must be as-you-type.

### Not yet completed

* I'd like to turn the results page into a two-column layout with a right sidebar that could show things such as:
	* A list of categories from the packages that would allow exploring from a search into related categories. I kind of love this idea.
	* The recent packages/recent releases lists.
	* Sponsorship CTA 💰
* Mobile layout for the search field. I know what to do here, but I have not yet done it.
* When a search returns no results, the opening paragraph should be better.
* Pagination is not styled.

## Screenshots

Screen Shot 2021-02-26 at 19 31 58@2x

Screen Shot 2021-02-26 at 19 32 21@2x

Screen Shot 2021-02-26 at 19 32 29@2x


finestructure:
Mmm, I can't run `make build-front-end` on the M1 - there seems to be an incompatibility with the arm image.

```
yarn install v1.22.5
[1/4] Resolving packages...
[2/4] Fetching packages...
info fsevents@1.2.13: The platform "linux" is incompatible with this module.
info "fsevents@1.2.13" is an optional dependency and failed compatibility check. Excluding it from installation.
info fsevents@2.3.2: The platform "linux" is incompatible with this module.
info "fsevents@2.3.2" is an optional dependency and failed compatibility check. Excluding it from installation.
[3/4] Linking dependencies...
[4/4] Building fresh packages...
[-/3] ⠄ waiting...
[2/3] ⠄ deasync
error /host/node_modules/deasync: Command failed.
Exit code: 1
```

Just recalled that last time I extracted the compiled resources from the image. Will try that.

daveverwer:
I want to give it another look over in the morning, but I think this is ready to go.

No major changes after our review of it this afternoon, but you might want to look through [this commit](https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/pull/957/commits/387bfd831f4f0e99356cd9f876ea7d3ee0c85dad) as you may have thoughts on some things in there.

daveverwer:
This is ready to go 👍

finestructure (approved):

PR #962, opened 2021-03-03
Fixed deprecation warning caused by the latest version of Highlight.js

daveverwer:
Merge after #957

Screen Shot 2021-03-03 at 11 24 29@2x


daveverwer:
This causes the yarn build process to break. Not ready for merging!

PR #967, opened 2021-03-04
Make the resourceReloadString be based on local file modification date rather than the current time.

daveverwer:
It was always a bit of a hack to use the current date/time to force the local dev environment to reload fresh JS/CSS resources. This is how it should have been done originally.

This changed was prompted by the addition of Turbolinks. If we reload the CSS/JS with *every* page load, like we were, then Turbolinks never gets to do its job and it's effectively impossible to test Turbolinks locally as it's always doing full resource re-loads.

### DIFF
--- a/Sources/App/Core/AppVersion.swift
+++ b/Sources/App/Core/AppVersion.swift
@@ -1,4 +1,8 @@
 // In a local development environment, appVersion will remain nil (as set here).
 // In the staging development environment, appVersion is set to the commit hash of the deployed version.
 // In production, appVersion is set either to released tag name, or to the commit hash if that does not exist.
+
+// Note: If the definition of appVersion ever changes, the `gitlab-ci.yml` file also
+// needs updating as this file is re-generated during the deployment process.
+
 let appVersion: String? = nil
--- a/Sources/App/Core/AppVersion.swift
+++ b/Sources/App/Core/AppVersion.swift
@@ -1,1 +1,4 @@
-let appVersion = "dev - will be overriden in release builds"
+// In a local development environment, appVersion will remain nil (as set here).
+// In the staging development environment, appVersion is set to the commit hash of the deployed version.
+// In production, appVersion is set either to released tag name, or to the commit hash if that does not exist.
+let appVersion: String? = nil
--- a/Sources/App/Views/PublicPage.swift
+++ b/Sources/App/Views/PublicPage.swift
@@ -127,6 +127,9 @@ class PublicPage {
         if let appVersion = appVersion {
             return appVersion
         } else {
+            // Running under test? It's annoying to need to update snapshots every time the CSS or JS is saved.
+            if let _ = NSClassFromString("XCTest") { return "test" }
+
             // Return the date of the most recently modified between the JavaScript and CSS resources.
             let jsModificationDate = modificationDate(forLocalResource: "main.js")
             let cssModificationDate = modificationDate(forLocalResource: "main.css")
--- a/Sources/App/Views/PublicPage.swift
+++ b/Sources/App/Views/PublicPage.swift
@@ -101,15 +101,37 @@ class PublicPage {
                     """))
     }
 
-    /// A query string that will force resources to reload after they change. In development this is the
-    /// current date, to force a reload every time. In production this is the date of the last deploy.
+    /// A query string that will force resources to reload CSS/JS resources change.
     /// - Returns: A string containing the query string.
     final func resourceReloadQueryString() -> String {
-        let secondsSince1970String = String(Int(Current.date().timeIntervalSince1970))
-        do {
-            return (try Environment.detect() == .development) ? secondsSince1970String : appVersion
-        } catch {
-            return secondsSince1970String
+
+        // This method is only called in a local development environment, so all paths
+        // can be relative to this source file.
+        func modificationDate(for resource: String) -> Date {
+            let relativePathToPublic = "../../../Public/"
+            let url = URL(fileURLWithPath: relativePathToPublic + resource,
+                relativeTo: URL(fileURLWithPath: #file))
+
+            // Assume the file has been modified *now* if the file can't be found.
+            guard let attributes = try? Foundation.FileManager.default.attributesOfItem(atPath: url.absoluteString)
+            else { return Date() }
+
+            // Also assume the file is modified now if the attribute doesn't exist.
+            let modificationDate = attributes[FileAttributeKey.modificationDate] as? Date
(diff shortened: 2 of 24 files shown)


Sources/App/Views/PublicPage.swift
```diff
@@ -101,15 +101,40 @@ class PublicPage {
...
+            // Also assume the file is modified now if the attribute doesn't exist.
+            let modificationDate = attributes[FileAttributeKey.modificationDate] as? Date
+            return modificationDate ?? Date()
+        }
+
+
+        // In staging or production appVersion will be set to a commit hash, or a tag name.
+        // It will only ever be nil when running in a local development environment.
+        if let appVersion = appVersion {
+            return appVersion
```
daveverwer, line 128: In staging and production environments, this is the exact same code as was previously being executed so there should be no change to staging/production caused by this.

daveverwer:
Hmm, it looks like tests are failing because this line maybe isn't working on Linux? Tests run fine locally.

```
if let _ = NSClassFromString("XCTest") { return "test" }
```

I just did a bit of research and [found this](https://bugs.swift.org/browse/SR-3333) which seems to confirm that `NSClassFromString` does not work on Linux.

Sources/App/Core/AppVersion.swift
```diff
@@ -1 +1,4 @@
-let appVersion = "dev - will be overriden in release builds"
+// In a local development environment, appVersion will remain nil (as set here).
+// In the staging development environment, appVersion is set to the commit hash of the deployed version.
+// In production, appVersion is set either to released tag name, or to the commit hash if that does not exist.
+let appVersion: String? = nil
```
finestructure, line 4: I suspect this will fail the builds on Gitlab, where we create the images, because the command is:

```
  script: |
    VERSION=${CI_COMMIT_TAG:-$CI_COMMIT_SHA}
    echo 'let appVersion = "'${VERSION}'"' > ./Sources/App/Core/AppVersion.swift
```

This would need to be changed to

```
  script: |
    VERSION=${CI_COMMIT_TAG:-$CI_COMMIT_SHA}
    echo 'let appVersion: String? = "'${VERSION}'"' > ./Sources/App/Core/AppVersion.swift
```

daveverwer:
> I just did a bit of research and [found this](https://bugs.swift.org/browse/SR-3333) which seems to confirm that `NSClassFromString` does not work on Linux.

I'm not coming up with much that allows detection of a test environment on Linux. Any ideas @finestructure?

Sources/App/Core/AppVersion.swift
daveverwer: Fixed

finestructure:
I would set an env variable that we inspect, like we do for "running in CI":

```swift
    var isRunningInCI: Bool {
        ProcessInfo.processInfo.environment.keys.contains("GITHUB_WORKFLOW")
    }
```

finestructure:
Or maybe just abuse one of the existing ones as a "hint", for instance `XCTestSessionIdentifier`:

```
(lldb) po ProcessInfo.processInfo.environment.keys
▿ Dictionary.Keys(["NSUnbufferedIO", "XCTestSessionIdentifier", "METRICS_PUSHGATEWAY_URL", "DEV_TWITTER_API_SECRET", "__XPC_DYLD_LIBRARY_PATH", "DEV_TWITTER_ACCESS_TOKEN_SECRET", "DEV_TWITTER_ACCESS_TOKEN_KEY", "DATABASE_HOST", "SWIFTUI_VIEW_DEBUG", "CA_ASSERT_MAIN_THREAD_TRANSACTIONS", "PWD", "LD_LIBRARY_PATH", "__CF_USER_TEXT_ENCODING", "MallocNanoZone", "XCTestBundlePath", "DATABASE_NAME", "GITLAB_PIPELINE_TOKEN", "LOGNAME", "BUILDER_TOKEN", "DEV_TWITTER_API_KEY", "XPC_FLAGS", "DYLD_LIBRARY_PATH", "SQLITE_ENABLE_THREAD_ASSERTIONS", "DATABASE_PORT", "USER", "SSH_AUTH_SOCK", "DYLD_FALLBACK_FRAMEWORK_PATH", "DYLD_FALLBACK_LIBRARY_PATH", "DYLD_FRAMEWORK_PATH", "DATABASE_PASSWORD", "PATH", "HOME", "CA_DEBUG_TRANSACTIONS", "SKIP_SNAPSHOTS", "SHELL", "OS_ACTIVITY_DT_MODE", "XPC_SERVICE_NAME", "LaunchInstanceID", "__XPC_DYLD_FRAMEWORK_PATH", "__CFBundleIdentifier", "XCTestConfigurationFilePath", "DATABASE_USERNAME", "COMMAND_MODE", "TMPDIR", "GITHUB_TOKEN", "SECURITYSESSIONID", "LOG_LEVEL", "__XCODE_BUILT_PRODUCTS_DIR_PATHS"])
  ▿ _variant : _Variant
    ▿ object : _BridgeStorage<__RawDictionaryStorage>
      - rawValue : (Opaque Value)
```

daveverwer:
> Or maybe just abuse one of the existing ones as a "hint", for instance `XCTestSessionIdentifier`:

That's a great idea. 👍

Sources/App/Views/PublicPage.swift
```diff
@@ -101,15 +101,40 @@ class PublicPage {
...
+        } else {
+            // Running under test? It's annoying to need to update snapshots every time the CSS or JS is saved.
+            if let _ = NSClassFromString("XCTest") { return "test" }
```
finestructure, line 131: Is there no way to keep this inside the tests? It doesn't feel like it's the right place to test for test execution inside the app module. If possible I think we should constrain those kind of checks to the test module.

daveverwer: I agree, and I considered this when I wrote it, but I am not sure how we would best achieve it. It's such a specific thing that pulling it all out into a separate struct/class that can then be mocked seems like overkill.

finestructure: Let me take a quick look, I think this can be done in a pretty simple way.

finestructure (approved):

Sources/App/Views/PublicPage.swift
daveverwer: That's a great solution 👍 Thanks

daveverwer:
> > Or maybe just abuse one of the existing ones as a "hint", for instance `XCTestSessionIdentifier`:
> 
> That's a great idea. 👍

For future reference, this environment variable does not seem to be set when running under CI.

PR #989, opened 2021-03-17
Fix rate limit reporting

finestructure:
This is important to roll out before I activate any additional ingestion jobs. The problem was that I only added rate limit reporting when _hit_ the limit. Which is why there's no data available 😅


daveverwer (approved):
Conditional logic is hard! 😂

PR #990, opened 2021-03-17
Updated to the latest Package Collection models

daveverwer:
Implemented manifests and the new fields. We always just push one manifest, as we don't support multiple, but we're compatible with the schema again 👍

### DIFF
--- a/Tests/AppTests/PackageCollectionTests.swift
+++ b/Tests/AppTests/PackageCollectionTests.swift
@@ -21,7 +21,9 @@ class PackageCollectionTests: AppTestCase {
         do {
             let v = try Version(package: p,
                                 packageName: "Foo",
+                                publishedAt: Date(timeIntervalSince1970: 0),
                                 reference: .tag(1, 2, 3),
+                                releaseNotes: "Bar",
                                 supportedPlatforms: [.ios("14.0")],
                                 toolsVersion: "5.3")
             try v.save(on: app.db).wait()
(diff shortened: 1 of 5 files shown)


Tests/AppTests/PackageCollectionTests.swift
```diff
@@ -21,7 +21,9 @@ class PackageCollectionTests: AppTestCase {
         do {
             let v = try Version(package: p,
                                 packageName: "Foo",
+                                publishedAt: Date(timeIntervalSince1970: 0),
```
finestructure, line 24: FYI: We've also got the `.t0`, `.t1` shortcuts, which are believe are available in all tests.

Tests/AppTests/PackageCollectionTests.swift
```diff
@@ -70,26 +72,32 @@ class PackageCollectionTests: AppTestCase {
...
+        XCTAssertEqual(res.verifiedCompatibility, [
+            .init(platform: .init(name: "ios"), swiftVersion: .init("5.2")),
+            .init(platform: .init(name: "macos"), swiftVersion: .init("5.3")),
+        ])
+        XCTAssertEqual(res.license, .init(name: "MIT", url: URL(string: "https://foo/mit")!))
+        XCTAssertEqual(res.createdAt, Date(timeIntervalSince1970: 0))
+
+        // The spec requires there to be a dictionary keyed by the default tools version.
+        guard let manifest = res.manifests[res.defaultToolsVersion]
+        else { XCTFail(); return; }
```
finestructure, line 87: Does

```
let manifest = try XCTUnwrap(...)
```

not work here? I'll take a quick look before merging.

finestructure (approved):

PR #991, opened 2021-03-17
Fix snapshots

daveverwer:
Merge after #990

I don't understand how these got out of sync, and I can not see *any* difference using Kaleidoscope's difference view.

Very strange. Can you try running these locally on your M1 please @finestructure?


finestructure:
All 17 failing for me 🤷

finestructure (approved):

PR #992, opened 2021-03-18
Move staging

finestructure:
I have manually deployed with this new pipeline to the new environment.

Before we can merge this, we need to point `staging.swiftpackageindex.com` at `40.87.31.205`. That's all it should take to transition 🤞 

CleanShot 2021-03-19 at 17 47 16@2x


finestructure:
Leaving this in draft so we can coordinate the DNS change and merge on Monday morning.

daveverwer:
🎉 Exciting times!

daveverwer (approved):

PR #999, opened 2021-03-24
Ssl db connection

finestructure:
Add an env variable to force TLS on db connections. The env variable is configured on `dev`, will manually deploy to test before undrafting.


finestructure:
It's up and running via the hosted db now.

Sources/App/configure.swift
```diff
@@ -22,12 +22,16 @@ public func configure(_ app: Application) throws {
...
+    let tlsConfig: TLSConfiguration? = Environment.get("DATABASE_USE_TLS")
+        .flatMap(\.asBool)
+        .flatMap { $0 ? .clientDefault : nil }
     app.databases.use(.postgres(hostname: host,
                                 port: port,
                                 username: username,
                                 password: password,
-                                database: database), as: .psql)
+                                database: database,
+                                tlsConfiguration: tlsConfig), as: .psql)
```
finestructure, line 34: `nil` is the default, so this doesn't change anything for `production` and since the CI tests pass (which don't have `DATABASE_USE_TLS` set either and use a non-TLS db this shows that it doesn't affect production deployments.

daveverwer (approved):

PR #1008, opened 2021-03-29
Deployment tweaks

finestructure:


daveverwer (approved):
Oh, including the job URL is a great idea 👍

PR #1011, opened 2021-03-30
Update the footer to include Microsoft Azure.

daveverwer:


Sources/App/Views/PublicPage.swift
```diff
@@ -322,6 +322,11 @@ class PublicPage {
                         .a(
                             .href("https://macstadium.com/"),
                             "MacStadium"
+                        ),
+                        .text("and"),
+                        .a(
+                            .href("http://azure.microsoft.com"),
```
finestructure, line 328: `http` not `https`? I mean it'll get redirected I'm sure, just looks odd :)

finestructure: I'll approve but disable auto-merge so you can decide if you want to change it or not.

finestructure (approved):

Sources/App/Views/PublicPage.swift
daveverwer: How did I make that mistake? I copied it from the browser address bar 😅

Issue #973, opened 2021-03-05
Decouple the load of a package's readme file from the page load


daveverwer:
There's a noticeable delay when loading the package pages, caused by the server-side fetch and render of the markdown.

Split this into two steps:

* Load the package page, without the README.
* Add a new server-sider method that returns just a chunk of HTML from Markdown parsed README and insert it into the document with an XHR.
* Alternatively, investigate swapping out the Markdown renderer for grabbing rendered HTML from GitHub.

PR #1018, opened 2021-04-07
Fetch the Readme HTML URL in addition to the Markdown URL

daveverwer:
Necessary for #973.

⚠️ Contains a schema change ⚠️

### DIFF
--- a/Sources/App/Commands/Ingest.swift
+++ b/Sources/App/Commands/Ingest.swift
@@ -90,6 +90,7 @@ func ingest(client: Client,
             database: Database,
             logger: Logger,
             packages: [Package]) -> EventLoopFuture<Void> {
+    logger.info("Ingesting \(packages.compactMap {$0.id})")
     AppMetrics.ingestCandidatesCount?.set(packages.count)
     let metadata = fetchMetadata(client: client, packages: packages)
     let updates = metadata.flatMap { updateRepositories(on: database, metadata: $0) }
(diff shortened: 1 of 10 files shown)


Sources/App/Commands/Ingest.swift
```diff
@@ -90,6 +90,7 @@ func ingest(client: Client,
             database: Database,
             logger: Logger,
             packages: [Package]) -> EventLoopFuture<Void> {
+    logger.info("Ingesting \(packages.compactMap {$0.id})")
```
daveverwer, line 93: I added this additional logging during development but thought it might actually be useful to leave in permanently. It logs the package IDs of all packages about to be ingested.

Tests/AppTests/PackageShowModelTests.swift
```diff
@@ -248,7 +248,7 @@ class PackageShowModelTests: SnapshotTestCase {
             defaultBranch: "default",
             name: "Alamofire",
             owner: "Alamofire",
-            readmeUrl: "https://github.com/Alamofire/Alamofire/master/README.md"
+            readmeUrl: "https://raw.githubusercontent.com/foo/bar/main/README.md"
```
daveverwer, line 251: As part of this PR, I made the test data we're using for the readmeUrl a little more real-world. It was a little confusing for the test URLs to be to HTML representations of the markdown when in actuality we're getting raw Markdown.

Sources/App/Commands/Ingest.swift
finestructure: Mmm, I'm not sure how valuable that is in production... We're ingesting 100 packages at a time every two minutes and would get a wall of 100 lines with id that aren't super useful in the logs.

How about we make this a `debug` and then you can use a lower log level for testing?

daveverwer: Fixed

finestructure (approved):

PR #1027, opened 2021-04-12
Update url scheme for "Try in a playground"

finestructure:


daveverwer (approved):
All good apart from the snapshot diffs.

finestructure:
No worries, was easy to edit! It's the image tests that fail but the html snapshot is generated from the same test as the image. Could pull those apart tbh if you insist on never buying an Mx machine ;)

daveverwer:
Or just let me know when you'd like me to run snapshots 😂

PR #1028, opened 2021-04-13
Decouple the loading of the README file from the Package page load.

### LINKED ISSUE #973: Decouple the load of a package's readme file from the page load

daveverwer:
Fixes #973

This does *not* yet include the new plan to parse the rendered HTML directly from GitHub. This PR simply (hah, simply!) decouples our existing README loading machinery from the Package page load.

The Package show action no longer makes a network request, which should speed it up considerably. A loading indicator is displayed while the README content is fetched, parsed, and rendered, then it's asynchronously loaded with Turbo.


Tests/AppTests/Util+Snapshotting.swift
```diff
@@ -17,6 +17,14 @@ extension Snapshotting where Value == () -> HTML, Format == String {
...
 }
 
+extension Snapshotting where Value == () -> Node<HTML.BodyContext>, Format == String {
+    public static var html: Snapshotting {
+        Snapshotting<String, String>.lines.pullback { node in
+            Current.siteURL = { "http://localhost:8080" }
+            return node().render(indentedBy: .spaces(2))
+        }
+    }
+}
```
daveverwer, line 27: There must be a way to do this without duplicating this code, but it is beyond my abilities.

Tests/AppTests/Util+Snapshotting.swift
```diff
@@ -61,5 +69,48 @@ extension Snapshotting where Value == () -> HTML, Format == NSImage {
...
+                fatalError("Snapshotting: 💥 Failed to write index.html: \(error)")
+            }
+
+            // Load the HTML file into the web view with access to public directory
+            webView.loadFileURL(htmlURL, allowingReadAccessTo: baseURL)
+
+            return webView
+        }
+    }
+}
```
daveverwer, line 114: Same here.

Sources/App/Views/TurboFrame.swift
```diff
@@ -0,0 +1,25 @@
+import Foundation
+import Vapor
+import Plot
+
+class TurboFrame {
```
finestructure, line 5: Nit: maybe we should pick a more generic name that’s not JS library derived? Not sure if there is a good one, just feels a bit specific. Maybe just `Frame`, `PageSnippet`, `Partial`, `PartialPage` or something.

Tests/AppTests/Util+Snapshotting.swift
finestructure: We should be able to `pullback` to/from the existing strategy. I’ll take a look tomorrow!

Tests/AppTests/Util+Snapshotting.swift
finestructure: Dto

finestructure:
The image snapshots are all missing the READMEs now, which is due to the async JS loading, right? Is there any way to bring that back for at least one of the tests?

Tests/AppTests/Util+Snapshotting.swift
finestructure: The `.html` one I can confirm works but the `.image` one I can't really test. The output it creates shows a diff but we've had that all along, so you'd need to rerun this with my changes to confirm it's as expected.

(The only thing that could really be an issue is the `.body(...)` wrapping to make the contexts align but that shouldn't cause a diff, esp since even the `.html` is unaffected.)

finestructure:
This doesn't seem to load READMEs for me:

CleanShot 2021-04-14 at 06 55 36@2x

This is with a prod snapshot from a moment ago and the readme url fields are set for Alamofire:

CleanShot 2021-04-14 at 06 57 00@2x

daveverwer:
> The image snapshots are all missing the READMEs now, which is due to the async JS loading, right? Is there any way to bring that back for at least one of the tests?

Yes, all missing from package pages now, but I added a specific snapshot test to render the readme page fragment that gets included, which does the same job.

daveverwer:
> This doesn't seem to load READMEs for me:
>
> This is with a prod snapshot from a moment ago and the readme url fields are set for Alamofire:

Are there any errors in your JavaScript console?

Sources/App/Views/TurboFrame.swift
daveverwer: Yea I played with names like `PageSnippet` but with the methods like `frameIdentifier` which really do need the word frame in them I figured we were already quite tightly tied to terms from the library. I guess `PageFrame` would work, but I'm not sure it's much better.

daveverwer: The more I think about this, I think `TurboFrame` is the right name. This *is* very tied to that JavaScript library, so if we ever change that there will need to be a lot of changes here, and renaming the class can be one of them.

finestructure (approved):

PR #1032, opened 2021-04-14
Add Universal Links for "Try in a Playground" and a download page.

### LINKED ISSUE #934: Better flow for people without the Playgrounds app installed
When people click on "Try in a Playground" without the app installed, they should be taken to a help page that explains the feature and offers the app for download.

I just investigated the technique for this a little bit and it looks like this will be a bit of JavaScript to test for the scheme, followed by a link to the help page.

### LINKED ISSUE #1031: Create AASA file for universal link
From previous setup: add AASA file to be served from `.well-known/apple-app-site-association` with content:

```json
{
    "applinks": {
        "apps": [],
        "details": [
            {
                "appID": "A42K5AU657.com.swiftpackageindex.playgrounds",
                "paths": ["/app/*"]
            }
        ]
    }
}
```

I don't quite remember the significance of `/app/*` in `paths`. I think that might be the deep link to open the app from `swiftpackageindex.com/app/`.

daveverwer:
Fixes #1031 
Fixes #934 

Not ready for merging yet. I need a style for images without a shadow and a better download button, but this is roughly what I think it should look like. All feedback welcome

Screen Shot 2021-04-14 at 20 05 04@2x

Screen Shot 2021-04-14 at 20 08 06@2x


Resources/Markdown/try-package.md
```diff
@@ -0,0 +1,17 @@
+---
+page-title: Try in a Playground
+description: Try any Swift package in a Swift Playground by downloading the Swift Package Index Playgrounds app for macOS.
+---
+
+[download-url]: https://github.com/SwiftPackageIndex/SPI-Playgrounds-Releases/raw/main/releases/SPI-Playgrounds-0.2.0.app.zip
```
Sherlouk, line 6: This feels like it could be frustrating to update in the future - could we duplicate 0.2.0 and call it `SPI-Playgrounds-latest.app.zip` and update that file every time in conjunction with storing each version?

daveverwer: Yea we've been chatting about this today and it won't ship like this. 👍

finestructure: I haven't found a good way to create a redirect or symbolic link to the latest version so I've simply updated our release `Makefile` to take care of the packaging.

It's now as simple as placing the `.app` bundle at the top level, running `make release VERSION=x.y.z`, and then doing `git push`.

The `make` command will copy the zip to `./releases/SPI-Playgrounds-$VERSION.app.zip` for the appcast feed but leave `SPI-Playgrounds.app.zip` at the top level for the latest release download.

The URL for is then: https://github.com/SwiftPackageIndex/SPI-Playgrounds-Releases/raw/main/SPI-Playgrounds.app.zip

I've just pushed 0.3.0 where all this is in place, so this URL is live.

finestructure: One thing we should check: is it ok to use GH as a file share / update service? I wouldn't think we're going to create a lot of traffic but it might trip something up.

daveverwer: > One thing we should check: is it ok to use GH as a file share / update service? I wouldn't think we're going to create a lot of traffic but it might trip something up.

I think we'll be fine, and if we get a warning (which I don't think we will) we can move it to the Azure CDN.

daveverwer:
I'm not very happy with how the image came out, but I've tried three different designs this morning that include the web page menu and this is the one I hate least! 😅


Screen Shot 2021-04-15 at 13 24 28@2x

Screen Shot 2021-04-15 at 13 24 40@2x

daveverwer:
Ready for review.

finestructure (approved):

PR #1043, opened 2021-04-22
Fix pinning of prometheus and grafana on a single node swarm

### LINKED ISSUE #1042: `node.labels.storage_location` cannot be set to both `prometheus_data` and `grafana_data` on `d1`
Prometheus wasn't running anymore on staging due to this. We need separate labels so they can be applied to a single node swarm.

finestructure:
Fixes #1042

### DIFF
--- a/mon.yml
+++ b/mon.yml
@@ -20,7 +20,7 @@ services:
       placement:
         constraints:
           # placement for data persistence
-          - node.labels.storage_location == grafana_data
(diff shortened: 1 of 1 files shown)


finestructure:
Draft until I've confirmed it's working on staging.

finestructure:
Ok, this is working and I've labeled prod as well.

daveverwer (approved):

PR #1061, opened 2021-05-04
Trigger 5.4 builds

finestructure:
I'll trigger this manually on staging before we tag and let this loose on prod.


finestructure:
The index page automatically shows the 5.4 builds, which I don't think is a problem while we're catching up. The table is what's most prominent.

I've updated the html snapshot but I can't run the image snapshot. Could you update that please before we merge? :)

daveverwer:
> The index page automatically shows the 5.4 builds, which I don't think is a problem while we're catching up.

I think it's an advantage. No one is browsing those pages, and it gives *us* a nice way to see what's happening for a package.

daveverwer:
Snapshots updated.

daveverwer (approved):

finestructure:
All good.

CleanShot 2021-05-04 at 10 41 00@2x

PR #1064, opened 2021-05-04
Remove AASA file

finestructure:
There are perhaps other reasons we may need/want an AASA file down the line but perhaps we don't want to keep it around until then?

Just a proposal to remove it in the meantime, it's easy enough to bring it back by reverting this commit.


Sources/App/routes.swift
```diff
@@ -32,14 +32,6 @@ func routes(_ app: Application) throws {
         MarkdownPage(path: req.url.path, "docs/builds.md").document()
     }
 
-    app.get(SiteURL.aasa.pathComponents) { req -> Response in
-        let path = Current.fileManager.workingDirectory()
-            .appending("Resources/apple-app-site-association.json")
```
daveverwer, line 37: We should delete this file too.

daveverwer (approved):
Yep, no need to keep this around. It's in history if we ever need it. Approved pending deletion of that last file I mentioned in the review.

PR #1066, opened 2021-05-04
Catch trigger-builds exceptions

finestructure:
I have no repro but it didn't raise exceptions in manual on staging. Might fix it, it might not, but there's certainly no harm in catch the exceptions.


finestructure:
I don't think it explains the missing triggers though but it'll hopefully remove a red herring and allow the metrics to be reported, for more clarity about what's going on.

daveverwer (approved):

PR #1068, opened 2021-05-05
Removed the "back" link on the SPI Playgrounds app download page

daveverwer:
We're sending people directly to this page from various places, so a JavaScript back button is going to send people all over the place. Better to just describe what they should do.

Also updated the URL to the new subdomain.


finestructure (approved):

finestructure:
This didn't require a snapshot update?

daveverwer:
Nope, it's a markdown file.

finestructure:
Yes, just surprised that we've not rendered that out in a test!

daveverwer:
We render one for testing, but not this one!

PR #1070, opened 2021-05-05
Send only "major.minor" Swift version build triggers

### LINKED ISSUE #1062: Trigger builds without Swift patch version
We're currently triggering builds with Swift version parameters like "5.3.0", which is overly specific. Our builder actually supports simply "5.3" as input and being specific can make builds fail if the versions don't match the exact definition of what what "5.3" means between trigger (server) and builder.

This [was a problem](https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/pull/1057) when I changed 5.3.0 to 5.3.3 on the builder side before updating the server.

We should simply trigger major.minor only, dropping the patch number to decouple server and builder.

finestructure:
Fixes #1062 

I'll manually test this on staging before we deploy

### DIFF
--- a/Sources/App/Core/Gitlab.swift
+++ b/Sources/App/Core/Gitlab.swift
@@ -63,7 +63,7 @@ extension Gitlab.Builder {
                         "BUILDER_TOKEN": builderToken,
                         "CLONE_URL": cloneURL,
                         "REFERENCE": "\(reference)",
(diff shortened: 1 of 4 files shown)


daveverwer (approved):
I'll let you merge when tested.

finestructure:
Working as expected: https://gitlab.com/finestructure/swiftpackageindex-builder/-/jobs/1237518395

CleanShot 2021-05-05 at 11 46 08@2x

PR #1075, opened 2021-05-09
Disable 5.4 builds

finestructure:
Once we've deployed this to prod, we should bump the builders back up to concurrency 4 so we can catch up. It clearly wasn't the problem.

I'll add an issue to run tests with specific packages we see taking long to figure out if this is perhaps due to specific packages.


daveverwer (approved):

daveverwer:
I can bump the concurrency up once it's deployed

finestructure:
Great. I'll fix this snapshot and then deploy.

PR #1084, opened 2021-05-10
Image snapshots for Swift 5.4 builds had been forgotten

daveverwer:


finestructure:
We'll need to remember to revert this when re-enabling 5.4, or re-create them then.

finestructure (approved):

PR #1086, opened 2021-05-10
Replace Parcel with Rollup

daveverwer:
The Parcel situation had become untenable. We were locked on a nightly, and more recent nightlies were not compatible. Rollup seems like a better fit, and is definitely more stable.

Don't merge yet, I'm having an issue with running it in the docker images.


daveverwer:
The test deployment worked and this is now ready to go.

@finestructure I'd be interested to know if `make build-front-end` and `make serve-front-end` work for you on the M1 now that we are no longer using Parcel.

finestructure (approved):

finestructure:
> That should have been the case for the last version too, but as far as I was aware you were not able to run it? I thought you said that it crashed, or somehow didn't work.

No, it was a recent update (last week) to Docker for Mac to support emulation for Intel images.

PR #1090, opened 2021-05-11
Switch to GitHub rendered README content

### LINKED ISSUE #824: Support anchor links in README files
See the Alamofire README for a great example of this. The table of contents at the start is full of relative links that don't work because the headings do not have `id` attributes, or anchors further down the document.

I don't think this is (easily) fixable client side, and I think it should probably be the responsibility of the Markdown parser. The way it works on GitHub is that the Markdown automatically generates anchor elements:

So, in the README.md, the heading for Donations is a simple heading:

Screenshot 2020-11-27 at 15 32 26@2x

The HTML generated looks like this:

Screenshot 2020-11-27 at 15 31 20@2x

It uses hyphens for spaces and lowercases everything. For example, the Resolved Radars generates this:

Screenshot 2020-11-27 at 15 32 53@2x

There are probably some more rules that I'm not considering, like apostrophes and other special characters. 

Ideally,  this would be an option on the GFM parser we're using. Let's check that before we implement anything custom.

Follow-on from #410

### LINKED ISSUE #828: Use single markdown -> HTML converter throughout
Bit of discussion in https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/pull/827 about this.

Currently we use Cmark+GFM for README markdown (on Package pages) and Ink for everything else (such as the privacy page). 

It'd be nice if eventually we removed the need for the Ink system to improve markdown compatibility and ensure we have consistency throughout the website.

This is not currently possible as we use the 'metadata' feature of Ink which allows for a small bit of key/value information at the top of a markdown file which is turned into the page title/description in MarkdownPage.swift. 

We'd need to reproduce this capability in the Cmark+GFM solution or approach it entirely differently. 

This is not a priority at this time but wanted to track it for future reference.

### LINKED ISSUE #830: Syntax highlighting affects build logs
It's nice and colourful but I don't think we want that really :)

Screenshot 2020-11-30 at 09 34 51

### LINKED ISSUE #843: Allow HTML tags/attributes to be whitelisted during Markdown processing
Example: https://staging.swiftpackageindex.com/guykogus/CodableJSON (See 'Installation' section)

A relatively common feature on READMEs is the use of the 'details' HTML tag in order to make collapsible sections of documentation.

These are currently completely ignored by our rendering of READMEs.

<details>
<summary>Title</summary>
Contents with a predetermined title
</details>

<details>
No Title defaults to 'Details'
</details>

Follow-on from #410

### LINKED ISSUE #1051: JavaScript Not Running on Async README
I believe this may have been caused by loading it in asynchronously, I assume the JS library is not being activated on the dynamically loaded content?

Example: https://swiftpackageindex.com/Sherlouk/AccessibilitySnapshotColorBlindness

daveverwer:
Fixes #1051
Fixes #830
Fixes #843
Fixes #824 
Fixes #828 

- [x] README content now downloaded from the rendered GitHub page rather than the Markdown.
- [x] README content is parsed and the relevant parts are extracted using [SwiftSoup](https://swiftpackageindex.com/scinfu/SwiftSoup)
- [x] Implement CSS styles for the GitHub flavoured syntax highlighting. [This gist](https://gist.github.com/Ibrahim-Islam/0661f4e60d5d5c81e22f) seems like a good place to start.
- [x] Remove GitHub specific UI from the rendered readme. For example, the heading anchor images and "copy code block" elements. See below for screenshots.
- [x] Anchor links are not working. They should not cause the page to reload, but they do.
- [x] Remove the `data-readme-base-url` attribute on the README `article` tag, if it's no longer needed. It appears that it may not be.
- [x] Remove old GFM Markdown classes and dependencies.
- [x] Remove the local highlighting JavaScript.


daveverwer:
Ready to review! Look at that glorious list of fixed issues too! 🚀

Sources/App/Views/PackageController/PackageReadme+Model.swift
```diff
@@ -1,26 +1,86 @@
...
+                    if (linkUrl.host == nil && linkUrl.path.starts(with: "/")) {
+                        guard let newLinkUrl = URL(string: "https://github.com\(linkUrl.absoluteString)")
+                        else { continue }
+                        try linkElement.attr("href", newLinkUrl.absoluteString)
+                    }
+                }
+            } catch {
+                // Errors are being intentionally eaten here. The worst that can happen if the
+                // HTML selection/parsing fails is that relative links don't get corrected.
+                return
```
daveverwer, line 82: This, and the one above, feels all sorts of wrong. However, I think it's the right choice.

It doesn't matter what kind of exception happens here. We don't want to break the page because of a badly formatted URL or link element.

Resources/Scripts/main.js
```diff
@@ -7,19 +7,20 @@ import { SPIPackageListNavigation } from './package_list_navigation.js'
...
 window.externalLinkRetargeter = new ExternalLinkRetargeter()
 window.spiPackageListNavigation = new SPIPackageListNavigation()
 window.spiCopyPackageURLButton = new SPICopyPackageURLButton()
 window.spiCopyBadgeMarkdownButtons = new SPICopyBadgeMarkdownButtons()
 window.buildLogNavigation = new SPIBuildLogNavigation()
-window.spiReadmeProcessor = new SPIReadmeProcessor()
 new SPIAutofocus()
 new SPIPlaygroundsAppLinkFallback()
 
+customElements.define('spi-readme', SPIReadmeElement)
```
daveverwer, line 22: First custom HTML tag! 🚀

finestructure (approved):

PR #1093, opened 2021-05-13
Add note exporter

finestructure:
CleanShot 2021-05-13 at 12 08 54@2x


daveverwer (approved):

finestructure:
Requires

```
docker service rm mon_prometheus
docker config rm mon_prometheus_cfg
```

on prod before deploying, because configs can't be updated.

finestructure:
Bloody autocomplete, only just seen that the branch ended up being called `note` explorer...

finestructure:
Tagging and deploying now

PR #1097, opened 2021-05-14
Add 5.4 to build matrix, remove 4.2

finestructure:
I've recreated one screenshot as an example but they'll all need updating with your machine.

Let's not deploy this today though :)

Fixes #1063

PR #1101, opened 2021-05-17
Fixes #974

### LINKED ISSUE #974: Internal server error when searching for `*`
https://swiftpackageindex.com/search?query=*

finestructure:
Fixes #974 

Merge after #1097


daveverwer (approved):
I know it's unlikely we'll ever see packages with these characters in their names, but would it have been possible to escape rather than sanitize?

finestructure:
Let me check this again but IIRC there was no escape I could get to work.

PR #1106, opened 2021-05-18
Fix build metrics scrape target

finestructure:
When we changed to docker swarm deployment I changed the app container's name from `app` to `app_server`. This broke the scrape target in prometheus' config and is the reason the "Build Reports" panel is blank:

CleanShot 2021-05-18 at 15 26 39@2x

### DIFF
--- a/restfiles/post-build.restfile
+++ b/restfiles/post-build.restfile
@@ -1,6 +1,6 @@
 variables:
-    # base_url: http://localhost:8080/api
-    base_url: https://staging.swiftpackageindex.com/api
+    base_url: http://localhost:8080/api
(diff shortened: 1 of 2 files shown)


restfiles/post-build.restfile
```diff
@@ -15,10 +15,9 @@ requests:
             Authorization: Bearer ${builder_token}
         body:
             json:
-                logs: some logs
-                platform:
-                    name: unknown
-                    version: test
```
finestructure, line 21: Outdated build report format, updated while debugging this

daveverwer (approved):

finestructure:
CleanShot 2021-05-18 at 16 43 29@2x

PR #1111, opened 2021-05-19
Package update

finestructure:


daveverwer (approved):

finestructure:
There must be a Linux conditional in Plot where it's using the old url:

```
@@ −1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
−<rss version="2.0" xmlns:atom="http://www.w3.org/2005/Atom" xmlns:content="http://purl.org/rss/1.0/modules/content/">
+<rss version="2.0" xmlns:atom="http://www.w3.org/2005/Atom" xmlns:content="http://purl.org/rss/1.0/modules/content">
   <channel>
     <title>feed title</title>
     <description>feed description</description>
     <link>https://SwiftPackageIndex.com</link>
Test Case 'RSSTests.test_render_feed' failed (0.056 seconds)
```

Issue #980, opened 2021-03-08
Use the GitHub organisation title when available


helje5:
Small feature request. The index currently uses the organisation part of the repository URL as the organisation "title". For example "Macro-swift" in here (but also appears in other contexts):

![Screenshot 2021-03-08 at 17 40 58](https://user-images.githubusercontent.com/7712892/110351759-73d6a180-8035-11eb-81f1-0c0f8ef598c9.png)

It would be cool if those places would actually use the GitHub organisation name, when available. E.g. "Macro.swift" in this case: https://github.com/Macro-swift:

![Screenshot 2021-03-08 at 17 42 29](https://user-images.githubusercontent.com/7712892/110351985-a8e2f400-8035-11eb-8b7c-4419287d0f41.png)

(the two usually differ if the organisation name uses character which are not allowed in the GH backend, e.g. the `.` in this case).

daveverwer:
Great idea. We'll collect that information as part of #975. Thanks Helge

elfenlaid:
It's hard to find all the appropriate places. Though, here's what I've got after a search for `model.owner` field usage and following the discussion:

- The author's (organization) page:
    - https://swiftpackageindex.com/Macro-swift
- The package's page
    - "More packages from this author" -> "More packages from <organization>"?

What do you think about truncating long organization names?

- The author's page 
![Screenshot 2021-05-25 at 14 55 25](https://user-images.githubusercontent.com/765790/119509353-3dc5d580-bd79-11eb-80fe-94bf9cbb6bf7.png)
- The package's page
![Screenshot 2021-05-25 at 16 47 17](https://user-images.githubusercontent.com/765790/119509329-38688b00-bd79-11eb-9f12-31eb515e69ae.png)

daveverwer:
This is great progress. Some things to note:

* Looking at the real-world data we've gathered, better test data would include spaces. There are some long owner names, up to 38 characters to be specific, but they all have several spaces in them. I'd imagine wrapping will work better than truncation.
* There may be some extreme cases in the future where we get a very long owner name without spaces. I'd suggest we cope with this as and when it happens as I don't think it's likely to be a big problem.
* We should fallback to the owner username when the owner name is blank, there are 102 owners with a `NULL` owner_name

I've attached a CSV file of the longest owner names. I've also DMed you a copy of the latest database so you can test with real-world data.

Issue #975, opened 2021-03-05
Fetch the URL for the account/organisation along with the GitHub metadata


daveverwer:
Ideally via the GraphQL request, I think it's the [`avatar_url`](https://docs.github.com/en/rest/reference/users) property we need.

Basically, this image here:

<img width="393" alt="Screen Shot 2021-03-05 at 17 33 28@2x" src="https://user-images.githubusercontent.com/5180/110151884-ea2a9800-7dd8-11eb-8fe6-9cf0b36e8262.png">

daveverwer:
We should also fetch the organisation/account full name while we do this to make #980 possible

elfenlaid:
I'll have a look at it and #980 if everyone is okay with that

elfenlaid:
By the way, what are the next steps for the issue after the data is collected?

I assume the avatar should appear somewhere alongside the package's name at the package's page header?

daveverwer:
> By the way, what are the next steps for the issue after the data is collected?

#980 follows on from this for the organisation name. The avatar is a little different and doesn't currently fit well into the site design. I have a couple of ideas where we could use it, but no plans immediately for that info.

elfenlaid:
> > By the way, what are the next steps for the issue after the data is collected?
> 
> #980 follows on from this for the organisation name. The avatar is a little different and doesn't currently fit well into the site design. I have a couple of ideas where we could use it, but no plans immediately for that info.

Ah, got it. So we can probably close this issue at this point or otherwise rename it 🤔

daveverwer:
Yep, this one is done! Thanks @elfenlaid

PR #1116, opened 2021-05-20
Fetch additional package metadata

elfenlaid:
Additional fields are:

- `ownerAvatarUrl` for #975
- `ownerAvatarName` as referenced [in the discussion](https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/pull/1116#issuecomment-845024446)
- `isInOrganization` for #980

From my understanding of Github's GraphQL API, the `Repository.owner` field is shared by both `User` and `Organization` entities. E.g., if a repository belongs to an organization, `login` and `avatarUrl` should point to the organization's metadata. Otherwise, `login` and `avatarUrl` point to the author's account. ([Explorer configuration for API probing](https://gist.github.com/elfenlaid/61d5a72a20278eca948e6e9ff2a5b62c))

So, it's not clear only from the `owner` field whether the author is an organization or a person. Thus, the additional `Repository.isInOrganization` field helps to distinguish the owner's type.

Probably I've missed something (maybe you have a running revamp plan of Repo's ownership), so looking forward to the feedback :)

### DIFF
--- a/Sources/App/Core/Github.swift
+++ b/Sources/App/Core/Github.swift
@@ -259,6 +259,7 @@ extension Github {
                     }
                     owner {
                       login
+                      avatarUrl
                     }
                     releases(first: 20, orderBy: {field: CREATED_AT, direction: DESC}) {
                       nodes {
@@ -270,6 +271,7 @@ extension Github {
                       }
                     }
                     stargazerCount
+                    isInOrganization
                   }
                 }
                 """)
@@ -292,6 +294,7 @@ extension Github {
             var owner: Owner
             var releases: ReleaseNodes
             var stargazerCount: Int
+            var isInOrganization: Bool
             // derived properties
             var defaultBranch: String? { defaultBranchRef?.name }
             var lastIssueClosedAt: Date? {
@@ -335,6 +338,7 @@ extension Github {
 
         struct Owner: Decodable, Equatable {
             var login: String
+            var avatarUrl: String
         }
 
         struct OpenPullRequests: Decodable, Equatable {
--- a/Sources/App/Core/Github.swift
+++ b/Sources/App/Core/Github.swift
@@ -260,6 +260,12 @@ extension Github {
                     owner {
                       login
                       avatarUrl
+                      ... on User {
+                        name
+                      }
+                      ... on Organization {
+                        name
+                      }
                     }
                     releases(first: 20, orderBy: {field: CREATED_AT, direction: DESC}) {
                       nodes {
@@ -338,6 +344,7 @@ extension Github {
 
         struct Owner: Decodable, Equatable {
             var login: String
+            var name: String
             var avatarUrl: String
         }
 
--- a/Sources/App/Core/Github.swift
+++ b/Sources/App/Core/Github.swift
@@ -344,7 +344,7 @@ extension Github {
 
         struct Owner: Decodable, Equatable {
             var login: String
-            var name: String
+            var name: String?
             var avatarUrl: String
         }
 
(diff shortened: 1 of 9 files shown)


Sources/App/Core/Github.swift
```diff
@@ -259,6 +259,13 @@ extension Github {
                     }
                     owner {
                       login
+                      avatarUrl
+                      ... on User {
+                        name
+                      }
+                      ... on Organization {
+                        name
+                      }
```
finestructure, line 268: Interesting, I had no idea there's an `...` operator!

I'm guessing this is sort of like an enum or union type for `name`. Is there a guarantee in the schema that there'll always be either a `User` or an `Organisation`? Or do we potentially need to make this `String?` in our `Owner` `Decodable`?

The graphql console isn't working for me at the moment so I can't check unfortunately.

Sources/App/Core/Github.swift
```diff
@@ -335,6 +344,8 @@ extension Github {
 
         struct Owner: Decodable, Equatable {
             var login: String
+            var name: String
+            var avatarUrl: String
```
finestructure, line 348: Just a reminder to confirm these two are `String!` in the GraphQL schema.

Sources/App/Migrations/027/UpdateRepositoryAddOwnerFields.swift
```diff
@@ -0,0 +1,20 @@
+import Fluent
+
+
+struct UpdateRepositoryAddOwnerNameAndOwnerAvatarUrlAndIsInOrganizationFlag: Migration {
```
finestructure, line 4: Nit: I'd keep file and struct name in sync, i.e. just call this `UpdateRepositoryAddOwnerFields` (I don't think it collides with any previous name). In hindsight we should probably have namespaces these with `Migration001` etc.

finestructure (approved):
Looks great, thanks @elfenlaid !

Just added a couple of comments and questions but otherwise 👍

Sources/App/Core/Github.swift
elfenlaid: TL;DR:

- `avatarUrl` and `login` are safe to leave as `String`
- `name` should be `String?`

Oh, there're two instances of `...` in GraphQL I'm aware of:

- [GraphQL Fragments](https://graphql.org/learn/queries/#fragments) - they preserve query's locality
- [GraphQL Interfaces](https://graphql.org/learn/schema/#interfaces) and [GraphQL Unions](https://graphql.org/learn/schema/#union-types) - `... on Organization` is similar to `if let organization = owner as? Organization` to my understanding

So, long story short, there's `interface RepositoryOwner` and two instances that conform to it: `User` and `Organization`. `Repository` declares `owner: RepositoryOwner!`

- `avatarUrl: URI!` and `login: String!` are the `RepositoryOwner` interface part. Thus guaranteed to be presented
- While  `User` and `Organization` both contain the` name: String` parameter, it's not part of the `RepositoryOwner` interface

The query means "Hey if the owner typecasts to User or Organization types, get its name":

```graphql
... on User {
    name
}
... on Organization {
    name
}
```

Sources/App/Migrations/027/UpdateRepositoryAddOwnerFields.swift
elfenlaid: > I'd keep file and struct name in sync 

Oops 😳

> (I don't think it collides with any previous name)

Presumable the unique index on `_fluent_migrations.name` should fail if names collide

> In hindsight we should probably have namespaced these with Migration001 etc.

Yeah, maybe [timestamps](https://github.com/hexpm/hexpm/tree/main/priv/repo/migrations ) instead of the incremental counter should work as well. Counters are more prone to conflicts.

finestructure:
Closing and re-opening to make the tests run

finestructure:
Superseded by #1127

PR #1118, opened 2021-05-20
Introduce Structured Data (JSON-LD)

### LINKED ISSUE #1092: Implement JSON-LD on Package pages
https://schema.org/SoftwareSourceCode

Sherlouk:
(Will) Closes #1092 

@daveverwer - Let's discuss what we want to actually be inside of this schema 👍 

Example of one of these types:
```
{
  "@context": ["http://schema.org/"],
  "@type": "SoftwareSourceCode",
  "identifier": "@mona/LinkedList",
  "name": "LinkedList",
  "description": "One thing links to another.",
  "keywords": ["data-structure", "collection"],
  "version": "1.1.1",
  "codeRepository": "https://github.com/mona/LinkedList",
  "license": "https://www.apache.org/licenses/LICENSE-2.0",
  "readme": "https://github.com/mona/LinkedList/blob/master/README.md",
  "issueTracker": "https://github.com/mona/LinkedList/issues",
  "programmingLanguage": {
    "@type": "ComputerLanguage",
    "name": "Swift",
    "url": "https://swift.org"
  },
  "author": {
      "@type": "Person",
      "@id": "https://github.com/mona",
      "givenName": "Mona",
      "middleName": "Lisa",
      "familyName": "Octocat"
  }
}
```


daveverwer:
Thanks for this James! busy day today, so I may not get to this today, but I will get to it!

Looks like you've put the obvious information in already, which is great. I'll check to see if there's anything else we can add. Otherwise, we'll amend this as we add authors, keywords, etc etc...

Sherlouk:
Just looking at some SEO garb, we may want to look into adding stuff like this too? https://developers.google.com/search/docs/data-types/sitelinks-searchbox

daveverwer:
> Just looking at some SEO garb, we may want to look into adding stuff like this too? https://developers.google.com/search/docs/data-types/sitelinks-searchbox

Great idea, but as you suggest it'd be better as a separate PR. Tracked as #1129.

daveverwer:
Alright, I finally spent some time with this. The output looks great and validates successfully. Here's my thoughts on fields:

* We should set the `url` to the package page on the SPI.
* We may not know that the package *only* contains Swift, but is it fair to say that every package includes Swift? If so, we could add `programmingLanguage` and set it to Swift, either as plain text or as a [ComputerLanguage](https://schema.org/ComputerLanguage)?
* Could we set `runtimePlatform`? My best guess at what we should set it to would be the latest version of Swift we have verified support for.
* Could we set `targetProduct`? The biggest issue by far here is that we have compatibility information for multiple platforms and this seems to only support one [SoftwareApplication](https://schema.org/SoftwareApplication).
* We could set `dateCreated` to be the date of the first commit, which I believe we have stored in `first_commit_date` in `repositories`.
* We could set `dateModified` to be the date of the most recent commit which we store as `last_commit_date` in `repositories`.
* Could we use `sourceOrganization` for the organisation name. If so, let's use the name which was just added in #1127 rather than the `repositoryOwner` field.
* Is `text` the content of the README? I'm really not sure about this one. This might be too much of a stretch.

Fields for later:
* `creator` and/or `maintainer` - Once we have author information. Although why can't this be an array? Ugh. We could potentially use `creditText` if a thing can truly only have one creator.
* `keywords` - Once we have keywords!

daveverwer:
> * Is `text` the content of the README? I'm really not sure about this one. This might be too much of a stretch.

I already think this is too much of a stretch, and we also no longer have a plain text representation of this data.

Sherlouk:
Sounds like some good information, leave it with me and I'll flesh it out some more!

Sherlouk:
Sorry for the delay getting back round to this. Few things happening in my personal/work life limiting my current time available - I'll get an update out as soon as possible!

Sherlouk:
Think we're closer now. 

`programmingLanguage` - unsure on this one. In theory it's quite easy for a package to not have Swift, but the broad majority definitely will? We can include it if you want, but it would be for all.

`targetProduct` - I can't see a nice way to handle this as we have multiple target products and the schema expects only one.

daveverwer:
Great work James! Thanks for your continued efforts on this.

> `programmingLanguage` - unsure on this one. In theory it's quite easy for a package to not have Swift, but the broad majority definitely will? We can include it if you want, but it would be for all.

Yes, I think this is fine. We don't have definitive information for this but it's safe to say that the overwhelming majority of packages will be Swift, and all of them work with Swift, I'd like to add this.

> `targetProduct` - I can't see a nice way to handle this as we have multiple target products and the schema expects only one.

The only thing I can think here is that maybe there's implicit support in these schemas for arrays on these keys. The one I keep coming back to is `author`. Surely they can't have expected software source code to only have a single person author when designing this schema, and yet it's a singular `Organization` or `Person` object. I have a feeling there might be an implicit array on some of these types.

It might be worth adding an array and seeing if it validates through the [testing tool](https://search.google.com/structured-data/testing-tool/)?

Sherlouk:
I've tried hand crafting the JSON through the validation tool using an array and it seems to accept it - so will give it a go. Will add the Swift language too!

Leave it with me

Sherlouk:
I've looked more into the exact needs of the SoftwareApplication which is what targetProduct come up under, and I'm not sure we can 100% tick the boxes. See here for more information: https://developers.google.com/search/docs/data-types/software-app#softwareapplication

It's no longer throwing errors if I define the OS as a "DeveloperApplication", but it doesn't feel accurate. The names we have are also not super appropriate, they seemingly also want versions not just OS names. I think this is something we'd want to pull from the `Package.swift` "platforms" section?

daveverwer:
I agree. Let's forget `targetProduct` then 👍

daveverwer:
Testing

daveverwer (commented):
I started to review this properly this morning and just pushed a few changes. 

* I moved the structured data outside the nested divs and up with the HTML comments. We may want to rename `bodyComments` at some point, but I think this is OK for now, and this location is certainly a better place for the markup to live.
* I moved the Plot extension inside the Plot extensions file. I'm trying to keep all Plot extensions in one place, even those that will never be upstreamed, like this.
* I renamed the `Schema` protocol as it was a bit generic for a top-level type.

However, I just had a quick chat with @finestructure about this and we were wondering if we could live without that protocol at all. Could we make `structuredData` take any Codable? I ran out of time to investigate it properly so I thought I'd pose is as a question instead 😰

Sherlouk:
> Could we make `structuredData` take any Codable?

Yes. I made the decision to add a small amount of safety by requiring the two schema fields. It would work absolutely fine without that requirement, but of course opens up to the accidental misuse of the API.

If you folks prefer to remove that, I'm happy to make that change 🙂

finestructure:
> small amount of safety by requiring the two schema fields

Ah, I see. I actually brought up the protocol use because it wasn't clear to me without that context. What's interesting is that if you drop the protocol and just require `Codable` for `PackageSchema` it surfaces a small bug: `CodingKeys` is missing `sourceOrganization`. (@daveverwer that's why it didn't compile for you while we were screen sharing.)

I think in this case it may be better to let the compiler enforce `Codable` via the protocol and simply ensure `context` and `type` via tests. Just my 2 cents!

Sherlouk:
Patched the issue raised above (good spot!) and removed the protocol in favour of a direct Encodable solution.

daveverwer (approved):
Love this PR. I'll rebase it on `main` and then merge. I want to also push a Plot change live with it too so I'll coordinate both and deploy this afternoon.

Thanks James!

Sherlouk:
Thanks Dave! Should give us a foundation to improve on as we add more features and data points. Of course we've just added keywords (topics) so we can probably follow this up quite quickly with those.

daveverwer:
Deploying as 2.40.1

PR #1120, opened 2021-05-21
Reset analysis metrics, switch all Counters to Gauges

### LINKED ISSUE #932: Investigate "New Versions" grafana panel display
Our "New Version" panel doesn't make sense and needs closer inspection.

CleanShot 2021-02-16 at 09 28 39@2x

If this is a count it should be increasing. It's either wrongly named our the count is getting reset due to the push gateway in between.

Or actually - since we're running this from a process, the problem is probably that the count is being reset on each invocation. That's no good.

finestructure:
I'll deploy from branch and test on staging before marking ready for review.

Fixes #932 

- [x] test on staging
- [x] apply to other batch process metrics if successful

### DIFF
--- a/Sources/App/Commands/Analyze.swift
+++ b/Sources/App/Commands/Analyze.swift
@@ -23,7 +23,7 @@ struct AnalyzeCommand: Command {
         let logger = Logger(component: "analyze")
         let threadPool = context.application.threadPool
 
-        resetMetrics()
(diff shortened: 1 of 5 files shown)


daveverwer (approved):

finestructure:
Looking better now:

CleanShot 2021-05-21 at 15 18 12@2x

PR #1131, opened 2021-05-24
Add SPI Playground app launch analytics

### LINKED ISSUE #1130: Add Plausible event for opening the SPI Playgrounds app
Attach to any `spi-playgrounds:` link as it's impossible to see what usage the app is truly getting.

daveverwer:
Fixes #1130

### DIFF
--- a/Resources/Scripts/playgrounds_app_link.js
+++ b/Resources/Scripts/playgrounds_app_link.js
@@ -1,3 +1,5 @@
+import { measurePlausibleEvent } from './plausible_analytics.js'
+
 export class SPIPlaygroundsAppLinkFallback {
   constructor() {
     document.addEventListener('click', (event) => {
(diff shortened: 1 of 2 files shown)


Resources/Scripts/playgrounds_app_link.js
```diff
@@ -29,6 +31,11 @@ export class SPIPlaygroundsAppLinkFallback {
         setTimeout(() => {
           explainerElement.classList.add('hidden')
         }, 5000)
+
+        // Trigger an analytics event and pass the dependency as a custom property
+        const targetUrl = new URL(linkElement.href)
+        const dependencies = targetUrl.searchParams.get('dependencies')
+        measurePlausibleEvent('SPI Playgrounds Launch', { props: { dependency: dependencies } })
```
daveverwer, line 38: Deliberately using the singular `dependency` here so we can measure this in a different way if/when we support multiple dependencies.

finestructure: So is this reporting on any url with `dependencies=` as a parameter? Should we also look for `spi-playgrounds://open` in the url? Or is this configured server side somewhere?

finestructure (approved):

daveverwer:
Tagged as `2.35.1`

Resources/Scripts/playgrounds_app_link.js
daveverwer: This code only executes on those links, that check is further up ^^

finestructure: Oh, I see, it was already in and not part of the diff :+1:

PR #1139, opened 2021-05-26
Use organization title or user name as display owner name

### LINKED ISSUE #980: Use the GitHub organisation title when available

elfenlaid:
Resolves #980

Two pairs of screenshots to demonstrate the text truncation behavior. The long words are going to be truncated; the rest will be wrapped to the following line.

- Reeeally not sure about the `Package's page` changes 🤔
- UI tests should fail at this point
- CSS classes probably need a more delicate attitude

### Author's Page

Screenshot 2021-05-26 at 11 19 08
Screenshot 2021-05-26 at 11 22 21

### Package's Page

Screenshot 2021-05-26 at 11 22 35
Screenshot 2021-05-26 at 11 25 24

### DIFF
--- a/Resources/Styles/base.scss
+++ b/Resources/Styles/base.scss
@@ -178,3 +178,12 @@ a.download {
   visibility: hidden;
   transition: visibility 0s 0.2s, opacity 0.2s linear;
 }
+
+.trimmed {
+  overflow: hidden;
+  text-overflow: ellipsis;
+}
+
+.collapsable {
+  min-width: 0;
+}
--- a/Sources/App/Controllers/AuthorController.swift
+++ b/Sources/App/Controllers/AuthorController.swift
@@ -33,6 +33,7 @@ struct AuthorController {
             .map {
                 AuthorShow.Model(
                     owner: $0.first?.repository?.owner ?? owner,
+                    ownerName: $0.first?.repository?.ownerDisplayName ?? owner,
                     packages: $0.sorted(by: { $0.score ?? 0 > $1.score ?? 0 })
                                 .compactMap { AuthorShow.PackageInfo(package: $0) }
                 )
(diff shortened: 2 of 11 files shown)


finestructure:
Looks great! Just a thought that occurred to me while looking at those: For the sidebar, we could simply replace long owner names with `More packages from this author` beyond a certain length.

Maybe it's even a change we'll want to use in general. Buttons benefit from being static and the meaning is clear when you're on a package page.

elfenlaid:
> Maybe it's even a change we'll want to use in general. Buttons benefit from being static and the meaning is clear when you're on a package page.

The same thoughts as well, I think the dynamic title looks a bit off with the button's case

daveverwer:
> For the sidebar, we could simply replace long owner names with `More packages from this author` beyond a certain length.

I remember at the time we added this button we struggled with this too and almost went with "More packages from this author". Some display names are much longer than the usernames are, and so this is an issue.

I think, rather than try and do two different things on this button, we should just change this button to always be "More packages from this author" and not try to be dynamic based on the length.

elfenlaid:
> I think, rather than try and do two different things on this button, we should just change this button to always be "More packages from this author" and not try to be dynamic based on the length.

Totally agree, that's where the PR is now

daveverwer (approved):
This is wonderful, thanks so much @elfenlaid!

daveverwer:
Ready to merge. Could you rebase or merge main back in and push that and it will auto-merge.

PR #1141, opened 2021-05-27
Extend product library type

### LINKED ISSUE #878: Extend `Product.Type.library`
One field currently not matching the required data format for package collections is `Product.Type` for the `library` case.

The package collection format defines it as follows:

```swift
public enum ProductType: Equatable {
    public enum LibraryType: String, Codable {
        case `static`
        case `dynamic`
        case automatic
    }

    case library(LibraryType)
    case executable
    case test
}
```

We're currently using

```swift
struct Product: Decodable, Equatable {
    enum `Type`: String, CodingKey, CaseIterable {
        case executable
        case library
    }
    ...
}
```

The extra `test` case is easy enough to add but the associated value on `library` is going to be harder to support.

We need to consider existing values in the db that are just `.library` while also handling new values `.library(static)` etc.

We need to do the following:

- [x] define new library type data format
- [x] add migration that transforms the column (pretty sure we need a format change to `.json` to handle the structure, i.e. we need to drop and add the column
  - [x] need to decide if we want to leave it to re-processing to re-populate the column or if we'll try to migrate data
  - [x] if we migrate data, treat current `.library` as `.library(automatic)` (this isn't necessarily correct in the sense that it's the same as the package manifest but it's presumably semantically ok, at least for the purposes of the package collection)
  - [x] then again, this might be the argument to just clear out the column and re-process
- [x] ensure all tools versions we support are data format compatible with the new version, i.e. do they all dump `.library(...)` and not just `.library`

finestructure:
~Currently blocked by https://github.com/vapor/postgres-kit/issues/207~

~There might be ways to work around the issue by using a different custom encoding format but let's see if there's perhaps an upstream fix available. That would allows us to get rid of the custom encoding in Swift 5.5.~

Simple workaround in place that's easy to remove should the upstream issue be fixed.

I'll leave this in draft while I deploy from the branch to let staging run an ingestion cycle with the schema update and product type update.

~Actually using the new format for package collections will be a separate PR.~

I've added this to this PR, because the change was actually trivial.

Fixes #878

### DIFF
--- a/Sources/App/Models/Manifest.swift
+++ b/Sources/App/Models/Manifest.swift
@@ -29,24 +29,42 @@ struct Manifest: Decodable, Equatable {
         var platformName: Name
         var version: String
     }
-    struct Product: Decodable, Equatable {
-        enum `Type`: String, CodingKey, CaseIterable {
+
+    enum LibraryType: String, Decodable {
+        case automatic
+        case `dynamic`
+        case `static`
+    }
+
+    enum ProductType: Equatable {
+        case executable
+        case library(LibraryType)
+        case test
+
+        enum CodingKeys: CodingKey {
             case executable
             case library
+            case test
         }
+    }
+
+    struct Product: Decodable, Equatable {
         var name: String
         var targets: [String] = []
-        var type: `Type`
+        var type: ProductType
     }
+
     struct Target: Decodable, Equatable {
         var name: String
     }
+
     struct ToolsVersion: Decodable, Equatable {
         enum CodingKeys: String, CodingKey {
             case version = "_version"
         }
         var version: String
     }
+
     var name: String
     var platforms: [Platform]?
     var products: [Product]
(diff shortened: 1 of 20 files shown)


Sources/App/Models/Manifest.swift
```diff
@@ -56,18 +74,21 @@ struct Manifest: Decodable, Equatable {
...
+        }
+        switch key {
+            case .executable:
+                self = .executable
+            case .library:
+                var unkeyedValues = try container.nestedUnkeyedContainer(forKey: key)
+                let value = try unkeyedValues.decode(Manifest.LibraryType.self)
+                self = .library(value)
+            case .test:
+                self = .test
```
finestructure, line 91: `dump-package` uses a different JSON format for its associated values (one that's not Swift 5.5 forward compatible) and we need to have our own custom decoding for it.

finestructure:
Right, because `Product`s are attached to `Version`s, we need to force a version update in order to re-populate `products.type`.

I'll trigger this now on staging.

daveverwer (approved):

finestructure:
Starting to populate but it's going to take a while to process all versions.

CleanShot 2021-06-01 at 10 37 20@2x

We have almost 60k versions and the biggest batch size I dare submit are 1k, because this goes up to around half the memory available on `d1`:

CleanShot 2021-06-01 at 10 39 15@2x

I think I need to change how we're re-analysing versions so the batch size can be 10k or more but we break this down into smaller batches and flush within the command.

finestructure:
For reference, batch size 1k takes 70m to process on `d1` (with the full batch in memory).

finestructure:
This is almost done processing on staging now, it's taken quite a while and needs monitoring due to memory (despite small batches). I'll tag and deploy this to prod tomorrow morning first thing and then re-process versions there.

finestructure:
Not exactly sure what is going on but running `re-analyze-versions` seems to be leaking memory like crazy. Even if you run it as a one-off command `exec`d into a container does it increase the container memory without releasing it on termination.

And this isn't just a display artefact of `ctop`: staging just ground to a halt because the system was swapping. Even on reboot it just happily reinstantiated the container with 6.3G of memory allocated.

I've run `re-analyze-versions` in the past and while I didn't monitor memory with `ctop` a the time, I didn't notice and slow-downs, either. Weird. And I don't want to let this loose on prod if it behaves like this.

finestructure:
Processing on staging is complete. I've added resource limits for prod deployment so it'll be ok to run the re-analysis in the container (even though it's not clear where the memory bloat is originating from).

finestructure:
There is _no_ memory issue when running the version update on prod. There are a few minor differences in setup and I'm having to inspect memory consumption with a different tool for uninteresting reasons. It's all quite puzzling but not to the point where I'm going to worry about this further :)

PR #1167, opened 2021-06-10
Collections performance

finestructure:
Before:

before

After:

after

### DIFF
--- a/Sources/App/Core/PackageCollection+generate.swift
+++ b/Sources/App/Core/PackageCollection+generate.swift
@@ -14,10 +14,7 @@ extension PackageCollection {
     typealias Product = PackageCollectionModel.V1.Product
     typealias ProductType = PackageCollectionModel.V1.ProductType
     typealias Target = PackageCollectionModel.V1.Target
-}
 
-
-extension PackageCollection {
     enum Filter {
         case urls([String])
         case author(String)
--- a/Sources/App/Core/PackageCollection+generate.swift
+++ b/Sources/App/Core/PackageCollection+generate.swift
@@ -23,7 +23,7 @@ extension PackageCollection {
                          overview: String? = nil,
                          revision: Int? = nil) -> EventLoopFuture<PackageCollection> {
         versionQuery(db: db)
-            .filter(\.$url ~~ packageURLs)
+            .filter(App.Package.self, \.$url ~~ packageURLs)
             .all()
             .map { versions in
                 Dictionary(grouping: versions, by: { $0.package })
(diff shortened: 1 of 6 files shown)


daveverwer (approved):
Well this looks promising!

Sources/App/Core/PackageCollection+generate.swift
```diff
@@ -15,42 +15,49 @@ extension PackageCollection {
...
+                    .filter(App.Package.self, \.$url ~~ packageURLs)
+        }
+
+        return query.all()
+            .map { versions in
+                // Multiple versions can reference the same package, therefore
+                // we need to group them so we don't create duplicate packages.
+                // This requires App.Package to be Hashable and Equatable.
+                Dictionary(grouping: versions, by: { $0.package })
+                    .sorted(by: { $0.key.url < $1.key.url })
```
finestructure, line 56: I tried to avoid making `Package` conform to `Equatable` and `Hashable`, because it’s quite a sweeping change and basing it solely on `id` comparison (while valid in this case) may not be ideal generally.

I used an `IDHashable<T: Identifiable>` container and while it worked it caused some (solvable) downstream issues. (For instance, `Package` would need to be `Identifiable` which, while seemingly straightforward doesn’t compile.)

A non-generic variant worked but I’m not sure it’s an improvement over just making `Package` conform directly.

Tests/AppTests/PackageTests.swift
```diff
@@ -6,7 +6,27 @@ import XCTVapor
...
 
 final class PackageTests: AppTestCase {
-    
+
+    func test_Equatable() throws {
+        XCTAssertEqual(Package(id: .id0, url: "1".url),
+                       Package(id: .id0, url: "2".url))
+        XCTAssertFalse(Package() == Package())
+        XCTAssertFalse(Package(id: .id0, url: "1".url) == Package())
+    }
```
finestructure, line 15: Note the equatable behaviour.

Tests/AppTests/PackageTests.swift
```diff
@@ -6,7 +6,27 @@ import XCTVapor
...
+            Package(id: .id0, url: "1".url),
+            Package(id: .id0, url: "2".url),
+            Package(url: "3".url),
+            Package(url: "4".url)
+        ]
+        XCTAssertEqual(packages.map { "\($0.id)" }.sorted(),
+                       [UUID.id0.uuidString, "nil", "nil"])
+        XCTAssertEqual(packages.map { "\($0.url)" }.sorted(),
+                       ["1", "3", "4"])
+    }
```
finestructure, line 28: Note the hashable behaviour.

finestructure:
This is good to go now.

I’ve added some FYI comments and I’ll go ahead and merge now.

We should test a few notable collections on staging to ensure nothing’s broken before pushing live.

PR #1170, opened 2021-06-11
Async await

finestructure:
Draft PR for illustration purposes and as a WIP for Swift 5.5 + async/await adoption.

Initial test conversion of the `reconcile` command, the first of a number of commands.

All tests are passing with a Swift 5.5 toolchain. It's only failing in CI because we're testing with 5.4 for the moment.

### DIFF
--- a/Sources/App/Commands/Reconcile.swift
+++ b/Sources/App/Commands/Reconcile.swift
@@ -6,7 +6,7 @@ struct ReconcileCommand: Command {
     struct Signature: CommandSignature { }
     
     var help: String { "Reconcile package list with server" }
-    
+
     func run(using context: CommandContext, signature: Signature) throws {
         let logger = Logger(component: "reconcile")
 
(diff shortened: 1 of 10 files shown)


Sources/App/Commands/Reconcile.swift
```diff
@@ -50,14 +56,13 @@ func diff(source: [URL], target: [URL]) -> (toAdd: Set<URL>, toDelete: Set<URL>)
...
-    let insert = toAdd.map { Package(url: $0, processingStage: .reconciliation) }.create(on: db)
-    let delete = toDelete
-        .map { url in
-            Package.query(on: db)
-                .filter(by: url)
-                .delete()
-        }.flatten(on: db.eventLoop)
-    return insert.and(delete).transform(to: ())
+    let insert = toAdd.map { Package(url: $0, processingStage: .reconciliation) }
+    try await insert.create(on: db)
```
finestructure, line 62: TODO: collapse

PR #1171, opened 2021-06-11
Fix collections module name

### LINKED ISSUE #1169: Include `moduleName` in package collections
@neonacho via Slack:

> I think there is a small bug around modules, what I see in the generated collection is this:
> ```json
> "targets": [
>                 {
>                   "name": "CombineSchedulers"
>                 },
>                 {
>                   "name": "CombineSchedulersTests"
>                 }
>               ]
> ```
> 
> In the collections we generate, there is an additional "moduleName" property that we rely on for autocompletion in Xcode.
> 
> Would be possible for y’all to add that attribute when generating your collections?

finestructure:
Fixes #1169 

Open questions:

- [x] is it ok to include `StringMangling.swift`? (Adding TSC as a dependency would force us to include sqlite dev dependency into Linux base images)
- [x] do we need to careful which target names we supply or do we need to determine module "importability" somehow? If so, how?

cc @neonacho @yim-lee

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -22,10 +22,7 @@ let package = Package(
         .package(url: "https://github.com/scinfu/SwiftSoup.git", from: "2.3.2"),
         .package(name: "SwiftPM",
                  url: "https://github.com/apple/swift-package-manager.git",
-                 .revision("swift-DEVELOPMENT-SNAPSHOT-2021-05-04-a")),
-        .package(name: "swift-tools-support-core",
-                 url: "https://github.com/apple/swift-tools-support-core.git", from:"0.2.2")
-
+                 .revision("swift-DEVELOPMENT-SNAPSHOT-2021-05-04-a"))
     ],
     targets: [
         .target(name: "App", dependencies: [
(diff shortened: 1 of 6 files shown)


yim-lee:
> is it ok to include StringMangling.swift?

TSC is Apache 2.0 so as long as you follow the license term it should be ok.

> to include sqlite dev dependency into Linux base images

I might have asked before but what are the concerns with adding sqlite dependency?

> do we need to careful which target names we supply or do we need to determine module "importability" somehow?

You'd want targets that are in at least one product:
https://github.com/apple/swift-package-collection-generator/blob/main/Sources/PackageCollectionGenerator/PackageCollectionGenerate.swift#L313-L315

Starting from Swift 5.4 `swift package describe` should provide the info we need (e.g., `C99name` and `Product memberships`) so in the future swift-package-collection-generator will switch to that and remove the logic from its code.

finestructure:
> > is it ok to include StringMangling.swift?
> 
> TSC is Apache 2.0 so as long as you follow the license term it should be ok.
> 
> > to include sqlite dev dependency into Linux base images
> 
> I might have asked before but what are the concerns with adding sqlite dependency?
> 
> > do we need to careful which target names we supply or do we need to determine module "importability" somehow?
> 
> You'd want targets that are in at least one product:
> https://github.com/apple/swift-package-collection-generator/blob/main/Sources/PackageCollectionGenerator/PackageCollectionGenerate.swift#L313-L315
> 
> Starting from Swift 5.4 `swift package describe` should provide the info we need (e.g., `C99name` and `Product memberships`) so in the future swift-package-collection-generator will switch to that and remove the logic from its code.

PR #1174, opened 2021-06-14
Add visual warning when the repository is archived

ricardopereira:
In the package listing, in my opinion, it should visually warn the user when the repository has been archived by the owner. 

**Example:**

Before:
Screenshot 2021-06-14 at 15 50 21

After (add a warning symbol):
Screenshot 2021-06-14 at 15 48 00

### DIFF
--- a/Sources/App/Views/PackageController/PackageShow+Model.swift
+++ b/Sources/App/Views/PackageController/PackageShow+Model.swift
@@ -213,7 +213,7 @@ extension PackageShow.Model {
         var releasesSentenceFragments: [Node<HTML.BodyContext>] = []
         if isArchived {
             releasesSentenceFragments.append(contentsOf: [
-                .strong("No longer in active development."),
+                .strong("⚠️ No longer in active development."),
                 " The package author has archived this project and the repository is read-only. It had ",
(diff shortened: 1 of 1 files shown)


daveverwer (changes requested):
Hey Ricardo, and thanks very much for this.

I think if we need to make this more obvious, I'd want to pull that text out into a separate line and use the icon that we have on the left-hand side of the list items to indicate it's important.

I'm tracking that now as #1175.

If you want to have a go at doing that, that'd be great. I can make the icon for you and add it once you have the text re-positioned. Let me know if you'd like to do that.

ricardopereira:
Thanks @daveverwer, I'll try do that. Something like this?

Screenshot 2021-06-15 at 10 47 46

daveverwer:
Yep, that looks great! I might tweak the icon to be different but I'd like to run it locally before deciding that. Could you push those changes to this branch please and I'll check it out.

Thanks!

daveverwer:
Thanks again for this @ricardopereira. I'm going to close this PR now as it's superseded by #1221 from @selichter!

Thanks to you both.

PR #1176, opened 2021-06-15
Rss releases notes

elfenlaid:
Closing #421

That's how [NetNewsWire.app](https://netnewswire.com/) renders the changes in `releases.rss` file:

<details>
<summary>Version with release notes</summary>
Screenshot 2021-06-15 at 12 19 58
</details>

<details>
<summary>Version without release notes (the old format)</summary>
Screenshot 2021-06-15 at 12 53 11
</details>

By the way, turns out Github's GraphQL API supports HTML description out of the box. So, for the RSS case, it seems more fitting than [an additional REST API request](https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/issues/421#issuecomment-859423277)

The change is backward compatible with the current database. It uses the same logic as the absent release notes – `releases.rss` renders an empty `content` field. 

Ideally, we should drop `content` for empty notes. But, alas, I failed to express how to do that with Plot's RSSFeed items, at least for now.
Also, it seems that RSS Readers use descriptions field info in this case.


Sources/App/Models/Release.swift
```diff
@@ -12,7 +12,7 @@ struct Release: Codable, Equatable {
 
 extension Release {
     init(from node: Github.Metadata.ReleaseNodes.ReleaseNode) {
-        description = node.description
+        description = node.descriptionHTML
```
finestructure, line 15: So this effectively changes `Version.releaseNotes` from containing plain text to html (no schema change obv), because it gets initialised from `Release.description`.

There is one other place where `version.releaseNotes` is currently being used: `PackageCollection.Package.Version.init?` where it gets assigned to `summary`.

We need to check if it's ok to change this from plain text to html. This might be exposed in Xcode and we wouldn't want to make it show HTML markup.

finestructure:
This looks great!

The only thing I believe we need to check is the release notes format issue in package collections.

Sources/App/Models/Release.swift
elfenlaid: I've completely missed it 😨 Thanks for the heads up

daveverwer (approved):
Aside from Sven's comment, in terms of how this works, this looks great! Thanks so much, Egor!

Sources/App/Models/Release.swift
elfenlaid: Instead of replacing the `releaseNotes` content, I've introduced the `releaseNotesHTML` field. So, `PackageCollection` uses plain ol' markdown `releaseNotes`, and `RSSFeed` was switched to `releaseNotesHTML`

finestructure:
Sorry for not replying sooner, Egor! I took a brief look this morning but then got sucked into some tricky SQL stuff and completely forgot to return to it 😅

I'll come back to this PR and merge first thing in the morning!

Sources/App/Core/RSS.swift
```diff
@@ -102,14 +94,33 @@ extension RecentRelease {
...
+                    .p(.text(packageSummary ?? ""))
+                )
+            ),
+            .content(
+                .unwrap(releaseNotesHTML) { notes in
+                    layout(
+                        .div(.raw(notes))
+                    )
+                }
+            )
```
finestructure, line 123: I noticed that all our snapshots come with 

```
<content:encoded><![CDATA[]]></content:encoded>
```

I guess because we don't have html release notes set - which we should do for at least one of them.

How does an empty `content:encoded` manifest itself in RSS readers? Will this just be ignored?

Or should we move the `unwrap` up one level?

```swift
             .unwrap(releaseNotesHTML) { notes in
                 .content(
                     layout(
                         .div(.raw(notes))
                     )
                 }
             )
```

such that the whole block is omitted if `releaseNotesHTML` is `nil`?

Tests/AppTests/RSSTests.swift
```diff
@@ -77,6 +77,7 @@ class RSSTests: SnapshotTestCase {
                         commitDate: Date(timeIntervalSince1970: TimeInterval($0)),
                         packageName: "pkg-\($0)",
                         reference: .tag(.init($0, 0, 0), "\($0).0.0"),
+                        releaseNotes: "<p>Awesome Release Notes</p>",
```
finestructure, line 80: Ah, that's why `content:encoded` is blank - this should be `releaseNotesHTML:`

Sources/App/Core/RSS.swift
finestructure: I've gone ahead and made that change. Let me know that makes sense and I'll merge!

finestructure (approved):

Sources/App/Core/RSS.swift
elfenlaid: > How does an empty content:encoded manifest itself in RSS readers?

They ignore it and use description instead, but lets be safe

> I've gone ahead and made that change. Let me know that makes sense and I'll merge!

Many thanks, that's what I've been looking for! The top level `unwrap` 👌

elfenlaid:
> Sorry for not replying sooner, Egor! I took a brief look this morning but then got sucked into some tricky SQL stuff and completely forgot to return to it 😅
> 
> I'll come back to this PR and merge first thing in the morning!

I'm flattered, but if deprioritizing my PRs would make your day easier, do that! Heck, weeks of waiting won't irritate me :)
Everybody's is different, of course. But for me, a contribution is a privilege. It costs maintainers' time and efforts. So, please don't mind any delays and thanks for your hard work 🙏

elfenlaid:
Eh, rebasing the branch to the latest main...

finestructure:
Thanks again for tackling this. It'll make my RSS feed so much more useful! :)

Will deploy as 2.39.0 🚀

daveverwer:
This is a great step forward Egor! Thanks so much!

Can't wait to see the new feeds coming through, I'm about to do my weekly trawl through my RSS, so I'll be reading!

daveverwer:
Look how glorious it is 😍

Screenshot 2021-06-17 at 21 49 20@2x

finestructure:
It's great! :)

elfenlaid:
Btw, have you ever considered generating a code diff and attach it to the changelogs 🤔 For example, like [HexDiff](https://diff.hex.pm/diff/phoenix/1.5.8..1.5.9) does (Hex is Elixir's package manager.)

Screenshot 2021-06-18 at 08 06 34

Screenshot 2021-06-18 at 08 06 44

PR #1177, opened 2021-06-15
Slightly refactor current search and assert actual search SQL

finestructure:
This is just a small search refactor and an additional assert on the actual search SQL, in preparation of changing it for author and topic search.

Pulling this forward to make the actual functional changes more obvious in upcoming PRs.

### DIFF
--- a/Sources/App/Core/Search.swift
+++ b/Sources/App/Core/Search.swift
@@ -76,11 +76,9 @@ enum Search {
     static func query(_ database: Database,
                       _ terms: [String],
                       page: Int,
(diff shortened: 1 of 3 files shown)


Sources/App/Core/Search.swift
```diff
@@ -76,11 +76,9 @@ enum Search {
     static func query(_ database: Database,
```
finestructure, line 76: `query` used to create the actual query object and run the fetch. It's now to separate functions `query` and `fetch` so that we can render and test the generated SQL from `query`. Just a split and rename.

daveverwer (approved):

PR #1178, opened 2021-06-16
Test SQL for multiple search terms

finestructure:
Just another test for our search query. I'm pulling these forward because I need to make significant changes to the search query and I want to keep the PRs as small as possible.

I'll expand on the req changes in the next PR.


finestructure:
Closing this. There's more to come that can go together.

PR #1180, opened 2021-06-18
Refactor search

finestructure:
This should be a non-functional change, preparing for topic search

### DIFF
--- a/Sources/App/Core/Search.swift
+++ b/Sources/App/Core/Search.swift
@@ -137,7 +137,7 @@ enum Search {
         }
 
         // page is one-based, clamp it to 0-based offset
-        let offset = max(0, (page - 1) * pageSize)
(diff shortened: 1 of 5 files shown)


daveverwer (approved):

Sources/App/Core/Search.swift
```diff
@@ -124,7 +124,34 @@ enum Search {
...
+                                            offset: offset,
+                                            limit: limit) else {
+            return nil
+        }
+
+        return db.select()
+            .column("*")
+            .from(
+                SQLAlias(SQLGroupExpression(inner), as: SQLIdentifier("t"))
+            )
```
finestructure, line 154: This looks a bit silly now with just an empty select wrapper but I wasn't sure how nesting works with SQLKit and we need this as a wrapper around the `UNION` selects.

finestructure:
We need to test this a bit on staging before deploying (on Monday).

PR #1181, opened 2021-06-18
Populate topics

finestructure:
- adds `repositories.topics` to db schema
- extends Github GraphQL query to fetch topics
- populates `repositories.topics` from GH metadata

I've got these changes as three separate branches in case we need to break it down further.

I'll deploy this from the branch to get test data into the staging db and we can then decide if we want to merge or roll back while I test topic search.

### DIFF
--- a/Sources/App/Core/Github.swift
+++ b/Sources/App/Core/Github.swift
@@ -267,6 +267,14 @@ extension Github {
                         name
                       }
                     }
+                    repositoryTopics(first: 20) {
+                      totalCount
+                      nodes {
+                        topic {
+                          name
+                        }
+                      }
+                    }
                     releases(first: 20, orderBy: {field: CREATED_AT, direction: DESC}) {
                       nodes {
                         description
--- a/Sources/App/Core/Github.swift
+++ b/Sources/App/Core/Github.swift
@@ -267,14 +267,6 @@ extension Github {
                         name
                       }
                     }
-                    repositoryTopics(first: 20) {
-                      totalCount
-                      nodes {
-                        topic {
-                          name
-                        }
-                      }
-                    }
                     releases(first: 20, orderBy: {field: CREATED_AT, direction: DESC}) {
                       nodes {
                         description
@@ -283,6 +275,14 @@ extension Github {
                         publishedAt
                         tagName
                         url
+                      }
+                    }
+                    repositoryTopics(first: 20) {
+                      totalCount
+                      nodes {
+                        topic {
+                          name
+                        }
                       }
                     }
                     stargazerCount
--- a/Sources/App/Core/Github.swift
+++ b/Sources/App/Core/Github.swift
@@ -308,6 +308,7 @@ extension Github {
             var openPullRequests: OpenPullRequests
             var owner: Owner
             var releases: ReleaseNodes
+            var repositoryTopics: RepositoryTopicNodes
             var stargazerCount: Int
             var isInOrganization: Bool
             // derived properties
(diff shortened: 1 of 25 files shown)


Sources/App/Core/Github.swift
```diff
@@ -277,6 +277,14 @@ extension Github {
                         url
                       }
                     }
+                    repositoryTopics(first: 20) {
```
finestructure, line 280: NB: capping list of topics at 20

Sources/App/Models/Repository.swift
```diff
@@ -91,26 +94,29 @@ final class Repository: Model, Content {
...
          package: Package,
          authors: [Author] = [],
-         summary: String? = nil,
          commitCount: Int? = nil,
+         defaultBranch: String? = nil,
          firstCommitDate: Date? = nil,
+         forks: Int? = nil,
+         forkedFrom: Repository? = nil,
+         isArchived: Bool? = nil,
+         isInOrganization: Bool? = nil,
```
finestructure, line 116: There's quite a bit of commit noise here but I finally twitched and had to sort this long list of parameters.

That's one of the reasons I created a number of branches so we can make this PR smaller.

finestructure:
I think this is good to go now. I've tested the topic landing page query on staging and am getting fast results, even without an index (4k rows is not a lot after all):

```sql
explain analyze
select distinct r.name, topics
from repositories r, unnest(topics) t
where t like '%test%'
```

```
HashAggregate  (cost=5469.30..5506.49 rows=3719 width=81) (actual time=8.072..8.102 rows=67 loops=1)
  Group Key: r.name, r.topics
  ->  Nested Loop  (cost=0.00..5450.24 rows=3812 width=81) (actual time=0.773..7.970 rows=111 loops=1)
        ->  Seq Scan on repositories r  (cost=0.00..647.12 rows=3812 width=81) (actual time=0.021..2.355 rows=3812 loops=1)
        ->  Function Scan on unnest t  (cost=0.00..1.25 rows=1 width=0) (actual time=0.001..0.001 rows=0 loops=3812)
              Filter: (t ~~ '%test%'::text)
              Rows Removed by Filter: 4
Planning Time: 29.417 ms
Execution Time: 8.368 ms
```

daveverwer:
One thing I wanted to raise here is whether storing the topics in this way will make building on this data more challenging. I had pictured storage of topic data as a more traditional database relationship with a many<>many join table between a list of repositories/packages and a list of topics.

I'm happy to go with whatever works best, but we will want to pull out a full list of topics for a `/topics` page that shows all topics, and then also list packages by topic so we can make pages like `/topics/networking`, maybe even RSS feeds or similar in the future.

All of this may be possible to keep performant with this data storage, but I wanted to raise it before we set it in stone.

daveverwer:
Haha I think you answered my question as I was typing that up 😂

finestructure:
Yes, that's why I didn't want to merge until I've tested that :)

Also:

```
explain analyze
select distinct t
from repositories r, unnest(topics) t
```

```
HashAggregate  (cost=9224.12..9225.12 rows=100 width=32) (actual time=11.405..11.799 rows=4226 loops=1)
  Group Key: t.t
  ->  Nested Loop  (cost=0.00..8271.12 rows=381200 width=32) (actual time=0.074..8.323 rows=13828 loops=1)
        ->  Seq Scan on repositories r  (cost=0.00..647.12 rows=3812 width=69) (actual time=0.029..2.429 rows=3812 loops=1)
        ->  Function Scan on unnest t  (cost=0.00..1.00 rows=100 width=32) (actual time=0.001..0.001 rows=4 loops=3812)
Planning Time: 0.255 ms
Execution Time: 12.110 ms
```

So both queries are fast, even without any additional indexes.

daveverwer:
The only other thing I'd raise here is the naming of the keyword data inside our system as `topics`. I know that's how GitHub refers to it, but I'm not sure it's the best name.

I was thinking about this in terms of `keywords` rather than topics. I know the field name doesn't preclude referring to it in different ways, but I think when we do come to displaying this data on the site I'd want to refer to them as keywords rather than topics, and I think the URL structure would fit better as `/keywords/networking` than `/topics/networking`.

What do you think?

finestructure:
Ah yes, that's better.

daveverwer (approved):

finestructure:
staging db rolled back to `main`'s schema and re-deployed so we can apply this merge once I've updated the field.

PR #1183, opened 2021-06-21
Unique, lowercased, sorted topics

finestructure:
Normalise topics so we can search via `'foo' = any(topics)` rather than `ilike`. This would mean we discard any cases for topic display. At a glance, topics are already lowercased and uniqued on the Github side. This is just ensuring that on our end so we can assume it for queries.

Does that make sense, @daveverwer ?

This is just a PR into the topics branch, so ok to merge. I'll then deploy from the branch again and grab another snapshot when it's ingested.


finestructure:
Hold on, this may not be necessary.

finestructure:
It does help.

PR #1185, opened 2021-06-21
Add Sitelinks Search Box

### LINKED ISSUE #1129: Add support for Google Sitelinks Search Box
sitelinks01

https://developers.google.com/search/docs/data-types/sitelinks-searchbox

Sherlouk:
https://developers.google.com/search/docs/data-types/sitelinks-searchbox

Closes #1129

### DIFF
--- a/Sources/App/Core/Extensions/Array+queryString.swift
+++ b/Sources/App/Core/Extensions/Array+queryString.swift
@@ -1,8 +1,10 @@
 
 extension Array where Element == QueryParameter {
-    func queryString(includeSeparator: Bool = true) -> String {
+    func queryString(includeSeparator: Bool = true, encoded: Bool = true) -> String {
         guard !isEmpty else { return "" }
-        let query: String = self.map { $0.encodedForQueryString }.joined(separator: "&")
+        let query: String = self
+            .map { encoded ? $0.encodedForQueryString : $0.unencodedQueryString }
+            .joined(separator: "&")
         return includeSeparator ? "?" + query : query
     }
 }
--- a/Sources/App/Views/Home/HomeIndex+StructuredDataSchema.swift
+++ b/Sources/App/Views/Home/HomeIndex+StructuredDataSchema.swift
@@ -20,7 +20,7 @@ extension HomeIndex {
                 target: .init(
                     urlTemplate: SiteURL.search.absoluteURL(parameters: [
                         .init(key: "query", value: "{search_term_string}")
-                    ])
+                    ], encodeParameters: false)
                 ),
                 queryInput: "required name=search_term_string"
             )
--- a/Sources/App/Views/Home/HomeIndex+StructuredDataSchema.swift
+++ b/Sources/App/Views/Home/HomeIndex+StructuredDataSchema.swift
@@ -1,0 +1,63 @@
+import Foundation
+
+extension HomeIndex {
+    
+    struct IndexSchema: Encodable {
+        enum CodingKeys: String, CodingKey {
+            case context = "@context", type = "@type"
+            case url, potentialAction
+        }
+        
+        var context: String = "https://schema.org"
+        var type: String = "WebSite"
+        
+        let url: String
+        let potentialAction: SearchActionSchema
+        
+        init() {
+            self.url = SiteURL.home.absoluteURL()
+            self.potentialAction = .init(
+                target: .init(
+                    urlTemplate: SiteURL.search.absoluteURL(parameters: [
+                        .init(key: "query", value: "{search_term_string}")
+                    ])
+                ),
+                queryInput: "required name=search_term_string"
+            )
+        }
+    }
+    
+    struct SearchActionSchema: Encodable {
+        enum CodingKeys: String, CodingKey {
+            case context = "@context", type = "@type"
+            case target, queryInput = "query-input"
+        }
+        
+        var context: String = "https://schema.org"
+        var type: String = "SearchAction"
+        
+        let target: EntryPointSchema
+        let queryInput: String
+        
+        init(target: EntryPointSchema, queryInput: String) {
+            self.target = target
+            self.queryInput = queryInput
+        }
+    }
+    
+    struct EntryPointSchema: Encodable {
+        enum CodingKeys: String, CodingKey {
+            case context = "@context", type = "@type"
+            case urlTemplate
+        }
+        
+        var context: String = "https://schema.org"
+        var type: String = "EntryPoint"
+        
+        let urlTemplate: String
+        
+        init(urlTemplate: String) {
+            self.urlTemplate = urlTemplate
+        }
+    }
+}
--- a/Sources/App/Views/Home/HomeIndex+View.swift
+++ b/Sources/App/Views/Home/HomeIndex+View.swift
@@ -41,6 +41,7 @@ enum HomeIndex {
                     .div(
                         .class("inner"),
                         .h3("The place to find Swift packages."),
+                        .structuredData(IndexSchema()),
                         .searchForm(),
                         .unwrap(model.statsClause()) { $0 }
                     )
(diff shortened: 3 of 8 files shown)


Sources/App/Views/Home/HomeIndex+StructuredDataSchema.swift
```diff
@@ -0,0 +1,63 @@
...
+        
+        let url: String
+        let potentialAction: SearchActionSchema
+        
+        init() {
+            self.url = SiteURL.home.absoluteURL()
+            self.potentialAction = .init(
+                target: .init(
+                    urlTemplate: SiteURL.search.absoluteURL(parameters: [
+                        .init(key: "query", value: "{search_term_string}")
```
Sherlouk, line 22: One thing I'm not 100% sure on is the fact that this parameters function will encode the `{` and `}` values - this is not what Google's example shows. It's plausible that our encoding here will actually prevent the whole feature from working.

We may want to try something more specific such as:
```swift
SiteURL.search.absoluteURL() + "?search={search_term_string}"
```
Or we could pass a new option through to `disableEncoding: true` or something which skips that step? This might be a bit nicer

Sources/App/Views/Home/HomeIndex+View.swift
```diff
@@ -41,6 +41,7 @@ enum HomeIndex {
                     .div(
                         .class("inner"),
                         .h3("The place to find Swift packages."),
+                        .structuredData(IndexSchema()),
```
Sherlouk, line 44: Not sold on the location of this, open to suggestions. The structured data is technically specific to the homepage, but it's only relevant to the search box (currently) which is why I put it here.

Could just as easily make sense right at the top.

daveverwer, line 44: Let's keep this in the `bodyComments` for now. That method still needs a rename, but it's a better place for this data.

In an ideal world, I think this would fit best just *before* the `</body>` tag. Maybe a method named `bodyPreClosing` (ugh that's an awful name), `bodyFooter` (ugh, makes it seem like it's visual), or something similar (suggestions welcome)

Maybe just `andFinally...` 😂

daveverwer: hah sorry I was also commenting on this as you were typing yours up. Closing this conversation in favour of the one below

Sources/App/Views/Home/HomeIndex+StructuredDataSchema.swift
daveverwer: Is there a way for us to test this? I'm not sure there is. We block Google from indexing staging.

It's not the end of the world if we have to test it on prod, it's not like it'll break anything on the site itself.

Sources/App/Views/Home/HomeIndex+View.swift
Sherlouk: Didn't see it when I added it, only just realised it's an optional override from parent class. Moved 🙂

daveverwer: I moved the other one for the package data there too, but I just pushed it up in a commit so it wasn't super obvious. Thanks for moving it!

How about just `preBody` and `postBody` to replace `bodyComments` and add the one at the end?

daveverwer: Probably a separate PR though, no need to corrupt this one with that change. I'm just still thinking about it 😂

Sources/App/Views/Home/HomeIndex+StructuredDataSchema.swift
Sherlouk: The other issue is we don't know how long it takes for Google to reindex and enable this feature even if it were to work 😅

daveverwer: Can we infer it from looking at other sites that implement this? 🤔

Sources/App/Views/Home/HomeIndex+View.swift
Sherlouk: Aye happy to add these new functions as part of a separate PR and apply it to both schemas. Will create `preBody` and `postBody` to go at the top and bottom respectively.

Sources/App/Views/Home/HomeIndex+StructuredDataSchema.swift
Sherlouk: If we're to infer, I feel like we have to follow Google's own documentation which has it unencoded. Would you like me to go down the `disableEncoding` idea and get us as close to the docs as possible?

daveverwer: I tried looking for another site that supported it and found https://www.asos.com/ does.

However, it does not appear to implement the `SearchAction` structured data. I wonder if they're doing it a different way?

Sherlouk: Google definitely have some intelligence they use to capture most website's search features without explicit support for the structured data (either form) - generally for a lot of big websites that magically fit their algorithm!

daveverwer: Let's go as close to the docs as possible then 🚀

Sources/App/Views/Home/HomeIndex+View.swift
Sherlouk: On second thoughts, got bored - have done it as part of this PR 😛  (sorry)

Sources/App/Views/Home/HomeIndex+StructuredDataSchema.swift
Sherlouk: Have implemented my solution, let me know if you'd rather tackle it in a different way.

Sources/App/Core/Extensions/Array+queryString.swift
```diff
@@ -1,8 +1,10 @@
 
 extension Array where Element == QueryParameter {
-    func queryString(includeSeparator: Bool = true) -> String {
+    func queryString(includeSeparator: Bool = true, encoded: Bool = true) -> String {
         guard !isEmpty else { return "" }
-        let query: String = self.map { $0.encodedForQueryString }.joined(separator: "&")
+        let query: String = self
+            .map { encoded ? $0.encodedForQueryString : $0.unencodedQueryString }
```
finestructure, line 6: Nit: could perhaps call this just `$0.queryString` or push the `encoded` flag into `QueryParameter.queryString(encoded: Bool = false)`.

Sources/App/Views/PublicPage.swift
```diff
@@ -275,6 +277,18 @@ class PublicPage {
         .empty
     }
     
+    /// Optional content that will be inserted just after the start of the body.
+    /// - Returns: An optional element, or group of elements.
+    func preBody() -> Node<HTML.BodyContext> {
```
finestructure, line 282: If I hadn't seen the naming discussing I'd think that either the name or the doc string got out of sync 🤪

Some alternatives that might work perhaps: `bodyOpening|Closing`, `opening|closingBody`, `leading|trailingBody`, `bodyLeading|Trailing`.

finestructure (approved):
Just a couple of nits which we could address in a follow up if considered worthwhile.

Sources/App/Views/PublicPage.swift
Sherlouk: I hate naming things 😂

finestructure: There's always caching if you need a break 🤣

daveverwer: Me too, and this is a tricky one.

I don't hate `preBody` and `postBody`. They are not pre/post the body tag, but they are pre/post the body itself.

PR #1186, opened 2021-06-21
Add keywords (topics) to package schema

Sherlouk:
We do not currently show this data anywhere else in the UI, so this would be a first (?) - not sure if that equates to anything but thought if we do have it then we might as well use it somewhere and this was simple enough!


daveverwer (approved):

daveverwer:
Yes, this will be the first place we'll use this data! 🚀

It's actually the only real-world way we have to test it outside of a database query right now 😂

daveverwer:
Let's test this on staging, it'll deploy within the next 10 mins to there.

daveverwer:
From staging

Screenshot 2021-06-21 at 22 56 16@2x

Sherlouk:
0 errors. 0 warnings.

That's what I like to see 😄  That's awesome! Thanks for sharing the example Dave

PR #1187, opened 2021-06-22
Update badge to include logo and brand colour

### LINKED ISSUE #993: Add the SPI logo and colours to the shields.io badges
`?logo=data:image/png;base64,…`

> Insert custom logo image (≥ 14px high). There is a limit on the total size of request headers we can accept (8192 bytes). From a practical perspective, this means the base64-encoded image text is limited to somewhere slightly under 8192 bytes depending on the rest of the request header.

from https://shields.io

Sherlouk:
Closes #993


Sources/App/Controllers/API/Package+Badge.swift
```diff
@@ -67,15 +69,18 @@ extension Package {
...
 
+        let message = badgeMessage(badgeType: badgeType)
         return Badge(schemaVersion: 1,
                      label: label,
-                     message: badgeMessage(badgeType: badgeType),
-                     color: "blue",
-                     cacheSeconds: cacheSeconds)
+                     message: message ?? "unavailable",
+                     isError: message == nil,
+                     color: message == nil ? "inactive" : "F05138",
```
Sherlouk, line 77: Shows the brand colour only if it's all working fine. If we're in "unavailable" mode for whatever reason, it's set to the inactive grey colour.

isError purely means that the colour can't be overridden which means it's always the inactive grey colour when we can't load the details. They can override the brand colour should they choose to do so.

This logic wasn't on the original issue but I hope you agree it makes sense given the context!

Sherlouk:
Worth saying, I haven't been able to test this but if someone has NGROK setup and wants to give it a stab then feel free! The contents looks correct as per their docs. Worst case we throw it on staging and take a look.

daveverwer:
I just manually deployed this branch to staging so it's there for testing now. I just generated a couple of images and they look like this:

Screenshot 2021-06-22 at 16 43 29@2x
Screenshot 2021-06-22 at 16 43 48@2x

That logo is super small. I think we might want to make a new version that's even simpler. Maybe even just the Swift logo on its own. I'm not sure I can make the boxes look good at that size.

Thoughts?

finestructure:
I think that looks great. It’s still recognisable SPI, which we’d lose if it’s just the Swift logo. Might suffer a bit on non-retina screens but hey :)

daveverwer:
OK.

Well, I think I can improve it from what we have here 😂 I'll experiment with it tomorrow.

Sherlouk:
I'm with Sven - they look great, yes it's small but it's still recognisable. I don't think the Swift logo on it's own adds any value, it's either the SPI brand or (in my opinion) nothing at all.

Thanks for sharing the images though!

daveverwer:
I think this one from the other branch is looking better.

Screenshot 2021-06-22 at 21 00 33@2x

However, this is a real problem on Safari:

Screenshot 2021-06-22 at 21 01 35@2x

Screenshot 2021-06-22 at 21 01 54@2x

Sherlouk:
We have a lens flare on the SPI logo!? 😂  That looks ridiculous... and very frustrating.

finestructure:
Is it an option to render it out into a PNG?

It seems the issue is the aliasing with the alpha channel - seeing how the Swift logo is sharp. Perhaps we can provide variants without alpha that match the badge colours?

Also, I liked the initial version with the two background squares a little better but I think they all work well and even with the messed up aliasing it looks ok in Safari IMO. I checked the latest Safari Tech Preview (yesterday's 126) and unfortunately there's no help coming there...

finestructure:
Or add a little halo, like the circle masking in Tweetbot:

CleanShot 2021-06-23 at 07 20 23@2x

Although that might look quite busy on such a small icon 🤷

daveverwer (approved):
Thanks for this James!

Sherlouk:
My PRs are never simple 'eh 😂  Sorry Dave, but thanks for the help!

daveverwer:
No no, it's all me. I can't leave the front end of anything alone 😂

PR #1189, opened 2021-06-22
Tiny badge icon

daveverwer:
Trying a new tiny badge layout. 😅

Also normalised the naming and location of logo assets. @Sherlouk The SVGs folder in `Resources` shouldn't really even be deployed, it's only there to generate the image data in `images.scss`. Anything that's rendered publicly should go in `Public`.


Sources/App/Controllers/API/Package+Badge.swift
```diff
@@ -102,7 +102,7 @@ extension Package {
     
     static private func loadSVGLogo() -> String? {
         let pathToFile = Current.fileManager.workingDirectory()
-            .appending("Resources/SVGs/logo-simple.svg")
+            .appending("Public/images/logo-tiny.svg")
```
Sherlouk, line 105: Looks like we use `SiteURL.images("logo-small.png")` for the favicon - I wonder if the same syntax could be used here?

Sherlouk:
> The SVGs folder in `Resources` shouldn't really even be deployed

Aye - I checked the Dockerfile and saw they were included so thought it fair game, especially as other logos were present there. I understand the desire to keep them in Public/ though 👍

Sources/App/Controllers/API/Package+Badge.swift
daveverwer: SiteURL.images is going to generate a relative (or absolute, I guess!) URL, and don't we need a path to the file on disk here?

Sherlouk: You're entirely correct, yes. Ignore me 😄

daveverwer:
I wonder if there's caching on the shields.io site? I just deployed this version to staging to test it but I'm still getting the old logo.

Sherlouk:
We cache for 6 hours 🙂  Should be able to try a different package though?

daveverwer:
OK, I'm a bit happier with these. I went back to the larger number of boxes but with a bigger swift bird in the middle. Looks as good as it's going to in that space.

Safari:

Screenshot 2021-06-23 at 12 57 36@2x
Screenshot 2021-06-23 at 12 57 44@2x

Chrome:

Screenshot 2021-06-23 at 12 57 54@2x
Screenshot 2021-06-23 at 12 57 58@2x

daveverwer:
I'll merging this into your PR @Sherlouk and then I'll merge that one

PR #1190, opened 2021-06-23
Add topic search

### LINKED ISSUE #1179: Changes to Search (for topics, authors, ...)
# Changes to Search

With our upcoming changes to support topics, we need to update our search in order to be able to surface these results. Similar changes are required to support author and custom package collection searches.

A lot of what I'm writing below I've actually implemented or prototyped so I could get a feel for if/how it would work. All of this is related in certain ways and even though we can adopt these changes incrementally, deciding on how to add topics and how to query them sets us on a given path.

## Current Search

Search currently works as follows:

- we take a list of search terms `[String]` and sanitise them
- we then concatenate a set of fields we want to search to make up our "haystack": package name, summary, repository name, repository owner
- for each search term, we add a clause `where(haystack, contains, needle)`
- there's an `ORDER BY` clause specifically designed to bring exact package matches to the front

The result set is a `DBRecord` with fields `packageId`, `packageName`, `repositoryName`, `repositoryOwner`, `summary`.

This is turned into a `Search.Result` by passing these values 1:1 and also adding a `packageURL` which is the relative `SiteURL` to the result.

## Topic Search

When adding topic searches (and later author and collection searches), we need to be able to distinguish the result type based on what was matched in order to build the resulting URL we link to.

One way of doing that (the simplest I could come up with after a few attempts) would be the following:

- Create individual queries to find topics, authors, custom collections.
- `UNION` them together in the order in which we want to display results in search (for instance topics > authors > collections > packages)
	- Mostly this order should only affect what comes before general packages, because it's unlikely (but not impossible) that something matches both as a topic and as an author, for example.
- Each individual query has an extra field `match_type` which indicates what type or result it is, i.e. `topic`, `author`, `collection`. This will allow the result processor to create the correct link for each result. It will also allow us to add icons for the type of result to the search page.

This is a significant change to one of our most central features, so we need to be careful when revamping it, both from a correctness but also performance point of view.

Before adding topics, we should move the current package search over to this `match_type` query.

Because we're concatenating individual queries and not changing the package query fundamentally we should be pretty safe from regressions and be able to add new types of searches incrementally.

## Example queries

Here's an example of our current (as of 2.39.0 and ealier) search for `["vapor", "sql"]`:

```sql
SELECT "id", "package_name", "name", "owner", "summary" 
FROM "search" 
WHERE 
concat("package_name", ' ', coalesce("summary", ''), ' ', "name", ' ', "owner") ~* 'vapor'
AND concat("package_name", ' ', coalesce("summary", ''), ' ', "name", ' ', "owner") ~* 'sql'
AND "package_name" IS NOT NULL 
AND "owner" IS NOT NULL 
AND "name" IS NOT NULL 
ORDER BY lower("package_name") = 'vapor sql' DESC, "score" DESC, "package_name" ASC 
LIMIT 21 OFFSET 0
```

Initially, we'll prepend `'package' as match_type` as a new static column to prepare for new search.

Next, we `UNION` this query with a topic search:

```sql
(
	SELECT 'package' as match_type ... (rest see above)
)
UNION
(
	select 'topic', null, null, null, null, null, null
	from search
	where 'vapor sql' ilike any(topics)  -- TODO: check if ilike works here
	limit 1
)
```

`topics` is a new `TEXT[]` column on `repositories` which we can efficiently query as shown here. It supports indexing. https://www.postgresonline.com/journal/archives/228-PostgreSQL-Array-The-ANY-and-Contains-trick.html

Note that I've not expanded topic search into multiple term clauses like for packages. If we did that, like for instance

```sql
	where 'vapor' ilike any(topics)
        or 'sql' ilike any(topics)
```

then we wouldn't know which one actually matched. You wouldn't be able to search for multiple topics but I don't think that's something we need to worry about. (I should add that in general our search is `AND` with respect to its terms, that's why the `concat` clauses are `AND`ed and not `OR`ed`. So this just follows that same logic.)

## Landing page queries

Our current search results do not require any "landing page" queries, because they point at packages which we resolve via their `id`. This will be different for topics (but not for authors and collections, where we have a unique identifier - a name - similiar to the package `id`).

It's therefore important that we choose a representation for topics in the schema that is efficient not only for search where we stop after the first hit. For the landing page we need to find _all_ packages with a certain topic.

Our two options are:
- package/repository-topic many-to-many table
- a `topics text[]`  field (on repositories)

Pros/cons:
- inline field is an easier change, no new model class, relationships to set up
- with a relationship we need to consider orphans or delete cascading when deleting packages (which is rare but needs to be considered)
- query performance should be straightforward for a many-to-many relationship
- an inline field probably requires an index to allow fast topic filtering
- because topics aren't a curated list but rather a user configurable set with lots of spelling variations it doesn't _feel_ like something we should model as a relationship
- ingesting a new set of topics requires managing the existing relationships (probably by deleting and re-adding then) - not a huge deal but this is dead simple in case of an inline field

### Example queries

Many-to-many table:
```sql
select * from packages p
join repositories r on ... 
join topics_repositories tr on ...
where 
  tr.name ilike 'networking'
```

Inline field:
```sql
select * from packages p
join repositories r on ... 
where 'networking' ilike any(r.topics)
```

## Next Steps

I've got most of the pieces in place to

- prepare Search for topics/authors/etc
- add `repositories.topics TEXT[]`
- populate it from Github
- update Search to support topics

I haven't tested the topics landing page query yet because I can't easily ingest everything locally (not enough space on the SSD! - please like and subscribe 😁).

My plan is to deploy from a branch and test this on staging. This will introduce a schema change and make other deployments a bit tricky while I'm working on it. But if we deploy it and let it ingest, I can pull a copy from staging and then we can roll back to `main`. It's a good opportunity to practise schema rollbacks, too.

finestructure:
⚠️ Schema change ⚠️

This adds initial support for keyword results to search.

Performance is fine (tested locally against copy of staging with a good matcher `test`): 9ms.

```sql
-- test_query_sql
explain analyze
SELECT * FROM ((SELECT 'keyword' AS "match_type", "keyword", NULL AS "package_id", NULL AS "package_name", NULL AS "repo_name", NULL AS "repo_owner", NULL AS "score", NULL AS "summary" FROM "search", UNNEST("keywords") AS "keyword" WHERE "keyword" = 'test' LIMIT 1) UNION ALL (SELECT 'package' AS "match_type", NULL AS "keyword", "package_id", "package_name", "repo_name", "repo_owner", "score", "summary" FROM "search" WHERE CONCAT_WS(' ', "package_name", COALESCE("summary", ''), "repo_name", "repo_owner") ~* 'test' AND "package_name" IS NOT NULL AND "repo_owner" IS NOT NULL AND "repo_name" IS NOT NULL ORDER BY LOWER("package_name") = 'test' DESC, "score" DESC, "package_name" ASC LIMIT 21 OFFSET 0)) AS "t" ORDER BY "match_type" = 'keyword' DESC, "match_type" = 'package' DESC
```

```
Sort  (cost=164.57..164.62 rows=22 width=188) (actual time=8.750..8.751 rows=22 loops=1)
  Sort Key: ((('keyword'::text) = 'keyword'::text)) DESC, ((('keyword'::text) = 'package'::text)) DESC
  Sort Method: quicksort  Memory: 30kB
  ->  Result  (cost=0.00..164.08 rows=22 width=188) (actual time=0.332..8.723 rows=22 loops=1)
        ->  Append  (cost=0.00..163.75 rows=22 width=185) (actual time=0.330..8.715 rows=22 loops=1)
              ->  Limit  (cost=0.00..1.30 rows=1 width=216) (actual time=0.329..0.331 rows=1 loops=1)
                    ->  Nested Loop  (cost=0.00..4945.24 rows=3812 width=216) (actual time=0.328..0.328 rows=1 loops=1)
                          ->  Seq Scan on search  (cost=0.00..142.12 rows=3812 width=69) (actual time=0.015..0.039 rows=156 loops=1)
                          ->  Function Scan on unnest keyword  (cost=0.00..1.25 rows=1 width=32) (actual time=0.001..0.001 rows=0 loops=156)
                                Filter: (keyword = 'test'::text)
                                Rows Removed by Filter: 3
              ->  Subquery Scan on "*SELECT* 2"  (cost=162.07..162.33 rows=21 width=184) (actual time=8.371..8.380 rows=21 loops=1)
                    ->  Limit  (cost=162.07..162.12 rows=21 width=185) (actual time=8.368..8.373 rows=21 loops=1)
                          ->  Sort  (cost=162.07..162.14 rows=30 width=185) (actual time=8.367..8.368 rows=21 loops=1)
                                Sort Key: ((lower(search_1.package_name) = 'test'::text)) DESC, search_1.score DESC, search_1.package_name
                                Sort Method: top-N heapsort  Memory: 34kB
                                ->  Seq Scan on search search_1  (cost=0.00..161.33 rows=30 width=185) (actual time=0.063..8.160 rows=96 loops=1)
                                      Filter: ((package_name IS NOT NULL) AND (repo_owner IS NOT NULL) AND (repo_name IS NOT NULL) AND (concat_ws(' '::text, package_name, COALESCE(summary, ''::text), repo_name, repo_owner) ~* 'test'::text))
                                      Rows Removed by Filter: 3716
Planning Time: 0.272 ms
Execution Time: 8.821 ms
```

Fixes #1179


Sources/App/Core/Search+Result.swift
```diff
@@ -0,0 +1,94 @@
+extension Search {
+    enum Result: Codable, Equatable {
```
finestructure, line 2: NB: this is currently relying on Swift 5.5 Codable auto synthesis and will not compile on Swift 5.4 (hence the CI failure).

We can either add manual conformance or move to Swift 5.5 before (not as part of) this PR.

finestructure: I've added manual conformance now, with a deprecation warning when using 5.5.

Sources/App/Core/Search.swift
```diff
@@ -75,57 +82,77 @@ enum Search {
...
-            .select
+    }
+
+    static func keywordMatchQueryBuilder(on database: Database,
+                                         terms: [String]) -> SQLSelectBuilder {
+        guard let db = database as? SQLDatabase else {
+            fatalError("Database must be an SQLDatabase ('as? SQLDatabase' must succeed)")
+        }
+        let mergedTerms = SQLBind(terms.joined(separator: " ").lowercased())
+        // FIXME: bookend with `%`
```
finestructure, line 139: This initial implementation is doing a simple `"keyword" = 'test'` filtering with `LIMIT 1` when searching for `test`.

We'll want to also match `unit-testing`, changing that to `"keyword" like '%test%'` with some limit on the results, say `LIMIT 3`. When we do that we'll also need to `DISTINCT` the results (because the `UNNEST` will generate lots of duplicates).

More importantly, we'll need to define a sort order of partial keyword matches. I'm thinking something like

1. exact match
2. partial matches `%test%` ordered by number of occurrences, alphabetically

This will require quite a few additional changes, so I've left it out of this already big PR.

Tests/AppTests/SearchTests.swift
```diff
@@ -448,9 +435,44 @@ class SearchTests: AppTestCase {
...
+
+        XCTAssertEqual(res.results, [
+            .keyword(.init(keyword: "topic")),
+            // TODO: also match against package keywords
+            //            .package(.init(packageId: .id2,
+            //                           packageName: "2",
+            //                           packageURL: "2",
+            //                           repositoryName: "2",
+            //                           repositoryOwner: "foo",
+            //                           summary: ""))
```
finestructure, line 474: This is the other piece that I've skipped in order to get the first version out: matching against a package's keywords.

One simple way to do this is by appending the topics to the search string:

```sql
WHERE CONCAT_WS(' ', "package_name", COALESCE("summary", ''), "name", "owner", ARRAY_TO_STRING(keywords, ' ')) ~* 'foo'
```

I _think_ we'll want to do that. In practise it might pull in a lot of results for certain terms, like for instance `test`, but I suppose that's a good thing. You can sort of do a keyword search without actually picking the keyword result from the search results.

finestructure:
This is good to go in principle but we should give this some thorough testing on staging before deploying because it includes nontrivial changes to search.

I'll deploy from the branch for now.

finestructure:
I've tested this quite a bit and it seems to be working fine.

daveverwer (approved):
What an epic!

finestructure:
I'll merge and deploy this as 2.41.0 now

PR #1197, opened 2021-06-28
Topic search UI

finestructure:
Exposing keyword search in an initial version:

CleanShot 2021-06-28 at 08 40 35@2x

CleanShot 2021-06-28 at 08 40 52@2x

Labelling and naming the results to be tweaked but this adds the functionality.


Sources/App/Controllers/KeywordController.swift
```diff
@@ -0,0 +1,45 @@
...
+struct KeywordController {
+
+    static func query(on database: Database, keyword: String) -> EventLoopFuture<[Package]> {
+        Package.query(on: database)
+            .with(\.$repositories)
+            .join(Repository.self, on: \Repository.$package.$id == \Package.$id)
+            .filter(
+                DatabaseQuery.Field.path(Repository.path(for: \.$keywords), schema: Repository.schema),
+                DatabaseQuery.Filter.Method.custom("@>"),
+                DatabaseQuery.Value.bind([keyword])
```
finestructure, line 15: This is not the query I originally intended to write. Fluent doesn't seem to support expressing `'foo' LIKE ANY("keywords")` so this is an equivalent workaround which performs similarly.

I've to a question open in the Vapor Discord but not takers yet: https://discord.com/channels/431917998102675485/684159753189982218/858402370144960552

Sources/App/Core/SiteURL.swift
```diff
@@ -196,6 +203,9 @@ enum SiteURL: Resourceable {
             case let .docs(next):
                 return ["docs"] + next.pathComponents
 
+            case .keywords:
+                return ["keywords", ":keyword"]
```
finestructure, line 207: FWIW, there's a "dead" `keywords` Github account which we're sort of overlaying here. But even so, `swiftpackageindex.com/keywords` should still work even if we add `swiftpackageindex.com/keywords/foo`.

CleanShot 2021-06-28 at 08 09 55@2x

finestructure:
NB: #1198 which I'll address separately.

Sources/App/Core/SiteURL.swift
daveverwer: Yea this is fine.

daveverwer (commented):
I think we probably want to put the actual keyword searches behind a feature flag for now and have them only appear on staging. Couple of reasons:

* Keyword search makes less sense when the keywords are not exposed on the package pages, which Sarah is working on.
* I'd like to look at the display of search results and potentially bring icons (not emoji) into the list of search results.

If possible, we should only include keyword search results on staging deployments for now.

finestructure:
Yeah, the UI is a placeholder, even more so for the upcoming author search where there's no applicable emoji 😅

I won't fully feature flag this but disable it for prod.

finestructure:
This is good to go now.

daveverwer (approved):

daveverwer:
All good to go. I'll let you rebase/merge in main and then merge the PR.

finestructure:
BTW, those web view snapshots I included almost certainly need updating on Catalina, like the rest.

PR #1199, opened 2021-06-28
Fix keyword paging

### LINKED ISSUE #1198: Keyword matches appear on every search page
Search paging doesn't currently take into account that keyword results are now part of the total result set. This leads to the following page layout issue for page size `N`:

```
kw
pkg1
...
pkgN

---

kw
pkgN+2
pkg2N+1

---

kw
pkg2N+3
...
```

I.e. the `kw` result appears at the top of each result page and truncates out one package result, because the offsets and limits assume to be working on the package results only.

finestructure:
Fixes #1198 

Merge after #1197

### DIFF
--- a/Sources/App/Core/Search+Result.swift
+++ b/Sources/App/Core/Search+Result.swift
@@ -29,6 +29,24 @@ extension Search {
                     )
             }
         }
+
+        var isKeyword: Bool {
+            switch self {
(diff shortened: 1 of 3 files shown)


daveverwer (approved):
Feels like we should merge this into #1197 so they both go together when we have the feature flag?

PR #1192, opened 2021-06-24
add owner_name as collectionName and overview.

at-syot:
[use owner fullname](https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/issues/1160)
[add overview](https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/issues/1159)

I think this is not done yet, i just want to update you some progress and need to ask more about the issue.
I'm not sure that should i change a PackageCollection.generate func signature ? 
because i saw this has used by collection API route (only development env) to.

Result: 
Screen Shot 2564-06-24 at 17 35 08

PR #1201, opened 2021-06-29
Collection ownername

### LINKED ISSUE #1159: Add a collection overview to package collections
It currently says "This collection does not have an overview" when adding a collection to Xcode.

Screenshot 2021-06-08 at 12 59 46@2x

It should say "A collection of packages authored by Vapor from the Swift Package Index"

**Note:** Use the full owner name rather than just the owner username when building this description

### LINKED ISSUE #1160: Use full owner name when building package collections
We should use the full owner name here rather than the username.

Screenshot 2021-06-08 at 13 03 18@2x

finestructure:
Supersedes #1192 

Fixes #1159
Fixes #1160

### DIFF
--- a/Sources/App/Core/PackageCollection+generate.swift
+++ b/Sources/App/Core/PackageCollection+generate.swift
@@ -65,8 +65,10 @@ extension PackageCollection {
             .and(ownerQuery.first())
             .map { packages, repository in
                 var name = collectionName
+                var overview = String()
                 if let ownerName = repository?.ownerName {
                     name = "Packages by \(ownerName)"
+                    overview = "A collection of packages authored by \(ownerName) from the Swift Package Index"
                 }
                 
                 return PackageCollection.init(
--- a/Sources/App/Core/PackageCollection+generate.swift
+++ b/Sources/App/Core/PackageCollection+generate.swift
@@ -46,27 +46,6 @@ extension PackageCollection {
                 query = query
                     .filter(App.Package.self, \.$url ~~ packageURLs)
         }
-        
-        // generate ownerName from [App.Package]
-        
-        func ownerName(packages: [App.Package]) -> String {
-            let groupedPackagesByName = Dictionary(
-                grouping: packages,
-                by: { $0.repository?.ownerName ?? $0.repository?.owner })
-            
-            let names = groupedPackagesByName.enumerated().compactMap { $0.element.key }
-            switch names.count {
-            case 0:
-                // shouldn't be possible really
-                return ""
-            case 1:
-                return names.first!
-            case 2:
-                return names.joined(separator: " and ")
-            default:
-                return "multiple authors"
-            }
-        }
 
         return query.all()
             .map { versions in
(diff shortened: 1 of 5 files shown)


Sources/App/Core/PackageCollection+generate.swift
```diff
@@ -55,21 +55,48 @@ extension PackageCollection {
...
-                    packages: $0,
+                    packages: packages,
                     formatVersion: .v1_0,
                     revision: revision,
                     generatedAt: Current.date(),
                     generatedBy: authorName.map(Author.init(name:)))
             }
     }
+
+    static func ownerName(packages: [App.Package]) -> String {
```
at-syot, line 81: look better.

at-syot (approved):

finestructure:
Draft until I've added an explicit test for `ownerName` substitution.

finestructure:
The tests exposed a few little edge cases but it's all sorted now.

Thanks again @at-syot , this has been really helpful!

at-syot:
> The tests exposed a few little edge cases but it's all sorted now.
> 
> Thanks again @at-syot , this has been really helpful!

Oh, the thx should go to you guys, i didn't do much haha. the true is i'm not sure about an expected result and how much i can make an edit to the function, i mean, am i allow to do this ? that's why i open draft pull request first to update, something like that.
So thank you again to lets me do this issue and the improved code is look good to.

at-syot:
Just to confirm, after review the changed code, i assume that Package's collection name will stay the same with owner(short name), Only overview change to use owner_name's (multiple owner's full-name) instead. Am i understand correct ?

finestructure:
No, the name is also changing. You can see it in the validation of the new test that I added, `test_generate_ownerName`:

```swift
          XCTAssertEqual(res.name, "Packages by Foo Org")
         XCTAssertEqual(res.overview, "A collection of packages authored by Foo Org from the Swift Package Index")
```

What I did though was ensure that you can still provide your own name and overview to `generate` and they'll be used as is. Only if the parameters are unset (`nil`) will these changes take effect.

Does that make sense?

at-syot:
> No, the name is also changing. You can see it in the validation of the new test that I added, `test_generate_ownerName`:
> 
> ```swift
>          XCTAssertEqual(res.name, "Packages by Foo Org")
>          XCTAssertEqual(res.overview, "A collection of packages authored by Foo Org from the Swift Package Index")
> ```
> 
> What I did though was ensure that you can still provide your own name and overview to `generate` and they'll be used as is. Only if the parameters are unset (`nil`) will these changes take effect.
> 
> Does that make sense?

Ah, i see it now.

finestructure:
I should add that I ended up making what perhaps looks like a lot of changes on top of your changes but it's really just re-arranging a few things and making the tests pass.

The tests are a great guide and will often instruct you what to still need to do to complete a task.

For instance here a lot of the additional changes were due to the second filter option `Filter.urls([String])`. Fixing the related tests made me realise that we need to re-arrange this a bit more, but the basic flow remained unchanged.

daveverwer (approved):
Fantastic work @at-syot! This is a great step forward.

PR #1209, opened 2021-07-05
Swift 5.5

finestructure:
This switches to a base image based on a Swift 5.5 nightly.

I've also moved away from the `Dockerfile.base` image and placed it in a project that has build automation and some basic image tests in place, as well as using the image registry to host them: https://gitlab.com/finestructure/spi-base/

I'm not sure it's the right time to make the switch. I'm currently using Xcode 13 to work on SPI but we have to bear in mind that there are other contributors who perhaps aren't yet. We probably want to make sure we keep development and deployment versions in sync.


Sherlouk:
For whatever it's worth, I'm setup with Xcode 13 👍  So no issues for me either way

daveverwer (approved):

daveverwer (approved):

finestructure:
Unless anyone objects I propose we merge this so we can proceed with adopting 5.5 in the analysis and build system.

daveverwer:
👌

PR #1212, opened 2021-07-05
Set badge to 'pending' when builds incomplete

### LINKED ISSUE #953: Improve badge message when builds are pending
Stephen just pinged me, the badges seem to be broken:  https://swiftpackageindex.com/pointfreeco/swift-composable-architecture

82FF5B9A-265B-4B9B-82FA-62125B702978

Sherlouk:
Closes #953

### DIFF
--- a/Sources/App/Controllers/API/Package+Badge.swift
+++ b/Sources/App/Controllers/API/Package+Badge.swift
@@ -3,7 +3,7 @@ import Vapor
 
 extension Package {
 
-    enum CompatibilityResult<Value> {
+    enum CompatibilityResult<Value: Equatable>: Equatable {
         case available([Value])
         case pending
         
@@ -92,32 +92,40 @@ extension Package {
                 label = "Swift Compatibility"
         }
 
-        let message = badgeMessage(badgeType: badgeType)
+        let (message, success) = badgeMessage(badgeType: badgeType)
         return Badge(schemaVersion: 1,
                      label: label,
                      message: message,
-                     isError: message == nil,
-                     color: message == nil ? "inactive" : "F05138",
+                     isError: !success,
+                     color: success ? "F05138" : "inactive",
                      cacheSeconds: cacheSeconds,
                      logoSvg: Package.loadSVGLogo())
     }
 
 
-    func badgeMessage(badgeType: BadgeType) -> String {
+    func badgeMessage(badgeType: BadgeType) -> (title: String, success: Bool) {
         switch badgeType {
             case .platforms:
                 switch platformCompatibility() {
                 case .available(let platforms):
-                    return _badgeMessage(platforms: platforms) ?? "unavailable"
+                    if let message = _badgeMessage(platforms: platforms) {
+                        return (message, true)
+                    } else {
+                        return ("unavailable", false)
+                    }
                 case .pending:
-                    return "pending"
+                    return ("pending", false)
                 }
             case .swiftVersions:
                 switch swiftVersionCompatibility() {
                 case .available(let versions):
-                    return _badgeMessage(swiftVersions: versions) ?? "unavailable"
+                    if let message = _badgeMessage(swiftVersions: versions) {
+                        return (message, true)
+                    } else {
+                        return ("unavailable", false)
+                    }
                 case .pending:
-                    return "pending"
+                    return ("pending", false)
                 }
         }
     }
--- a/Sources/App/Controllers/API/Package+Badge.swift
+++ b/Sources/App/Controllers/API/Package+Badge.swift
@@ -3,10 +3,26 @@ import Vapor
 
 extension Package {
 
+    enum CompatibilityResult<Value> {
+        case available([Value])
+        case pending
+        
+        var values: [Value] {
+            if case .available(let values) = self {
+                return values
+            }
+            
+            return []
+        }
+    }
+    
     /// Returns a list of compatible swift versions across a package's significant versions.
     /// - Returns: Array of compatible `SwiftVersion`s
-    func swiftVersionCompatibility() -> [SwiftVersion] {
-        let builds = allSignificantBuilds()
+    func swiftVersionCompatibility() -> CompatibilityResult<SwiftVersion> {
+        let allBuilds = allSignificantBuilds()
+        if allBuilds.anyPending { return .pending }
+        
+        let builds = allBuilds
             .filter { $0.status == .ok }
         let compatibility = SwiftVersion.allActive.map { swiftVersion -> (SwiftVersion, Bool) in
             for build in builds {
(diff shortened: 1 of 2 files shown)


Sherlouk:
Tests are being awkward locally and not running - pushing up for a CI run but 🤞 it all works.

Sherlouk:
There's a feeling you get when things work first time and it's oh so good :chef_kiss:

Tests/AppTests/PackageTests.swift
```diff
@@ -756,11 +756,46 @@ final class PackageTests: AppTestCase {
...
+
+        // MUT
+        let res = p.swiftVersionCompatibility()
+
+        // validate
+        if case .pending = res {
+            return
+        }
+        
+        XCTFail("Compatibility result should have been set to pending")
```
finestructure, line 797: I'd make this

```swift
        // validate
        XCTAssertEqual(res, .pending)
```

via

```swift
    enum CompatibilityResult<Value: Equatable>: Equatable {
```

(It's a bit more constrained than technically necessary but if that's a concern we could also only make it `Equatable` in the test target - but I think we don't need to be concerned about that at this time and it makes the test nicer :)

Tests/AppTests/PackageTests.swift
```diff
@@ -787,11 +822,46 @@ final class PackageTests: AppTestCase {
...
+
+        // MUT
+        let res = p.platformCompatibility()
+
+        // validate
+        if case .pending = res {
+            return
+        }
+        
+        XCTFail("Compatibility result should have been set to pending")
```
finestructure, line 863: dot

Sources/App/Controllers/API/Package+Badge.swift
```diff
@@ -3,10 +3,26 @@ import Vapor
...
+        }
+    }
+    
     /// Returns a list of compatible swift versions across a package's significant versions.
     /// - Returns: Array of compatible `SwiftVersion`s
-    func swiftVersionCompatibility() -> [SwiftVersion] {
-        let builds = allSignificantBuilds()
+    func swiftVersionCompatibility() -> CompatibilityResult<SwiftVersion> {
+        let allBuilds = allSignificantBuilds()
+        if allBuilds.anyPending { return .pending }
```
finestructure, line 23: So one questions I have is this: do we want to make the whole value pending if just a single item is pending? If for some reason a job keeps dying the whole item will be pending.

Compatibility is a "wide" status item in that it aggregates across platforms and Swift versions: if any platform support a given swift version it's called compatible with that Swift version, and the same for platforms. The way I think of it is a logical "OR" across those two dimensions.

It feels like "pending" should also be "OR" and only "win" if all items are "pending". Chances are quite high that a single stuck build will be covered via the "OR"s across the Swift versions and platforms and "pending" would be unnecessarily strict.

Does that make sense?

Sherlouk: So to clarify, you'd rather change this from `if any pending` to `if all pending`?

finestructure: I _think_ so but I haven't actually played with this in earnest. I was looking at it through the lens of the issue report: we had a package with all builds pending and showed "unavailable". I think we want to show "pending" in that case and this patch would do that.

However, I remember that as soon as a few builds came through, the matrix started filling out (it does to quickly, because of the "OR", even when a number of builds are still missing) and I feel like we should reflect that in the badge as well.

Just wanted to open the discussion on that while it's warm!

Sherlouk: From a technical perspective, either are possible and super quick so happy to take a steer from you folks on that!

My personal opinion is that an incomplete badge could be misleading/confusing to users within the context of a README, and I'd rather see a badge which says "pending" rather than suggest I don't support a platform the user is looking for.

As somebody purely looking at a README, if these badges are a first impression and answer a quick question of compatibility without expecting the user to read much into it... I could've just lost a user!

Personal opinion, as always, your platform 😄

finestructure: That's a good point. The problem I see though is that a correct badge will be pending until all builds are complete even when it's already "filled out".

Say for instance all builds except the Linux 5.0 build are done: the whole build is pending and yet the badge would already correctly show Linux support, because the 5.2+ builds have completed. I.e. we'd "grey out" the badge until the slowest job is complete even though they late builds very likely won't add any more info.

Perhaps we could make this

```
[iOS | macOS | (some pending)]
```

but only if we lack info to fill it out. I.e. if we have a lagging build but already know from the other Swift versions that a platform is supported (likewise for Swift versions via other platforms) then we don't mark it pending.

finestructure: I'm thinking of this in terms of projecting the compatibility matrix down to the two axes for the badge labels:
 
CleanShot 2021-07-05 at 17 22 42@2x

Only if there's no "ok" in a column or row do we mark it as pending.

finestructure: And to contrast this to the PR as I understand it, that would display this instead:

CleanShot 2021-07-05 at 17 26 22@2x

daveverwer: > Only if there's no "ok" in a column or row do we mark it as pending.

I was just typing this up! This is exactly how it should work 👍

finestructure: Addendum: or if they're all failed - in case it's not pending either of course!

Sherlouk: It's worth saying if they all come back as failed, with both the current and this PR's version we would return "unavailable". Is this still fair? Or do we also now want an error case too?

finestructure: I think that’s fair. The badge only shows “supported” and an “all failed” is absence of the particular col/row. If there’s no “supported” at all for any row/col, “unavailable” seems like the only thing we can sensibly display (as I don’t think there’s a way to withdraw the whole badge).

Sherlouk: Switched to check for all satisfy pending, if we believe it'd make more sense to make it (as you suggested above) like `[iOS | macOS | (some pending)]` then let me know and I can follow that up.

finestructure: Looks good to me, thanks a lot, James!

I don't think we need to bother with the `(some pending)` for the moment. If I'm not mistaken we've simply omitted incomplete builds all along so this is not a regression.

We can always follow up if we change our minds.

finestructure (approved):

finestructure:
Great stuff, thanks James!

PR #1213, opened 2021-07-05
Display Release Tab within Package Page

### LINKED ISSUE #863: Display release information on the package page
Now that we have the release information in the database (#414), we should visualise it better, including release notes on the package page.

This is being tackled as part of the package page design changes but needs tracking.

Sherlouk:
Closes #863 

This pull requests adds a new tab bar to the package page. This tab bar has two tabs: the existing README and a new "Release Notes" page which links into the data added in #829.

This does not show release tag messages, a card has been raised under #1223 to support this. A message is shown if no releases are available, or if no notes are attached to a version. Release titles have a link to the GitHub page for convenience.

Some images:
Captured 2021-07-09 at 2 46 33
Captured 2021-07-09 at 2 46 58
Captured 2021-07-09 at 2 47 11

Tab bar hover state:
Captured 2021-07-09 at 2 46 46


Sherlouk:
This is what the releases tab is looking like.

Captured 2021-07-05 at 4 53 21

Sherlouk:
Captured 2021-07-05 at 4 55 43

Not all packages have details, but you do still get this...

Sources/App/Views/PackageController/PackageReleases+View.swift
```diff
@@ -0,0 +1,40 @@
...
+            super.init()
+        }
+
+        override func frameIdentifier() -> String {
+            "releases"
+        }
+
+        override func frameContent() -> Node<HTML.BodyContext> {
+            guard let releases = model?.releases
+            else { return .empty }
```
Sherlouk, line 21: Captured 2021-07-05 at 4 56 21

Unfortunately, some packages just return 0 releases (the repo has tags, just not releases).

This gets caught here in this guard, we'll want to return some form of empty state here but no idea what it would look like.

Sherlouk:
Going to take a pause on this until I get some feedback from Dave/Sven. Of course I still need to add/update tests and a bit of tidying but would like some direction 👍  No rush!

Sherlouk:
Captured 2021-07-05 at 5 00 21

Bit of an edge case but tuist/tuist add the version to the top of their notes, which looks like this...

daveverwer:
OK, as discussed on Discord. Tab bar design should look like this:

Standard State:

New Design - Normal

and when hovering over an unselected tab:

New Design – Hovered

* Tab bar background is `very-very-light-grey`
* Tab bar border is `light-grey`
* Selected tab background is `very-light-grey`
* Selected tab border is `light-green`

daveverwer:
> Captured 2021-07-05 at 4 55 43
> 
> Not all packages have details, but you do still get this...

So with this. I'm guessing this is the case where GitHub has defined "releases" but no release notes have been added by the package author?

In this case, we should include the version number and have a paragraph underneath it with "No release notes."

daveverwer:
> Captured 2021-07-05 at 5 00 21
> 
> Bit of an edge case but tuist/tuist add the version to the top of their notes, which looks like this...

Even though this is an edge case, I think it's worth fixing. We should look at the first heading in the release notes and remove the whole heading if it contains the version number. We should be able to use SwiftSoup for this server-side.

The alternative would be to only remove it if it's an exact match for *just* the version number, but I feel like the likely case for extra words here are things like the word "Version" before the version number, so let's remove it.

Sources/App/Views/PackageController/PackageReleases+View.swift
daveverwer: > Captured 2021-07-05 at 4 56 21
> 
> Unfortunately, some packages just return 0 releases (the repo has tags, just not releases).
> 
> This gets caught here in this guard, we'll want to return some form of empty state here but no idea what it would look like.

I think this shouldn't necessarily be a "Releases" tab, but "Release Notes" instead. I'm not sure I'd stick just a standard "Releases" tab down here with the README. We kinda cover "Releases" up in the top section with the significant versions.

So, with that distinction. I think we just have content saying "This package has no release notes".

Sherlouk:
Thanks for all the help there Dave, I'll get cracking on that as soon as possible! 

I'll also work on fixing the JS that you slated on Discord 😉

Sources/App/Views/PackageController/PackageShow+View.swift
```diff
@@ -181,8 +183,51 @@ enum PackageShow {
...
+                .hr(
+                    .class("minor")
+                ),
+                .section(
+                    .class("package_tabs"),
+                    .ul(
+                        .li(
+                            .a(
+                                .href("#readme"),
+                                .class("package_tab_link"),
```
daveverwer, line 218: I don't think we'll need this class name if we write the JS using `querySelectorAll`.

Sources/App/Views/PackageController/PackageShow+View.swift
```diff
@@ -181,8 +183,51 @@ enum PackageShow {
...
+                        )
+            )
+        }
+        
+        func tabBar() -> Node<HTML.BodyContext> {
+            .group(
+                .hr(
+                    .class("minor")
+                ),
+                .section(
```
daveverwer, line 212: It's definitely semantically valid to put the README and the Release Notes into a `section`, but I'm not sure about the tab bar itself. I think this could just be the `ul` on its own, with a class that indicates a tab bar.

daveverwer (changes requested):
I've made several general JS style suggestions inline, but overall I'd also recommend a slightly different approach to the JavaScript here.

Rather than keeping everything hard coded, I’d try and make the JS a little more generic so it can be used in other “tab” situations. 

Even better, we could make this JS a little more self-contained by creating a custom tag. If we put the `ul` element inside a custom `tab-bar` element, and possibly even put the tab content inside custom `tab-page` elements then this JS could be really simple and clean.

We should use a `data-` attribute on the tab bar link elements to pass down to the JS the data about which tab page to make visible, and then we can loop around all `tab-page` elements hiding the rest.

I’m in the middle of moving more of our JS to custom tags, so we should try and do this if possible. I’m happy to help with the learning curve, if you’d like!

Happy to chat through any of this!

Sherlouk:
I don't know how to process the fact that an open source contribution got a better code review than my actual work... but thanks Dave!

I knew I needed to rewrite the JS anyway but I like the approach you've detailed there so I'll be sure to give it a shot!

Resources/Scripts/package_tabs.js
```diff
@@ -0,0 +1,40 @@
...
+        var links = tabs[i].getElementsByTagName('a');
+
+        for (var j = 0; j < links.length; j++) {
+            var link = links[j];
+
+            if (!link.classList.contains("active")) {
+                document.getElementById(link.hash.substring(1)).style.display = "none";
+            }
+
+            link.addEventListener('click', function(event) {
```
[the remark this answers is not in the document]
Sherlouk: keydown is not working in my testing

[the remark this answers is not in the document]
daveverwer: Oh, sorry. That was just an example event. The `click` event is correct. I just meant the `(event) => {` style of closure.

Sherlouk:
I'm still experiencing crashes when running tests on my local machine so I am unable to update the snapshot tests. Equally I've not been able to create a test (as I would've liked to do) that tested the model initialiser with a real repository instance for the same reason.

I have added two new snapshot tests which should cover the various different states of the view. I've also added a handful of unit tests to test some of the specific logic.

Generally speaking this should be ready to go with a bit of help getting that green test status 😄

Sherlouk:
Rebased, tests added, good to review 👍

daveverwer:
> Even though this is an edge case, I think it's worth fixing. We should look at the first heading in the release notes and remove the whole heading if it contains the version number. We should be able to use SwiftSoup for this server-side.
> 
> The alternative would be to only remove it if it's an exact match for _just_ the version number, but I feel like the likely case for extra words here are things like the word "Version" before the version number, so let's remove it.

I'm not sure if you got to this, although there's a commit that looks promising. However, it's still showing the version number headers for tuist in the latest version.

Screenshot 2021-07-12 at 11 16 41@2x

Resources/Scripts/main.js
```diff
@@ -10,6 +10,7 @@ import { SPIBuildLogNavigation } from './build_log_navigation.js'
 import { SPIAutofocus } from './autofocus.js'
 import { SPIPlaygroundsAppLinkFallback } from './playgrounds_app_link.js'
 import { SPIReadmeElement } from './readme_element.js'
+import { SPITabBarElement } from './package_tabs.js'
```
daveverwer, line 13: Good rename of the class, but let's keep the filename in sync as `tab_bar_element.js`.

Sherlouk:
Hmm. I definitely tested `tuist` and it was working, leave it with me and I'll figure it out again.

Sources/App/Views/PackageController/PackageShow+View.swift
```diff
@@ -182,8 +184,42 @@ enum PackageShow {
...
+                            .data(named: "tab-page", value: "releases"),
+                            // Until the content is loaded, substitute a spinner.
+                            .spinner()
+                        )
+            )
+        }
+        
+        func tabBar() -> Node<HTML.BodyContext> {
+            .group(
+                .element(named: "tab-bar", nodes: [
```
daveverwer, line 210: Let's add a `Node` extension that defines `tab-bar`, as we have with `spi-readme`. The `ul` should still be in this file, just the `tab-bar` element.

We also might want to call it `spi-tab-bar`, just in case any extensions or similar define a `tab-bar` element.

Sources/App/Views/PackageController/PackageShow+View.swift
```diff
@@ -182,8 +184,42 @@ enum PackageShow {
                         source: SiteURL.package(.value(model.repositoryOwner),
                                                 .value(model.repositoryName),
                                                 .readme).relativeURL(),
-                        // Until the content is loaded, substitute a spinner.
-                        .spinner()
+                        .group(
```
daveverwer, line 187: I don't think this `group` is needed.

Sources/App/Views/PackageController/PackageShow+View.swift
```diff
@@ -182,8 +184,42 @@ enum PackageShow {
...
+                        )
+            )
+        }
+        
+        func releaseSection() -> Node<HTML.BodyContext> {
+            .turboFrame(id: "releases",
+                        source: SiteURL.package(.value(model.repositoryOwner),
+                                                .value(model.repositoryName),
+                                                .releases).relativeURL(),
+                        .group(
```
daveverwer, line 200: Same as above.

Resources/Styles/tab-bar.scss
```diff
@@ -0,0 +1,39 @@
...
+      }
+
+      &.active {
+        border-bottom: 3px var(--tab-bar-border-active) solid;
+      }
+    }
+  }
+}
+
+[data-tab-page].hidden {
```
daveverwer, line 37: Ah, I see the problem here and why you can't use the existing `hidden` class. 👍

Sources/App/Views/PackageController/PackageShow+View.swift
daveverwer: There's the very briefest moment when both tab pages are loaded. I think we should add the `hidden` class here at the server so both are hidden, and then one is shown.

daveverwer (changes requested):
From my side, everything is good to go once these minor changes are done! Thanks James!

daveverwer:
Tagging @finestructure for a quick review too.

finestructure (approved):

Sherlouk:
Spoken to Dave on Discord, but for the sake of public comms - the issue above with Tuist is very specific to a recent release of theirs and is too niche to warrant a wider fix.

It seems that [this version](https://github.com/tuist/tuist/releases/tag/1.46.0) has like 5 lots of changelogs in a single version. It's far from the norm, and so we've okayed it (unless it becomes way more common and troublesome)

daveverwer (approved):

daveverwer:
Deploying as 2.41.10

PR #1214, opened 2021-07-06
Tidy up

finestructure:
While reviewing James' PR I noticed there were two top-level functions `_badgeMessage` from my original implementation that should have been static functions on `Package`, really.

I wanted to clean that up and while I was in there I noticed the doc strings had drifted and that `values` should be `[Value]?` so `pending` can map to `nil` exclusively (whereas `[]` could also be `.available([])`.

Just small stuff that I thought I'd just fix up real quick rather than go back and forth in the existing PR. Hope you don't mind, James!

### DIFF
--- a/Sources/App/Controllers/API/Package+Badge.swift
+++ b/Sources/App/Controllers/API/Package+Badge.swift
@@ -166,21 +166,21 @@ extension Package {
             Set(
                 platforms
                     .map { p -> Value in
-            switch p {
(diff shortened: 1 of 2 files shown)


Sherlouk (approved):
Aha no worries! Looks good - I probably would've ended up moving around some of these functions if we were to add the pending note to the message anyways!

daveverwer (approved):

PR #1216, opened 2021-07-06
Fix collection verified compatibility

### LINKED ISSUE #1215: Swift Package Collections returning incorrect "verifiedCompatibility" matrix
Having looked at the collection JSON for my packages, I've noticed that the "verifiedCompatibility" object is inaccurate for my packages (and presumably others too...)

If we take a look at this package: https://swiftpackageindex.com/Sherlouk/swift-snapshot-testing-stitch

We can see that the build matrix has correctly identified that this package only works with iOS on Swift 5.2 and newer (it has a UIKit requirement so won't work on other platforms!)

If we take a look at the collection JSON: https://swiftpackageindex.com/Sherlouk/collection.json ([Download from GitHub](https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/files/6769599/Sherlouk-collection.json.zip))

We can see that for the same package the compatibility object includes *every* build target that we run against: including iOS Swift 5.0 which the package doesn't support, and even other platforms such as macOS, tvOS, and Linux - which my package doesn't support.

This is repeated for all other packages in my collection.

finestructure:
Fixes #1215

### DIFF
--- a/Sources/App/Core/PackageCollection+generate.swift
+++ b/Sources/App/Core/PackageCollection+generate.swift
@@ -282,7 +282,7 @@ private extension PackageCollection.ProductType {
 }
 
 
-private extension Array where Element == PackageCollection.Compatibility {
+extension Array where Element == PackageCollection.Compatibility {
     // Helper struct to work around Compatibility not being Hashable
     struct Pair: Hashable {
         var platform: PackageCollection.Platform
--- a/Sources/App/Core/PackageCollection+generate.swift
+++ b/Sources/App/Core/PackageCollection+generate.swift
@@ -294,8 +294,10 @@ extension Array where Element == PackageCollection.Compatibility {
             // Gather up builds via a Set to de-duplicate various
             // macOS build variants - spm, xcodebuild, ARM
             Set<Pair>(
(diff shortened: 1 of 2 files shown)


Sources/App/Core/PackageCollection+generate.swift
```diff
@@ -294,8 +294,10 @@ private extension Array where Element == PackageCollection.Compatibility {
             // Gather up builds via a Set to de-duplicate various
             // macOS build variants - spm, xcodebuild, ARM
             Set<Pair>(
-                builds.map { Pair.init(platform: .init(platform: $0.platform),
-                                       version: $0.swiftVersion.displayName) }
+                builds
+                    .filter { $0.status == .ok}
```
daveverwer, line 298: Some lines are more important than others! 😬

daveverwer (approved):

Sources/App/Core/PackageCollection+generate.swift
Sherlouk: Aha, the smallest things 'eh

Thanks for getting this one fixed so quickly Sven!

PR #1220, opened 2021-07-07
added archived list item

selichter:
- Added a list item for archived warning

PR #1221, opened 2021-07-07
Added a new archived repository list item

daveverwer:
Continuation of #1220.
Supercedes #1174.

### DIFF
--- a/Resources/Styles/package.scss
+++ b/Resources/Styles/package.scss
@@ -36,9 +36,9 @@
           background-position: top 2px left;
         }
         
-        li.history {
+        li.archived {
           grid-column-start: span 2;
-          background-image: var(--image-clock);
+          background-image: var(--image-warning);
         }
 
         li.history {
--- a/Resources/Styles/package.scss
+++ b/Resources/Styles/package.scss
@@ -35,6 +35,11 @@
           background-repeat: no-repeat;
           background-position: top 2px left;
         }
+        
+        li.history {
+          grid-column-start: span 2;
+          background-image: var(--image-clock);
+        }
 
         li.history {
           grid-column-start: span 2;
--- a/Sources/App/Views/PackageController/PackageShow+Model.swift
+++ b/Sources/App/Views/PackageController/PackageShow+Model.swift
@@ -199,16 +199,11 @@ extension PackageShow.Model {
     
     func archivedListItem() -> Node<HTML.ListContext> {
         if isArchived {
-            var archiveSentenceFragments: [Node<HTML.BodyContext>] = []
-            archiveSentenceFragments.append(contentsOf: [
-                .strong("No longer in active development."),
-                " The package author has archived this project and the repository is read-only."
-            ])
             return .li(
                 .class("archived"),
-                .group(archiveSentenceFragments)
+                .strong("No longer in active development."),
+                " The package author has archived this project and the repository is read-only."
             )
-            
         } else {
             return .empty
         }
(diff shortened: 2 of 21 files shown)


Resources/Styles/package.scss
```diff
@@ -35,6 +35,11 @@
           background-repeat: no-repeat;
           background-position: top 2px left;
         }
+        
+        li.history {
```
daveverwer, line 39: I think this might be a copy/paste erropr. I think it should be `archived`.

Resources/Styles/package.scss
```diff
@@ -35,6 +35,11 @@
...
+          grid-column-start: span 2;
+          background-image: var(--image-clock);
```
daveverwer, line 41: So we don't have a source for icons. I just draw them! 😬

I was going to draw a new one for this, but actually I think `var(--image-warning)` works well here.

Sources/App/Views/PackageController/PackageShow+Model.swift
```diff
@@ -232,6 +231,18 @@ extension PackageShow.Model {
         )
     }
     
+    func archivedListItem() -> Node<HTML.ListContext> {
```
daveverwer, line 234: Nitpick: Let's keep these ordered in the file the same way they are used in the output, so let's move this up to below the `authorsListItem`.

Sources/App/Views/PackageController/PackageShow+Model.swift
```diff
@@ -232,6 +231,18 @@ extension PackageShow.Model {
...
+         if isArchived {
+             return .li(
+                 .class("archived"),
+                 .strong("No longer in active development.")
```
daveverwer, line 238: I know I said in #1174 that the proposed wording was good, but now I look at it in place, I was wrong. 😬

We should move more of the text from the second line up into this item. So it should say:

> **No longer in active development.** The package author has archived this project and the repository is read-only.

Then, the `history` item should say:

> It had XXX commits and XXX releases before being archived.

daveverwer: Actually, now that `history` line reads strangely as it's not super clear what `It` refers to where it was when it followed on after the previous sentence. How about this for the `history` line:

> Before being archived, it had XXX commits and XXX releases.

daveverwer (commented):
I left some comments for you above, @selichter!

Thanks so much for this 👍

Resources/Styles/package.scss
selichter: Ugh. Yes. Thanks for catching that!

selichter:
@daveverwer Can you check my write access again?  I've made the changes locally, but can't push to this branch.

daveverwer:
> @daveverwer Can you check my write access again? I've made the changes locally, but I still have the notification that I don't have write access. Thanks!

It still says Pending Invite on GitHub so you should have an invite email that you'll need to accept. Did it get caught in spam?

selichter:
> > @daveverwer Can you check my write access again? I've made the changes locally, but I still have the notification that I don't have write access. Thanks!
> 
> It still says Pending Invite on GitHub so you should have an invite email that you'll need to accept. Did it get caught in spam?

The spam folder always gets me. I have it now. Thank you!

daveverwer (commented):
I just made one final, minor change here @selichter.

Instead of putting the sentence together in `archiveSentenceFragments` which then need wrapping in a `group`, we can just pout all of the fragments directly into the `li` item:

```swift
return .li(
    .class("archived"),
    .strong("No longer in active development."),
    " The package author has archived this project and the repository is read-only."
)
```

I committed that directly, and this is now good to go. Congratulations on your first PR to the SPI! 🎉

finestructure (approved):

finestructure:
Thanks, Sarah, and welcome to the SPI contributors! :) 🎉

PR #1222, opened 2021-07-07
Keyword pagination

### LINKED ISSUE #1217: Add pagination to author and keyword landing pages
e.g. https://staging.swiftpackageindex.com/keywords/swift should paginate instead of creating a monster page.

finestructure:
Fixes #1217

This adds pagination to the keyword package list.

I briefly looked into extending this to author package lists as well but a) it can't really be factored out and b) there are currently only 6 authors where it'd even paginate:

CleanShot 2021-07-07 at 13 57 53@2x

It's just not worth adding for authors.

### DIFF
--- a/Tests/AppTests/__Snapshots__/WebpageSnapshotTests/test_KeywordShow.1.txt
+++ b/Tests/AppTests/__Snapshots__/WebpageSnapshotTests/test_KeywordShow.1.txt
@@ -133,6 +133,7 @@
         <p>
           <strong>10 packages.</strong>
         </p>
+        <ul class="pagination"></ul>
       </div>
     </main>
     <footer>
--- a/Sources/App/Controllers/API/API+SearchController.swift
+++ b/Sources/App/Controllers/API/API+SearchController.swift
@@ -11,7 +11,7 @@ extension API {
             return search(database: req.db,
                           query: query,
                           page: page,
-                          pageSize: Constants.searchPageSize)
+                          pageSize: Constants.resultsPageSize)
         }
     }
 }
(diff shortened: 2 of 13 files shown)


Tests/AppTests/__Snapshots__/WebpageSnapshotTests/test_KeywordShow.1.txt
```diff
@@ -133,6 +133,7 @@
         <p>
           <strong>10 packages.</strong>
         </p>
+        <ul class="pagination"></ul>
```
finestructure, line 136: We'll also need to update the image snapshot before merging

daveverwer: I just updated the snapshots and there's no image diff with this change to the HTML. The `ul` would only render visually if there were a `li` inside it.

Maybe the test should have more test data so it shows actual pagination controls?

daveverwer (approved):

Tests/AppTests/__Snapshots__/WebpageSnapshotTests/test_KeywordShow.1.txt
finestructure: I've just had a look but unless we also fiddle with the snapshot size it's not going to show the page links:

CleanShot 2021-07-08 at 10 20 40@2x

I'd say we just leave it at that, ok?

daveverwer: Absolutely!

PR #1224, opened 2021-07-09
Fallback to commit message for release notes

### LINKED ISSUE #1223: Add Git tag/release message as backup to release notes
Extends on #1213 

Currently we use the release notes or we show a generic unavailable message. It would be good to have a backup to no release notes which would use the git tag message instead (still falling back to the generic message if neither exist)

See https://github.com/pusher/pusher-http-swift/releases as a terrible example of this (I'm sure others have better usage of this)

Sherlouk:
Requires #1213 first
Closes #1223 

Spoke to Dave briefly about this on Discord.

Some repositories generate releases via GitHub but do not provide release notes. In this case, GitHub will often fallback to the message of the underlying tag.

An example of this would be https://github.com/Snowy1803/PulsrMarkdown/releases/tag/v0.1 which currently shows "There are no release notes" with what is deployed on 1213.

### DIFF
--- a/Sources/App/Core/Github.swift
+++ b/Sources/App/Core/Github.swift
@@ -240,6 +240,7 @@ extension Github {
                     forkCount
                     isArchived
                     isFork
+                    isInOrganization
                     licenseInfo {
                       name
                       key
@@ -275,6 +276,9 @@ extension Github {
                         publishedAt
                         tagName
                         url
+                        tagCommit {
+                          message
+                        }
                       }
                     }
                     repositoryTopics(first: 20) {
@@ -286,7 +290,6 @@ extension Github {
                       }
                     }
                     stargazerCount
-                    isInOrganization
                   }
                 }
                 """)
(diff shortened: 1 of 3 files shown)


Sherlouk:
Sounds like the vibe for this has died, so going to bin it off for now

daveverwer:
Nope! I think there's good stuff here. I'm going to restore the branch for now as I don't think it's dead at all.

finestructure (approved):

finestructure:
LGTM!

daveverwer (commented):
Apologies for letting this languish for so long, James!

One concern I had with this was that we would be treating what a developer might have written as something that they think will never be seen by anyone and re-purposing it as a public release note.

I know that, technically, all open-source commit messages are available for viewing. Still, there's a difference between having to look in GitHub's commits view or run `git log` and having it aggregated by another web page as a public release note.

We could prefix the notes to explain they come from commit messages or even aggregate all the commit messages for a release, but I still think it changes the expectation of how a commit message will be used.

I think, sadly, we'll close this. Apologies again for leaving it to sit for such a long time!

finestructure:
I actually checked this earlier and it didn't seem to do that. The problem was that I only looked at the release page without clicking the ... buttons.

This is the case when it's only tags, not releases. And it doesn't seem to fill in the commit consistently.

Notice how 2.44.0 has no message:

image

🤷

PR #1225, opened 2021-07-09
Add release notes count badge to tab

Sherlouk:
Marking this one as a draft purely on the basis that I'm not even sure this is **desired**, let alone implemented correctly.

Probably one for Dave to make a call on. Close PR/delete branch if it's not something we care about.

One thing to note is that this is driven by GitHub release count (what the tab actually renders) as opposed to the release reference count shown elsewhere on the package page (which is based on any tags (?)) so it's plausible this number will be different. Having said that, if we showed tags on the release notes page it could make sense to remove this extra field in the DB.

⚠️ Schema change.

Branched off #1224 which is off #1213.

Captured 2021-07-09 at 4 27 29


Sherlouk:
Going to close this off - it feels too confusing when compared to the other release count number (which is based on tags). 'Why are they different?' - some user

PR #1227, opened 2021-07-12
Explore avoiding mock builder

finestructure:
Explored avoiding `MockPredicateBuilder`. I typically try to avoid mock objects as I find it really easy to end up write a "null" test where you're injecting assumptions on both sides and rather use the live objects where possible.

It's possible here but I'm not sure I love how it makes the SQL assertions a bit longer (and weirder).

What's worth keeping is moving `matchesPackageFilters` to an extension on `SQLSelectBuilder.

### DIFF
--- a/Sources/App/Core/Extensions/SQLSelectBuilder+ext.swift
+++ b/Sources/App/Core/Extensions/SQLSelectBuilder+ext.swift
@@ -19,4 +19,12 @@ extension SQLSelectBuilder {
     func column(_ expression: SQLExpression, as alias: String) -> Self {
         column(SQLAlias(expression, as: SQLIdentifier(alias)))
     }
+
+    func `where`(searchFilters: [SearchFilter]) -> Self {
+        self.where(group: { builder in
+            searchFilters
+                .prefix(20) // just to impose some form of limit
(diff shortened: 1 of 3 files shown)


Sherlouk:
Happy with the changes here 👍  If you want to rebase and push them down into my branch then that'd be grand, I'll pick it up again hopefully tonight to get it over the hill!

Sources/App/Core/Extensions/SQLSelectBuilder+ext.swift
```diff
@@ -19,4 +19,12 @@ extension SQLSelectBuilder {
     func column(_ expression: SQLExpression, as alias: String) -> Self {
         column(SQLAlias(expression, as: SQLIdentifier(alias)))
     }
+
+    func `where`(searchFilters: [SearchFilter]) -> Self {
+        self.where(group: { builder in
+            searchFilters
+                .prefix(20) // just to impose some form of limit
+                .reduce(builder) { $1.query($0) }
```
Sherlouk, line 27: Looks like the tests aren't passing, most likely because I renamed `query` to `where` based on your other review comment 😄

Sherlouk (approved):

PR #1230, opened 2021-07-13
Use extension to detect current environment

Sherlouk:
We use this little line of code in a handful of places, thought it'd be nice to wrap it in a quick extension and use that instead.


daveverwer (approved):
Yes yes yes!

PR #1234, opened 2021-08-09
sort package info alphabetically

### LINKED ISSUE #1163: Order packages on author's page alphabetically
combine-schedulers should be first etc

CleanShot 2021-06-09 at 08 56 33@2x

selichter:
Fixes #1163


daveverwer:
Thank you @selichter! I'll check this out and test it later 🚀

selichter:
> Thank you @selichter! I'll check this out and test it later 🚀

No rush on my end. I actually meant to mark the pr as a draft and forgot. Some of the styling is not quite right yet (using a background image in css to display the star isn't working for some reason?) and I assume I've broken a snapshot test that I haven't updated yet.

daveverwer:
> No rush on my end. I actually meant to mark the pr as a draft and forgot. Some of the styling is not quite right yet (using a background image in css to display the star isn't working for some reason?) and I assume I've broken a snapshot test that I haven't updated yet.

Oh, that's perfect then. I'll not test this by running it yet.

I did have a quick look through the code and everything seems good from a first glance, so I don't anticipate any major amendments to this in review! 🚀

Resources/Styles/package_list.scss
```diff
@@ -34,7 +34,15 @@
       p {
         margin: 5px 0;
         color: var(--page-text);
+        
+        &.stars {
```
selichter, line 38: @daveverwer can you take a look at the styling on this? It's never taking effect or getting compiled into the `main.css` when I run everything locally and I'm confused about what could be causing that.

selichter: Screen Shot 2021-08-11 at 8 57 25 PM
^^That is what I'm ending up with locally.

daveverwer: It's possible you're not running `make serve-front-end` as you develop.

The CSS and JavaScript only get compiled from the SASS and multi-file JS sources when Rollup processes them and minifies them.

`make build-front-end` does a one-time build into the correct directory and is what we run during deployment to make sure that the live CSS is always correct. `make serve-front-end` is what you should run locally, and it also watches the source files for changes and re-compiles whenever it needs to.

If it's not that, let me know and we can investigate more!

selichter: That's definitely it, but I didn't realize it as I was getting a styled site with just a `make run`. I'm pretty sure I ran `make serve front-end` on my first pr, so then I must have had a `main.css` hanging around from that somehow...

daveverwer: > I must have had a `main.css` hanging around from that somehow...

Yep.

It's not ideal, but it's the best we have come up with so far. The compiled `main.css`/`main.js` and associated files are not checked into source code, which is beneficial so you only need to commit the `.scss` files and source JavaScript files, not the minified ones.

The downside is that you can have an old version of those files hanging around for weeks without noticing it as it's not tracked by git.

Build phases for `Package.swift` files would save us here, but they're not quite here yet.

selichter:
@daveverwer this should be good for review whenever you have a minute.

daveverwer:
Closing and re-opening to kickstart CI

daveverwer:
Thanks Sarah! I'll review this properly today 👍

Sources/App/Views/PackageInfo.swift
```diff
@@ -16,13 +16,15 @@ struct PackageInfo {
...
+    var stars: Int
 }
 
 extension PackageInfo {
     init?(package: Package) {
         guard let repoName = package.repository?.name,
               let repoDescription = package.repository?.summary,
-              let repoOwner = package.repository?.owner
+              let repoOwner = package.repository?.owner,
+              let repoStars = package.repository?.stars
```
daveverwer, line 27: This is an interesting one that @finestructure will have an opinion on. He just removed the `repoDescription` code from line 25 here as it was causing packages to disappear from the index if they did not have a description. I think this may now introduce that same problem for packages that we've not managed to get stars for?

Just like the change Sven made in #1239, we could just use 0 if stars does not have a value.

I'd love your input on this @finestructure

Resources/Styles/package_list.scss
```diff
@@ -22,9 +22,9 @@
   li {
     background-color: var(--package-list-background);
     border-left: 3px var(--package-list-border) solid;
+    display: grid;
```
daveverwer, line 25: I don't think this line is necessary. The `a` in the `two_column` class which you have as a child of this is a grid which is what you're using to position the text. This may be left over from an earlier attempt?

Resources/Styles/package_list.scss
```diff
@@ -22,9 +22,9 @@
...
 
     a {
-      display: block;
       padding: 10px;
```
daveverwer, line 28: I've added an `align-items: center;` here to bring the display of the star information to be vertically centered.

daveverwer (commented):
This is looking really good Sarah! Thank you for your work on it.

Left some comments inline. I think the only other change I'd make is to use the `--image-star` SVG rather than the ★ character as text. Even though at this size, they look the same, it'll mean that if we ever change the design of that star image it gets updated in both places.

daveverwer:
I pushed a couple of changes onto the local copy of this @selichter, but notice that the PR is off your fork so you'll need to pull my changes across to your copy.

selichter:
Screen Shot 2021-08-17 at 9 25 08 PM

Sources/App/Views/PackageInfo.swift
finestructure: Yes, we'd want to substitute in 0 here (as I see Sarah has already done now 👍)

daveverwer (commented):
Thanks Sarah, this is looking great!

Unfortunately, the way that the image is being positioned isn't quite right when a package has lots of stars.

Screenshot 2021-08-18 at 09 22 22@2x

I think what you'll want to play with here is putting the stars text a span (which may need to be inside the `p` still) and then positioning the image as you have next to that. Then have the span align inside its container.

It may be possible to do this without a container `p`, but I'd need to play with it to figure that out. I'm happy to discuss this in more detail, please let me know if you'd like me to mock up what I mean!

selichter:
This one should do it. 
Screen Shot 2021-08-22 at 10 57 41 PM

Sources/App/Views/Author/AuthorShow+View.swift
```diff
@@ -67,8 +76,15 @@ enum AuthorShow {
...
                                     .href(package.url),
-                                    .h4(.text(package.title)),
-                                    .p(.text(package.description))
+                                    .class("two_column"),
+                                    .div(
+                                        .h4(.text(package.title)),
+                                        .p(.text(package.description))
+                                    ),
+                                    .p(
+                                        .span(.id("star_text")),
```
daveverwer, line 85: This is looking great, but I'd just make one small change. Elements with an `id` should be unique, like the `package_list` one on the `ul` above. There's only ever one of those elements in the document, so it's a candidate for an `id`. This span is repeated multiple times, so should be a `class`.

```suggestion
                                        .class("stars"),
```

Also, I had a quick play and even though I suggested it, we can get away without that extra `span`. See the CSS suggestion I'm making along with this one for a solution.

Resources/Styles/package_list.scss
```diff
@@ -35,10 +35,22 @@
...
     }
   }
+
+  #star_text {
+    background-image: var(--image-star);
+    background-repeat: no-repeat;
+    background-size: 30%;
+    background-position: center;
+    padding: 25px;
+  }
```
daveverwer, line 52: Updated to match the suggestion to use a `class` instead of an `id`. I also made the size of the image explicit rather than percentage-based, in case the size of the `p` ever changes.

```suggestion
  .stars {
    padding-left: 25px;
    background-size: 18px;
    background-repeat: no-repeat;
    background-position: top 2px left;
    background-image: var(--image-star);
  }
```

daveverwer (approved):
This is looking great @selichter! I've made a couple of minor suggestions on the HTML and CSS. After those, I think this is good to go!

If you're happy with it. Could you close this PR and open a new one based on a branch from this repository rather than your fork, please? I'll need to push a new set of image snapshots to it and I can't do that at the moment.

**Note:** For the future, it's no problem at all to create branches directly on this project, you have the permissions to do that and it makes collaborating on them much easier.

selichter:
Closing this in favor of a branch made off this repo, not a fork.

PR #1243, opened 2021-09-06
Make the 404 error page more generic.

### LINKED ISSUE #1219: Keyword route error page
We use the same 404 error page on any route which cannot be found.

I think it would be nice, for some routes, to offer a more helpful message.

Take this keyboard route: https://staging.swiftpackageindex.com/keywords/james-is-awesome

>Something went wrong. Sorry!
>404 - Not Found
>
>If you were expecting to see a page here, the site might be in the process of re-indexing this package. Please try again in an hour or two.

I think this would be nicer if this instead returned something like

> No packages were found with the keyword 'james-is-awesome'.

daveverwer:
Fixes #1219 

Also made the API a little more flexible so error messages can pass back a full `Node` instead of just a string.

No rush for a review on this @finestructure. It's not urgent.


finestructure (approved):

daveverwer:
Deploying as 2.43.3

PR #1248, opened 2021-09-17
1163 order author page packages

selichter:
- Sort an author's packages alphabetically
- Display the number of github stars each package has


selichter:
@daveverwer This branch should be good for you to review + update.

daveverwer (commented):
Thanks so much for moving this onto a project branch @selichter! It's much easier to work with here.

So, there was a small typo in the HTML with the wrong class name that I just pushed a fix for and I was just about to approve it when I noticed the keywords and search page snapshot tests were changing, and not in an expected way.

I think because the `a` in the `package_list` is now a `two_column`, it's breaking the keyword list and the search results list.

I'm not 100% clear on the best way to fix this tonight, but it's going to need a little more CSS to fix this:

Screenshot 2021-09-17 at 23 47 35@2x

See how all the padding is gone from the left of the item, and the bounding box for the `a` tag has moved *up* above the link for some reason. It's very strange!

I've committed the snapshots that are correct but left the ones that are broken. If you'd like, I can have a look at how to fix the CSS, or you can. I don't want to push more work on you!

daveverwer:
I couldn't let it rest and just figured it out. The removal of the `display: block` on line 27 of package_list.scss is causing it. So, that needs to be set to `block` to display correctly in the keywords and search results (where we don't have a two-column layout) but if it's set to `block` in the two-column layout, then of course it pushes the stars down under the package information.

That two-column layout needs a bit more work. Do you want to take a look at it?

I've pushed the change to reinstate the necessary `block` line in 15865b728a9aa7ed40711735a8f24bca3b77c423, but that breaks the `two_column`.

daveverwer:
Hey @selichter. I have an upcoming change I'll need to make on this page for a different reason. Would it be OK if I fixed up the CSS and merged this in so we don't get a difficult merge conflict?

PR #1262, opened 2021-09-28
Dependency resolution target

finestructure:
This places `getResolvedDependencies` in its own module so it can be reused in our builder.

The function isn't complex or likely to change often, so it could just as well be copied over. The more important thing to share is probably the `ResolvedDependency` struct, which might evolve. But that's also not a huge maintenance burden to keep aligned. We're doing this already with the `BuildResult` struct that needs to match the server's API input.

I'm on the fence if this is worth it. I'd probably be clearly in favour if it didn't require the introduction of the `DepedencyResolution.FileManager` protocol to keep the mocking functionality, which makes this slightly clunky as a module.

The builder PR this would enable is here: https://gitlab.com/finestructure/swiftpackageindex-builder/-/merge_requests/75/diffs

What do you think?

Todo:

- [x] decide if it's worth it to pull this into its own product
- [ ] add handling of additional data in `API.PostCreateBuildDTO` (probably in separate PR though)
- [ ] if we're pulling common things out, `API.PostCreateBuildDTO` should also be shared (it needs to be the same on both sides, it's `API.BuildResult` on the builder side. Wouldn't be part of the `DependencyResolution` module and in a follow-up PR.


daveverwer (approved):
Anything that we don't need to keep manually synced is probably worth it!

PR #1265, opened 2021-09-29
Pull the package name from the default version

### LINKED ISSUE #1259: Keyword Result Name Mismatch
https://swiftpackageindex.com/RougeWare/Swift-Simple-Logging

The name of this package appears differently in standard search results and recent packages vs keyboard results.

Probably just the wrong var being used

image
image
image

daveverwer:
Fixes #1259 

There will be a performance impact with this fix, but it really shouldn't be too bad.

There is a lack of tests for the pieces of the app that this covers. No explicit tests for `PackageInfo` and the `AuthorController` has some minimal tests, but nothing that tests the logic really. There's an argument for adding better tests for both of those things, but I'm not surethis PR is the right place for that.

### DIFF
--- a/Sources/App/Controllers/AuthorController.swift
+++ b/Sources/App/Controllers/AuthorController.swift
@@ -19,7 +19,7 @@ import Vapor
 
 struct AuthorController {
 
-    private func query(on database: Database, owner: String) -> EventLoopFuture<[Package]> {
+    static func query(on database: Database, owner: String) -> EventLoopFuture<[Package]> {
         Package.query(on: database)
             .with(\.$repositories)
             .with(\.$versions)
@@ -44,7 +44,7 @@ struct AuthorController {
             return req.eventLoop.future(error: Abort(.notFound))
         }
 
-        return query(on: req.db, owner: owner)
+        return Self.query(on: req.db, owner: owner)
             .map {
                 AuthorShow.Model(
                     owner: $0.first?.repository?.owner ?? owner,
--- a/Sources/App/Controllers/AuthorController.swift
+++ b/Sources/App/Controllers/AuthorController.swift
@@ -22,6 +22,7 @@ struct AuthorController {
     private func query(on database: Database, owner: String) -> EventLoopFuture<[Package]> {
         Package.query(on: database)
             .with(\.$repositories)
+            .with(\.$versions)
             .join(Repository.self, on: \Repository.$package.$id == \Package.$id)
             .filter(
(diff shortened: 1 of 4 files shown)


Sources/App/Controllers/AuthorController.swift
```diff
@@ -22,6 +22,7 @@ struct AuthorController {
     private func query(on database: Database, owner: String) -> EventLoopFuture<[Package]> {
         Package.query(on: database)
             .with(\.$repositories)
+            .with(\.$versions)
```
finestructure, line 25: This is certainly the safest thing to do, as you say at the slight (?) expense of pulling in all the versions.

We _could_ limit the impact by doing this:

```swift
        Package.query(on: database)
            .with(\.$repositories)
            .join(Repository.self, on: \Repository.$package.$id == \Package.$id)
            .join(Version.self, on: \Version.$package.$id == \Package.$id)
            .filter(
                DatabaseQuery.Field.path(Repository.path(for: \.$owner), schema: Repository.schema),
                DatabaseQuery.Filter.Method.custom("ilike"),
                DatabaseQuery.Value.bind(owner)
            )
            .filter(Version.self, \.$latest == .defaultBranch)
            .all()
```

which _should_ also populate the `.$versions` property we use in `package.name()` but prune the versions to just the one we need.

The problem is that we're tying that query to the implementation of `name()`. This is not a unique problem tbh, there are a few places were we have to write a query in ways that'll make subsequent accessors work properly. I think the only remedy is tests, really.

Don't we have a test that we could copy and adapt for this scenario? I think it'd be a right thing to do while we're at it.

finestructure: This wasn't quite correct. The `.with(\.$versions)` is still needed but with that in the SQL then becomes:

```sql
SELECT packages.*, repositories.*, versions.* 
FROM "packages" 
INNER JOIN "repositories" ON "packages"."id" = "repositories"."package_id" 
INNER JOIN "versions" ON "packages"."id" = "versions"."package_id" 
WHERE "repositories"."owner" ilike $1 
AND "versions"."latest" = $2 ["owner", Optional(default_branch)]
```

finestructure: I'll open a separate PR to address the join.

I've also added some tests, which I'll add via a PR into this branch now.

daveverwer: This is really interesting. I had no idea it worked like that. We certainly have other instances in the code where we're doing similar things to this that we might want to look at.

Thanks for looking at it, Sven!

PR #1276, opened 2021-10-06
Efficiency with joins

finestructure:


Sources/App/Controllers/AuthorController.swift
```diff
@@ -19,22 +19,24 @@ import Vapor
...
-    static func query(on database: Database, owner: String) -> EventLoopFuture<[Package]> {
-        Package.query(on: database)
-            .with(\.$repositories)
-            .with(\.$versions)
-            .join(Repository.self, on: \Repository.$package.$id == \Package.$id)
+    static func query(on database: Database, owner: String) -> EventLoopFuture<[Joined<Package, Repository, Version>]> {
+        Joined<Package, Repository, Version>
+            .query(on: database,
+                   \Repository.$package.$id == \Package.$id,
+                   \Version.$package.$id == \Package.$id)
```
finestructure, line 26: TODO: need to at least allow specifying `OUTER` or `RIGHT` joins or something, because as is (defaulting to `INNER`), we lose packages from the result set if they have no version or repository.

That's perhaps not possible or a problematic state anyway but I wouldn't want to introduce that restriction unnecessarily.

PR #1285, opened 2021-10-10
Release tab navigation

### LINKED ISSUE #1283: Release tweets should link to releases
I do use our releases tweets quite a bit (for example, this morning there’s a new Vapor release) and every time I click on the link it confuses me that I don’t see the release notes.

The reason we didn’t link to the releases was that we want to link to an SPI page from the tweet and not to Github. I think the better solution would be to create a “Release” page in SPI and link to that.

daveverwer:
Merge after #1284 
Fixes #1283 

I was halfway through replying to this issue and figured I'd just implement what I was suggesting 🚀

This isn't *ideal* as it relies on a `setTimeout` to scroll the releases/readme section to the top of the window, but for as long as those as loaded asynchronously this is going to be a problem. If the tab hasn't loaded within 500ms, it degrades gracefully.

### DIFF
--- a/Resources/Scripts/tab_bar_element.js
+++ b/Resources/Scripts/tab_bar_element.js
@@ -51,7 +51,7 @@ export class SPITabBarElement extends HTMLElement {
         // tab selected and focused, but scrolled into the middle of the screen.
         setTimeout((event) => {
           tabLinkElement.scrollIntoView(true)
-        }, 500)
+        }, 1000)
       }
 
       // Add click listener which will show the correct page when a user taps on a tab link


daveverwer:
As soon as CI has run, I'll deploy this branch to staging so you can test it @finestructure.

daveverwer:
I adjusted the timeout and it's much more reliable for me. This is ready to go, if you like it.

daveverwer:
That'll serve me right for trying to make a quick enhancement on a Sunday 😂

daveverwer:
> It would be nice if we could also/subsequently link to a proper page though that doesn’t rely on JS. For instance on the iPad I’ve had cases where it takes a reload to switch to the “Release Notes”.

Can you reproduce this? I tried a little earlier on an iPad and can't.

I think if we have a bug, we should fix it rather than re-work the feature.

finestructure:
> Can you reproduce this? I tried a little earlier on an iPad and can't.

Yes, although it's probably not a common case: Navigate to a package and then add `#releases` to the url and hit enter.

PR #1290, opened 2021-10-12
Prevent `EXC_BAD_ACCESS (code=EXC_I386_GPFLT)` when running snapshot tests

daveverwer:
This is not for merging, just for discussion.

Some of the `PackageShowModelTests` tests are now crashing. I'm not sure how many as the crash stops subsequent tests from running. 

I can't explain why these tests passed yesterday and fail today when there have been no package updates to Plot or SnapshotTesting since then. I also can't explain why changing the rendering style stops them from crashing.

Finally. I can't explain the snapshot diffs in 1347283eaa95c7c3c3a2ed31b4cfd4c9f073ec2a. I just can not explain them.

I don't like anything about this.


finestructure:
FWIW, the snapshot tests in `PackageShowModelTests` and `WebpageSnapshotTests` run (but `WebpageSnapshotTests` don't pass on my machine, obviously, `PackageShowModelTests` do) for `main` and all the revisions in the branch without crashing 🤷 

macOS 11.6 (20G165)
Xcode 13.0 (13A233)

PR #1295, opened 2021-10-14
Fully typed, generic query results container

finestructure:
This is a fully typed, generic container to express all our queries and their results. With the building blocks we have we can express

- a joined model of cardinality 2 and 3 (easily extensible to more)
- a tree of referenced models that are loaded via relationships, both for one and two child relationships (easily extensible to more)

The test `test_activity` that I've updated demonstrates that this actually works. The type we're using is `Ref<Joined<Package, Repository>, Ref2<Version, Build, Product>>` - which is exactly what `JPRVB` is in all its glory.

The way you read this is that there's a parent model that's a join between `Package` and `Repository`, which has a fulfilled child relationship on `Version`, which in turn has _two_ fulfilled child relationships on `Build` and `Product`:

```
(Package - Repository) -< Version
                             |
                             |-< Build
                             |
                             '-< Product
```

This gives us the following naming options:

1. `JPRVB` (actually `JPRVBP`)
1. `JoinedPackageRepositoryVersionBuild` (actually `JoinedPackageRepositoryVersionBuildProduct`)
1. `Ref<Joined<Package, Repository>, Ref2<Version, Build, Product>>`

My current order of preference is "none of these", 1, 3, 2, and there's quite a gap between "none" and 1.

Part of the problem with the name is that we've got a lot of types involved and we do use a number of different combinations in our results making it very easy to arrive at ambiguous names if we leave any of the types involved out.

I'll need to let it sink in but maybe we should name the type after what's using it instead of what it is. I.e. currently `JPRVB` is being used in `API.PackageController`. Maybe there's a name `PackageController.PackageResult` that we type alias the `Ref<...>` monster to.

Even if different parts use the same `Ref<...>` we still give them distinct names. There's no problem having the same type available under two different aliases. There's no risk of confusion and the types aren't exchanged between components. And where they are, we'll surely find a top level component where the type alias could live.

I should add that we could also just rename `JPRVB` to `PackageController.PackageResult` and create copies of that struct when we need those types elsewhere. It's a bit of duplication but avoids the `Ref` monster.

Something to think about.

### DIFF
--- a/Sources/App/Controllers/PackageController.swift
+++ b/Sources/App/Controllers/PackageController.swift
@@ -114,6 +114,11 @@ private func fetchReadme(client: Client, package: Joined<Package, Repository>) -
 
 
 extension PackageController {
+    //    (Package - Repository) -< Version
+    //                                 |
+    //                                 |-< Build
+    //                                 |
+    //                                 '-< Product
     typealias PackageResult = Ref<Joined<Package, Repository>, Ref2<Version, Build, Product>>
 }
 
--- a/Sources/App/Core/Ref.swift
+++ b/Sources/App/Core/Ref.swift
@@ -25,6 +25,21 @@ extension Product: Referencable {}
 extension Version: Referencable {}
 
 
+/// `Ref` and `Ref2`, together with `Joined` allow us to define typed query results
+/// that encode the query structure in the type. This is important to avoid
+/// triggering fatal errors when accessing relationships that have not been
+/// loaded. Result types based in these containers can expose accessors that are
+/// safe.
+/// An example use case is `PackageController.PackageResult`, which is a
+/// typealias of `Ref<Joined<Package, Repository>, Ref2<Version, Build, Product>>`,
+/// representing the following query result:
+/// ```
+/// (Package - Repository) -< Version
+///                              |
+///                              |-< Build
+///                              |
+///                              '-< Product
+/// ```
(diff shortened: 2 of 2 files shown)


daveverwer:
In terms of the options you've defined above, my preference would be 2, 3, then 1.

However, I like the idea of aliasing the types for controllers and I wonder if a combination of options 2 and 3 with that suggestion might work.

We could have a `JoinedPackageRepositoryVersionBuildProduct` which is a verbose simplification of `Ref<Joined<Package, Repository>, Ref2<Version, Build, Product>>` (even if it does lose some subtlety of what's happening) that could then be aliased to one or more controller specific types, which are far easier to use in controllers.

That way, we'd only have one definition of a `JoinedPackageRepositoryVersionBuildProduct` rather than duplicating definitions of `Ref<Joined<Package, Repository>, Ref2<Version, Build, Product>>`, which are harder to rationalise at a glance. We'd then have controller specific aliases of `JoinedPackageRepositoryVersionBuildProduct`.

daveverwer:
> ```
> (Package - Repository) -< Version
>                              |
>                              |-< Build
>                              |
>                              '-< Product
> ```

Also, these diagrams should be in the source next to the definition of the `JoinedPackageRepositoryVersionBuildProduct` types.

finestructure:
I might be better to complete review on #1292 and merge it into `main` instead of merging this PR here into #1292, making it even bigger and messing with our partial review.

PR #1328, opened 2021-11-03
Sign package collections

finestructure:
Just the basic signing POC at the moment to ensure this works and is testable.

NB: the enclosed certificate and PEM _do not_ create valid signed collections, although the signing itself produces a signed collection.

When importing into Xcode it is rejected with this beautiful dialog :) (as it should be with this key/cert pair):

CleanShot 2021-11-03 at 10 51 40@2x

- [x] set up signing key on dev
- [x] set up signing key on prod
- [x] deploy branch to dev and verify signature in Xcode
- [x] add test with revoked key

Deploy as 2.48.0 due to the env variable setup requirement.


.github/workflows/ci.yml
```diff
@@ -21,7 +21,7 @@ jobs:
     name: Test
     runs-on: ubuntu-latest
     container:
-      image: registry.gitlab.com/finestructure/spi-base:0.6.2
+      image: registry.gitlab.com/finestructure/spi-base:22eec776e982b298334fd9ed809fa2c78d0125b1
```
finestructure, line 24: This is referencing a new base image with sqlite3 libs, required to build the package collection module of `SwiftPM`.

Package.resolved
```diff
@@ -255,10 +255,10 @@
       },
       {
         "package": "SwiftPM",
-        "repositoryURL": "https://github.com/apple/swift-package-manager.git",
+        "repositoryURL": "https://github.com/finestructure/swift-package-manager.git",
         "state": {
-          "branch": "swift-DEVELOPMENT-SNAPSHOT-2021-05-04-a",
-          "revision": "0c9c8d55889405f39b093ba4868db375b5b7e16c",
+          "branch": "signer-default-observability-scope",
+          "revision": "25df5347eef59c5f0ae86f5e4ee7723ff4eb78f7",
```
finestructure, line 261: My branch contains a workaround for accessing `PackageCollectionSigning`. I've opened a PR against SPM here: https://github.com/apple/swift-package-manager/pull/3836

Package.swift
```diff
@@ -39,8 +39,7 @@ let package = Package(
         .package(url: "https://github.com/handya/OhhAuth.git", from: "1.4.0"),
         .package(url: "https://github.com/scinfu/SwiftSoup.git", from: "2.3.2"),
         .package(name: "SwiftPM",
-                 url: "https://github.com/apple/swift-package-manager.git",
-                 .revision("swift-DEVELOPMENT-SNAPSHOT-2021-05-04-a"))
+                 url: "https://github.com/finestructure/swift-package-manager.git", .branch("signer-default-observability-scope"))
```
finestructure, line 42: My branch contains a workaround for accessing `PackageCollectionSigning`. I've opened a PR against SPM here: https://github.com/apple/swift-package-manager/pull/3836

daveverwer (approved):

finestructure:
Great, works on macOS, fails on Linux :/

```
/__w/SwiftPackageIndex-Server/SwiftPackageIndex-Server/Tests/AppTests/PackageCollectionTests.swift:824: error: PackageCollectionTests.test_sign_collection : failed - The operation could not be completed. (PackageCollectionsSigning.PackageCollectionSigningError error 3.)
```

Tests/AppTests/PackageCollectionTests.swift
```diff
@@ -769,3 +772,120 @@ class PackageCollectionTests: AppTestCase {
         )
     }
 }
+
+
+class SigningTests: XCTestCase {
+
+    func test_sign_collection() throws {
+        try XCTSkipIf(!isRunningInCI && Current.collectionSigningPrivateKey() == nil, "Skip test for local user due to unset COLLECTION_SIGNING_PRIVATE_KEY env variable")
```
finestructure, line 780: We don't want this to fail for someone running the tests locally if they haven't got the private key set up, so we skip in that case. But we _don't_ skip if it's running in CI, in which case the key must be set up and the test should be run.

daveverwer (approved):

Package.swift
```diff
@@ -55,7 +54,7 @@ let package = Package(
             "SwiftPrometheus",
             "OhhAuth",
             "SwiftSoup",
-            .product(name: "PackageCollectionsModel", package: "SwiftPM")
+            .product(name: "SwiftPMPackageCollections", package: "SwiftPM")
```
finestructure, line 57: Unfortunately, this bring with it a warning. I'll try to get rid of it.

CleanShot 2021-11-08 at 07 22 35@2x

Sources/App/Controllers/API/API+PackageCollectionController.swift
```diff
@@ -20,10 +20,10 @@ extension API {
 
     struct PackageCollectionController {
 
-        func generate(req: Request) throws -> EventLoopFuture<PackageCollection> {
+        func generate(req: Request) throws -> EventLoopFuture<SignedCollection> {
```
finestructure, line 23: I'm just changing collections over to signed. I was considering making this controllable via an env variable so we could switch signing on and off easily. However, the return type needs to be one or the other.

What we can do is make it `Data` or `String`, i.e. pre-render inside the method. The result for the consumer would be exactly the same. I just didn't want to have to deal with the content-type headers etc so I've stuck with the simple solution which we know will work.

Tests/AppTests/Util+ext.swift
```diff
@@ -27,6 +27,10 @@ extension XCTestCase {
         ProcessInfo.processInfo.environment.keys.contains("GITHUB_WORKFLOW")
     }
     
+    var runImageSnapshotTests: Bool {
+        ProcessInfo.processInfo.environment.keys.contains("RUN_IMAGE_SNAPSHOT_TESTS")
+    }
```
finestructure, line 32: We were previously using `GITHUB_WORKFLOW` as a signal not to run the snapshot image tests. Because we're now also checking `isRunningInCI` for the signing tests, it's not possible to control both the singing and the snapshot tests independently.

Tests/AppTests/__Snapshots__/PackageCollectionTests/test_sign_collection.1.json
```diff
@@ -0,0 +1,59 @@
...
...KICAgICAgICAgICJkZWZhdWx0VG9vbHNWZXJzaW9uIiA6ICI1LjUiLAogICAgICAgICAgIm1hbmlmZXN0cyIgOiB7CiAgICAgICAgICAgICI1LjUiIDogewogICAgICAgICAgICAgICJwYWNrYWdlTmFtZSIgOiAiZm9vIiwKICAgICAgICAgICAgICAicHJvZHVjdHMiIDogWwogICAgICAgICAgICAgICAgewogICAgICAgICAgICAgICAgICAibmFtZSIgOiAicCIsCiAgICAgICAgICAgICAgICAgICJ0YXJnZXRzIiA6IFsKICAgICAgICAgICAgICAgICAgICAidCIKICAgICAgICAgICAgICAgICAgXSwKICAgICAgICAgICAgICAgICAgInR5cGUiIDogewogICAgICAgICAgICAgICAgICAgICJleGVjdXRhYmxlIiA6IG51bGwKICAgICAgICAgICAgICAgICAgfQogICAgICAgICAgICAgICAgfQogICAgICAgICAgICAgIF0sCiAgICAgICAgICAgICAgInRhcmdldHMiIDogWwogICAgICAgICAgICAgICAgewogICAgICAgICAgICAgICAgICAibmFtZSIgOiAidCIKICAgICAgICAgICAgICAgIH0KICAgICAgICAgICAgICBdLAogICAgICAgICAgICAgICJ0b29sc1ZlcnNpb24iIDogIjUuNSIKICAgICAgICAgICAgfQogICAgICAgICAgfSwKICAgICAgICAgICJ2ZXJzaW9uIiA6ICIxLjIuMyIKICAgICAgICB9CiAgICAgIF0KICAgIH0KICBdCn0.isrE2T3a5C6FFqx3Xtqo_We6ftiLolUJisToW1jsn9YPXDZQEdAfXcovtmN8YY_uIrNXt-mEe7gpNkOSSKW98NHOnXeNnF3eFA5Wu6Av-l7_BBUvvAheGY1ab1HOIa6ZrLez9L6BLPckr1vMegjyjwLoYnWXHaQ7qAXTTuzArf31vP5wekoLuoS5gfRWkD6OmFcpbk6zHum-x4xLTilHuqOwFL7vPD-kaCsVdI1ayc7mySialqswwo9ius0q9A4kpQMNUDQ3pRYVZjvaeDr9AhqPpxE-uTTmOC2FOtn3mPnImkCT9FiBgd5xX4a--QdhQQ84vyXQEQBiN417lS7TEQ"
```
finestructure, line 57: Something I found out while reading the signing code: This is a [JWT encoded payload + signature](https://github.com/apple/swift-package-manager/blob/signer-default-observability-scope/Sources/PackageCollectionsSigning/Signing/Signature.swift#L66):

```swift
    static func generate<Payload>(for payload: Payload,
                                  with header: Header,
                                  using signer: MessageSigner,
                                  jsonEncoder: JSONEncoder) throws -> Data where Payload: Encodable {
        let headerData = try jsonEncoder.encode(header)
        let encodedHeader = headerData.base64URLEncodedBytes()

        let payloadData = try jsonEncoder.encode(payload)
        let encodedPayload = payloadData.base64URLEncodedBytes()

        // https://www.rfc-editor.org/rfc/rfc7515.html#section-5.1
        // Signing input: BASE64URL(header) + '.' + BASE64URL(payload)
        let signatureData = try signer.sign(message: encodedHeader + .period + encodedPayload)
        let encodedSignature = signatureData.base64URLEncodedBytes()

        // Result: header.payload.signature
        let bytes = encodedHeader
            + .period
            + encodedPayload
            + .period
            + encodedSignature
        return bytes
    }
```

which means that the whole collection is contained as a base64 encoded duplicate within the signature. For large collections that's going to create unnecessarily big collections.

I suppose the problems are

- if you don't include the plain text JSON the data is useless unless you have infra to decode the signature
- if you don't include the payload in the signature you rely on the json being passed in exactly as is for verification

On the other hand: I'm sure Xcode has to throw away the plain text JSON anyway and only use the encoded one, right? Otherwise it's trivial to pass in mismatched signed/compromised content.

And it seems like it would be very tempting for consumers to treat the signature as authoritative and then use the plain JSON for simplicity. So arguably it might be better not to ship it in the first place.

I feel like we should discuss this with the SPM team.

daveverwer: I think discussing it with them is a good idea, but I think the first question should be to check that the encoded version is used in that way. I thought the signature was checked as a confirmation of the validity of the JSON data, not as something that would be decoded and used.

The other obvious disadvantage of *only* including the encoded version is that it presents a significant barrier to people just wanting to experiment with using collection data. If we decided to only include the signature, we'd immediately be trying to give a URL parameter or different URL location for an unsigned version of the data.

finestructure:
Had some trouble with the final verification: the private env variable was set to `protected` - which means that Gitlab restricts it to default branches and tags. I.e. because I was deploying from this branch it was not being set :)

CleanShot 2021-11-09 at 06 26 49@2x

I'll set it back to protected now after verification.

Tests/AppTests/PackageCollectionTests.swift
```diff
@@ -768,4 +771,98 @@ class PackageCollectionTests: AppTestCase {
...
+    }
+
+    func test_sign_collection_revoked_key() throws {
+        // setup
+        let collection: PackageCollection = .mock
+        // get cert and key and make sure the inputs are valid (apart from being revoked)
+        // so we don't fail for that reason
+        let revokedUrl = fixtureUrl(for: "revoked.cer")
+        XCTAssertTrue(Foundation.FileManager.default.fileExists(atPath: revokedUrl.path))
+        let revokedKey = try XCTUnwrap(fixtureData(for: "revoked.pem"))
```
finestructure, line 803: This is our now revoked key/cert pair from the original signing experiments.

PR #1329, opened 2021-11-04
New snapshots generated on M1

daveverwer:
Hopefully this'll put us both back in sync! 🤞


finestructure:
😞 

CleanShot 2021-11-04 at 14 41 52@2x

finestructure:
Some are green now though. I don't think we've had that before:

CleanShot 2021-11-04 at 14 46 39@2x

Must be the ones where only the banner had an svg on it.

finestructure (approved):

daveverwer:
How frustrating. Did you do a `make build-front-end` first to make sure you're all up to date with css?

finestructure:
ah no, I did not! Let me try that. That might actually be the problem.

Issue #1340, opened 2021-11-10
Clicking on "Copy Package CollectionURL" button triggers collection generation


finestructure:
I just discovered this while looking into instrumentation. Clicking this button:

<img width="874" alt="CleanShot 2021-11-10 at 16 27 25@2x" src="https://user-images.githubusercontent.com/65520/141141820-586f2b63-b00f-4926-9b6c-9049ddd968a4.png">

calls into `PackageCollectionController.generate`. I don't _think_ this should be happening - we should just be putting the string that's already on the page in the text field onto the clipboard, right?

I'm running this locally with a breakpoint in the method and it's hit as soon as I click the button.

daveverwer:
Yea, that shouldn't be calling out to the server at all. I'll investigate and fix

finestructure:
Ah great. I was poking around a bit but you'll find this much faster.

finestructure:
I wonder if this is a Safari (or Safari Tech Preview) thing: I've just _pasted_ the collection url into the browser bar - without hitting return - and the breakpoint's being hit. It seems to be aggressively pre-fetching urls?

finestructure:
I've looked at the requests for a "pre-fetch" and an "actual" GET and it's turned up something interesting:

Safari Tech Preview prefetch:
```
(lldb) po req
GET /apple/collection.json HTTP/1.1
Host: localhost:8080
Upgrade-Insecure-Requests: 1
Accept: text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8
User-Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_11_1) AppleWebKit/601.2.4 (KHTML, like Gecko) Version/9.0.1 Safari/601.2.4 facebookexternalhit/1.1 Facebot Twitterbot/1.0
Accept-Language: en-GB,en;q=0.9
Accept-Encoding: gzip, deflate
Connection: keep-alive
```

Safari Tech Preview actual:
```
(lldb) po req
GET /apple/collection.json HTTP/1.1
Host: localhost:8080
Upgrade-Insecure-Requests: 1
Accept: text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8
User-Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/15.4 Safari/605.1.15
[...]

PR #1346, opened 2021-11-11
Add metrics for package collection GET requests

finestructure:
This will over-count actual requests until we understand how/why #1340 is happening.

I've had a look into whether the request would allow us to tell the pre-fetch from an actual prefetch but it's doesn't look like it'd be reliable. There was a small difference in the user-agent string, which I've detailed in the linked issue.


finestructure:
Will deploy from branch and test on staging first.

finestructure:
Working fine but can confirm it's counting button clicks.

```
# TYPE spi_package_collection_get_total counter
spi_package_collection_get_total 1
```

daveverwer (approved):

PR #1379, opened 2021-11-17
Async await reconcile

finestructure:


daveverwer (approved):

Sources/App/Controllers/API/API+PackageController.swift
```diff
@@ -79,36 +79,37 @@ extension API {
...
                 case .ingest:
-                    return ingest(client: req.application.client,
+                    return try await ingest(client: req.application.client,
                                   database: req.application.db,
                                   logger: req.application.logger,
                                   limit: limit)
                         .map {
                             Command.Response(status: "ok", rows: limit)
                         }
+                        .get()
```
finestructure, line 100: This is brilliant: anywhere we have `EventLoopFuture<T>` as a return type, we can just call `.get()` on an event loop future to convert it to async await.

What this means for the `run` request handler is that even though we're only converting `reconcile` over to be `async`, we can easily make `run` itself `async` without having to also update `ingest` and `analyze`. We keep calling the existing `ingest(...) → EventLoopFuture<Void>` and invoke `.get()` on the result.

This means there's no risk that async conversion might force more and more updates as we get to the less "siloed" parts of the backend.

finestructure: Equally, we could have just convert `reconcile(...) async` back to an `EventLoopFuture` without touching `run(...)`'s signature at all:

```swift
                    let promise = req.eventLoop.makePromise(of: Void.self)
                    promise.completeWithTask {
                        try await reconcile(client: req.client, database: req.db)
                    }
                    return promise.futureResult
                        .flatMap {
                            Package.query(on: req.db).count()
                                .map { Command.Response.init(status: "ok", rows: $0) }
                        }
```

PR #1385, opened 2021-11-20
Consolidate and disarm the defaults in the DB dump scripts

daveverwer:
Rather than have two scripts, we now have one which takes parameters for the host, port, user, and database and runs `pg_dump` from that.

Also, this script now has defaults that dump the prod copy that we've been calling `spi-prod-db-1-dump` so if it's run without the correct env variables, it can't ever take prod down while we resolve the other issue.


daveverwer:
One thing to note here @finestructure: I'm not sure if the updated script runs with fish, I know it works differently with variables.

finestructure:
It’s shebanged to bash so that’s not an issue!

scripts/dump-db.sh
```diff
@@ -14,12 +14,16 @@
...
 echo
 if [[ $REPLY =~ ^[Yy]$ ]]
 then
-  # This command assumes the prod db to be available on port 7432 on localhost, via an ssh tunnel:
-  # ssh -i <your private key> -L 7432:db:5432 -p 2222 root@173.255.229.82
-  PORT=${SPI_PRODUCTION_DB_PORT:-'7432'}
-  pg_dump --no-owner -Fc -h localhost -p "$PORT" -U spi_prod@spi-prod-db-1 spi_prod > spi_prod_$(date +%Y-%m-%d).dump
+  # This command assumes the db will be available on $PORT on localhost, via an ssh tunnel:
+  # For example: ssh -i <your private key> -L 7432:db:5432 -p 2222 root@173.255.229.82
+  pg_dump --no-owner -Fc -h localhost -p ${PORT} -U ${USER}@${HOST} ${DATABASE} > ${HOST}-$(date +%Y-%m-%d).dump
```
finestructure, line 28: Maybe use `USER` for dump file name?

finestructure (approved):

PR #1388, opened 2021-11-23
Batch job duration metrics

### LINKED ISSUE #1381: Add metrics for batch job run times
We currently have to visibility into how long these are running per invocation.

We should adds metrics around this so we can monitor and have historic data if they're getting slower.

finestructure:
This adds duration metrics for our batch jobs.

Fixes #1381

### DIFF
--- a/Sources/App/Commands/Analyze.swift
+++ b/Sources/App/Commands/Analyze.swift
@@ -112,6 +112,7 @@ func analyze(client: Client,
              logger: Logger,
              threadPool: NIOThreadPool,
              mode: AnalyzeCommand.Mode) -> EventLoopFuture<Void> {
+    let start = DispatchTime.now().uptimeNanoseconds
     switch mode {
         case .id(let id):
             logger.info("Analyzing (id: \(id)) ...")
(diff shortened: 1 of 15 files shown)


daveverwer (approved):
If I understand this correctly, this will give us individual metrics for every reconciliation/analysis/ingestion job as well as for each build trigger (not the build itself, just the execution of the trigger?). Is that right?

How are we going to visualise this? As averages or as a list of recent tasks?

finestructure:
These will be part of the general metrics available via prometheus. We can build all sorts of dashboards off of them - averages, min/max, quantiles – but even just looking at the values in the explorer in Grafana will be instructive.

I'll deploy this from the branch and take a screenshot.

finestructure:
> If I understand this correctly, this will give us individual metrics for every reconciliation/analysis/ingestion job as well as for each build trigger (not the build itself, just the execution of the trigger?). Is that right?

For each command invocation, i.e. what our batch jobs are doing, yes. It'll reflect the batch sizes and the time it takes for them to run their candidate selection.

Regarding the build trigger, it's only the part that sends the post requests to Gitlab, yes.

finestructure:
CleanShot 2021-11-23 at 15 08 11@2x

Issue #1303, opened 2021-10-19
Add more metadata to package lists


daveverwer:
Having the number of stars showing against a package is something, but as I [mentioned here](https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/issues/1163#issuecomment-887329608), it's super reductive.

There's a whole load of metadata about a package that can inform which search result or package in a keyword/author view to click on. Until we have a "quality index" style system for packages that would reduce the metadata into one or two scores, we should pull out more information so that decisions are not _only_ based on stars.

Initially, I'd like to add one more field. It should be something along the lines of "last active", which could reduce several dates we have into one which sums up how recently a package has received maintenance.

So, if there has been a commit on the default branch, a release, or some activity from one of the maintainers (not just a new issue being opened, but something that required the maintainer to do something), then we should show something like:

```
owner/package – Active X months/weeks/days ago – 345 stars
```

Something like this:

<img width="501" alt="Screenshot 2021-10-19 at 15 03 07@2x" src="https://user-images.githubusercontent.com/5180/137926109-c269c99e-6259-47bc-a661-d5a171d65726.png">

PR #1389, opened 2021-11-23
Add “last activity” to the package list views

daveverwer:
Progress towards #1303 
Dependent on infrastructure upgrade to Postgres 13.

⚠️ SCHEMA CHANGE ⚠️

Screenshot 2021-11-23 at 13 11 22@2x

in a future PR, I am going to replace the ability to filter on last commit with filtering on this field instead, but this PR is already big enough.


daveverwer:
Tweaked the styling and coped better with mobile:

Screenshot 2021-11-23 at 18 37 12@2x

Screenshot 2021-11-23 at 18 37 22@2x

Sources/App/Models/Repository.swift
```diff
@@ -114,6 +114,9 @@ final class Repository: Model, Content {
     @Field(key: "summary")
     var summary: String?
 
+    @Field(key: "last_activity_at")
+    var lastActivityAt: Date?
+
```
finestructure, line 119: Nit: I've started a (feeble) attempt at ordering the fields alphabetically a while back when we got more than a handful of fields. It's hardly consistent because I started late and some initialisers would be a pain to have their arguments re-ordered. This one we could bump up to after `keywords` though.

Tests/AppTests/ApiTests.swift
```diff
@@ -53,6 +53,7 @@ class ApiTests: AppTestCase {
                        summary: "some package").save(on: app.db).wait()
         try Repository(package: p2,
                        defaultBranch: "main",
+                       lastCommitDate: Date(timeIntervalSinceReferenceDate: 0),
```
finestructure, line 56: Nit: `.t0`?

Tests/AppTests/ApiTests.swift
```diff
@@ -78,6 +79,7 @@ class ApiTests: AppTestCase {
                                   repositoryName: "name 2",
                                   repositoryOwner: "owner 2",
                                   stars: 1234,
+                                  lastActivityAt: Date(timeIntervalSinceReferenceDate: 0),
```
finestructure, line 82: Nit: `.t0`?

Tests/AppTests/Mocks/Search.Result+mock.swift
```diff
@@ -12,12 +12,51 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+import Foundation
```
finestructure, line 15: Nit: Doesn't look like this import is needed, I think Xcode pulled that in when it was `Date(...)` instead of `.t0` below.

Sources/App/Migrations/038/UpdateSearch3.swift
```diff
@@ -0,0 +1,91 @@
...
+              v.package_name
+            FROM packages p
+              JOIN repositories r ON r.package_id = p.id
+              JOIN versions v ON v.package_id = p.id
+            WHERE v.reference ->> 'branch' = r.default_branch
+            """
+
+        return db.raw(dropSQL).run()
+            .flatMap { db.raw(alterRepositoriesTable).run() }
+            .flatMap { db.raw(updateSearchView).run() }
```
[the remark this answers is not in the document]
daveverwer: Yes, this was intentional since they were so closely related and the second was so tied to the first. I'll update it to be two migrations.

Tests/AppTests/__Snapshots__/WebpageSnapshotTests/test_AuthorShow.1.txt
```diff
@@ -83,100 +83,140 @@
...
-                <span class="identifier">owner/name</span> &ndash; 
-                <span class="stars">3</span>
-              </small>
+              <ul class="metadata">
+                <li class="identifier">
+                  <small>owner/name</small>
+                </li>
+                <li class="stars">
+                  <small>3 stars</small>
+                </li>
```
finestructure, line 92: This diff and the corresponding image don't show any last activity info - I'm guessing that's because it's `nil` in the mock data, right? Just confirming, I don't think we need to add that necessarily.

Tests/AppTests/__Snapshots__/WebpageSnapshotTests/test_KeywordShow.1.txt
```diff
@@ -76,100 +76,140 @@
...
-                <span class="identifier">owner/name</span> &ndash; 
-                <span class="stars">4</span>
-              </small>
+              <ul class="metadata">
+                <li class="identifier">
+                  <small>owner/name</small>
+                </li>
+                <li class="stars">
+                  <small>4 stars</small>
+                </li>
```
finestructure, line 85: Dto, regarding missing last activity info (probably nil in mock).

finestructure (approved):
Looks great!

Tests/AppTests/__Snapshots__/WebpageSnapshotTests/test_AuthorShow.1.txt
daveverwer: Actually, I think I will add a date here as a secondary validation that the `packageList` work I did is being used in all these places.

PR #1390, opened 2021-11-23
Add a last activity filter

### LINKED ISSUE #1303: Add more metadata to package lists

daveverwer:
Merge after #1389
Fixes #1303 

I can see an argument to have both of these available, but I think until the feature is finished we should keep things simple. Changing this from last commit to last activity as I think that's a more useful filter parameter.

### DIFF
--- a/Sources/App/Core/SearchFilter.swift
+++ b/Sources/App/Core/SearchFilter.swift
@@ -37,7 +37,7 @@ struct SearchFilterParser {
     static var allSearchFilters: [SearchFilter.Type] = [
         StarsSearchFilter.self,
         LicenseSearchFilter.self,
-        LastCommitSearchFilter.self,
+        LastActiveSearchFilter.self,
     ]
     
     /// Separates search terms from filter syntax.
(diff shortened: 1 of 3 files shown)


Sherlouk:
Is there any reason to not support both?

Correct me if I'm wrong but last activity could include somebody other than the author creating a new issue or something - it doesn't necessary mean the project is actively being maintained? At least commit means code is actually changing - though a very crude measure of quality.

Sherlouk:
(This ticket also reminds me that filters still only work in staging 😭  - I should come back and finish up the UI for it)

daveverwer:
> Is there any reason to not support both?

Yea. As I mentioned there's an argument to have both but this feature is in a funny place as it never quite made it out of the door. If I had to pick between the two, I'd say last commit/issue/pr activity is better than *purely* last commit and I'm reluctant expand the scope of the feature (which then needs maintaining) before it's finished.

> Correct me if I'm wrong but last activity could include somebody other than the author creating a new issue or something - it doesn't necessary mean the project is actively being maintained? At least commit means code is actually changing - though a very crude measure of quality.

Not quite. Creating an issue wouldn't count as activity. Closing or merging a pull request, closing an issue, or committing code is what counts as activity. I'd class all of those as active maintenance.

daveverwer:
> (This ticket also reminds me that filters still only work in staging 😭 - I should come back and finish up the UI for it)

I would very much support that! It's so close to being a fantastic feature.

Sherlouk:
> As I mentioned

Yeah turns out I didn't read your message in the slightest, sorry 😂 

> Creating an issue wouldn't count as activity. Closing or merging a pull request, closing an issue, or committing code is what counts as activity

What if I create a ticket on your repo, and then close it myself?

daveverwer:
> What if I create a ticket on your repo, and then close it myself?

That would need testing. I have a feeling that it would count as activity, though.

Sherlouk:
> I have a feeling that it would count as activity

Good to know at least. A point for a future issue but when we add UI for this it'd be good to add a definition of "activity" somewhere so users know.

Happy with the code 👍

daveverwer:
Note: Move `viewDateFormatter` from #1401 out into a global helper as part of this.

daveverwer:
Superseded by #1406

PR #1394, opened 2021-11-24
Have the number of dependencies contribute towards package score

daveverwer:
Merge after #1392
Fixes #1294

PR #1395, opened 2021-11-25
Avoid magic value in nil case

finestructure:
This is how I'd avoid that `1_000` magic value mentioned here: https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/pull/1394#discussion_r756226123


daveverwer (approved):
I spent so long yesterday avoiding optionals I forgot what they were for 😂

PR #1401, opened 2021-11-26
Search Filter Improvements

### LINKED ISSUE #318: Narrow searches based on metadata
The *how* of this still needs to be discussed, but it's common to want to include/exclude packages based on a few different factors. For example:

* Is a package using an App Store compatible license?
* Does a package include libraries? (as opposed to only executables)
* Does a package have more than 20 dependencies?
* Has a package been released in the last 12 months?

### LINKED ISSUE #1228: Display feedback about any filters have been applied during a search
To enable better debugging of filtered search, and a better explanation of what happened for people searching we need to describe what the results are in a better way.

Currently, above the search results, it shows this:

Screenshot 2021-07-12 at 09 11 08@2x

We should break down what's being searched for and display a readable explanation of it. Something along the lines of this:

> Results for “test” where [Last commit date] [is >] [2021-01-01], [Number of stars] [is >] [500], and [Platform] [is compatible with] [iOS and macOS]

We should use styling to separate the different bits of that sentence surrounded by square brackets, so make sure each clause is wrapped in a span. I can help with the styling when it comes to that.

We could either display this in a sentence format or as a set of rows below the results line. Something along the lines of this:

Swift Package Index - 2021-07-12 09 26 06

If there is no actual search term, it should say "Results for all packages" where ...

Sherlouk:
## Description

This pull request aims to add some finishing touches to the search filter capability which has been enabled in staging for quite a while now (sorry). 

Specifically, this makes two changes:

1. Search filters are visually separated from the search term as part of the search results page
2. Search filters are applied only to packages, not to keywords/authors allowing for these results to show side-by-side

There will be further improvements to search filters including the ability to filter by keyword and more, but this can could come after it's production-ised and certainly as a separate PR.

## Association

Closes #1228
Closes #318
Related to #1396 
Conflicts with #1390


Sherlouk:
This is what this branch currently looks like. Dave has already said he hates everything I do visually and so will come through and clean up the UI anyway ❤️ 

Captured 2021-11-26 at 1 02 22

daveverwer:
> Dave has already said he hates everything I do visually and so will come through and clean up the UI anyway ❤️

I don't remember saying that! 😅 But I think this is 95% there already. I really like it. 👍

Sherlouk:
Just in case somebody reads this and thinks that Dave is like supremely evil, it was a joke 😅  Dave and I already agreed he was going to go through and do the UI on Discord ahead of me picking this up! (CSS is not my forte)

daveverwer:
This is looking really, really good. A few things I found while giving it a test tonight:

### No filter descriptions when no package results are displayed

Screenshot 2021-11-26 at 19 47 37@2x

So I had coded the CSS for the keyword results never expecting a situation where there would be a keyword result without any package results, which breaks the layout a bit! However, the bigger problem here is that without package results we're not seeing the filter description.

I wonder if we should make it so this page always displays package results, or maybe just always display them if filters have been applied. Looking at the design for a search that matches nothing, it's OK, but it'd also be fine if that text were below the "Matching packages" heading.

Screenshot 2021-11-26 at 19 52 48@2x

More to follow...

daveverwer:
### Bug with `last_commit` filters?

On staging, searching for `last_commit:>2020-01-01` returns plenty of results, but on this branch it returns nothing.

Screenshot 2021-11-26 at 20 48 27@2x

If this is unrelated to this code, we can log this as a separate issue.

daveverwer:
### Should we give the fields more descriptive names in the filter description?

Rather than:

Screenshot 2021-11-26 at 21 01 10@2x

What do you think about:

Screenshot 2021-11-26 at 21 02 16@2x

I'm not sure about this, but last_commit is not ideal.

It might just be that we need:

Screenshot 2021-11-26 at 21 04 06@2x

daveverwer:
Pushed up a few small CSS tweaks but this is in great shape on the front end. 🎉 Thanks James!

Sherlouk:
Appreciate the support Dave!

There's definitely a known (expected) bug where it will return zero results if you only include a filter. 

That's just down to how the search mechanism works at the moment because it needs _something_ to search and then just uses the filter to... well... filter down those options.

Probably need to add something in there which lets it search all packages if the term is empty and filters are present. I'd maybe think tackle that separately though?

As for the names as shown in the view, that's easily changeable. The createViewModel function just uses `Self.key` at the moment but there's zero reason that couldn't be a more human-readable string. Do you want to pick that up under this?

I was also thinking whether the date one wants to use a formatted value in the view model? So instead of "Last Commit is greater than 2020-01-01" it would be more like "Last Commit is greater than 1st Jan 2020". Again, something that could be changed purely in the createViewModel function (and why I made it a separate model).

daveverwer:
> There's definitely a known (expected) bug where it will return zero results if you only include a filter.
> 
> That's just down to how the search mechanism works at the moment because it needs _something_ to search and then just uses the filter to... well... filter down those options.
> 
> Probably need to add something in there which lets it search all packages if the term is empty and filters are present. I'd maybe think tackle that separately though?

One reason I put this in here rather than as a separate PR is that these search terms work on staging:

* `https://staging.swiftpackageindex.com/search?query=stars%3A%3E30000`
* `http://localhost:8080/search?query=stars%3A%3E30000`

daveverwer:
> As for the names as shown in the view, that's easily changeable. The createViewModel function just uses `Self.key` at the moment but there's zero reason that couldn't be a more human-readable string. Do you want to pick that up under this?

Yea, let's do that. I think I prefer option 3 out of all of the screenshots. How about you?

Sherlouk:
I wonder if the fact we do a union now for keyword/author search is killing the entire search.

That's the only change I can imagine causing that.

I'll try reverting that in the morning and test.

daveverwer:
> I wonder if the fact we do a union now for keyword/author search is killing the entire search.
> 
> That's the only change I can imagine causing that.
> 
> I'll try reverting that in the morning and test.

Let's write a test for it too to stop future regressions.

daveverwer:
> I was also thinking whether the date one wants to use a formatted value in the view model? So instead of "Last Commit is greater than 2020-01-01" it would be more like "Last Commit is greater than 1st Jan 2020". Again, something that could be changed purely in the createViewModel function (and why I made it a separate model).

Great idea 👍

Sherlouk:
Captured 2021-11-27 at 11 09 45

I've done a number of updates and this is where I'm at with it.

Sherlouk:
Think I've now responded to all of the feedback/bugs. Should be good for a review/test

Sherlouk:
I know I shouldn't be proud of this... but every commit passed on CI and that's a good feeling

Sherlouk:
As per chat with Dave,`appStoreCompatible` has now been renamed to `compatible`.

Sherlouk:
Current

```
last_commit 
stars
license
```

Coming Soon

```
last_activity
```

Future Ideas

```
last_release
dependencies
dependants
compatible (:>swift-5 :linux)
libraries
executables
keyword
author
```

Sources/App/Core/SearchFilter.swift
```diff
@@ -144,6 +162,12 @@ enum SearchFilterError: Error {
 struct StarsSearchFilter: SearchFilter {
     static var key: String = "stars"
     
+    private static var numberFormatter: NumberFormatter = {
+        let formatter = NumberFormatter()
+        formatter.numberStyle = .decimal
+        return formatter
+    }()
+
```
daveverwer, line 170: Just doing a final review and noticed this. We have a `starsFormatter` available from everywhere in `NumberFormatter+stars.swift`.

Sources/App/Core/SearchFilter.swift
```diff
@@ -250,16 +292,27 @@ struct LastCommitSearchFilter: SearchFilter {
         return formatter
     }()
     
+    static var viewDateFormatter: DateFormatter = {
+        let formatter = DateFormatter()
+        formatter.dateFormat = "d MMM yyyy"
+        formatter.locale = Locale(identifier: "en_US_POSIX")
+        formatter.timeZone = TimeZone(secondsFromGMT: 0)
+
```
daveverwer, line 300: We don't have one for dates but I wonder if we should pull this out so our formatters are a little more organised? Equally, I could do this as part of expanding the filters to include `last_active`. Noted for inclusion in #1390.

Sources/App/Views/Search/SearchShow+View.swift
```diff
@@ -28,7 +28,7 @@ extension SearchShow {
 
         override func pageTitle() -> String? {
             if model.query.count > 0 {
-                return "Search Results for &ldquo;\(model.query)&rdquo;"
+                return "Search Results for &ldquo;\(model.term)&rdquo;"
```
daveverwer, line 31: Well spotted! 😍

daveverwer (approved):
Tested locally and everything looks perfect!

I did just notice one tiny code clean up in my final review with the number formatter.

Sources/App/Views/Search/SearchShow+View.swift
Sherlouk: Think I caught all uses. Worth noting that this does a little funky in the scenario where there are no terms (only filters) but I don't think it's a major problem.

Sources/App/Core/SearchFilter.swift
Sherlouk: Probably makes sense to have a generic "search filter date formatter" for all filters which have a date input. My typical coding brain would say refactor that once there's a need, so in 1390, though. I'm happy to sort out that rebase+cleanup if need be once this gets merged through.

Sources/App/Core/SearchFilter.swift
Sherlouk: Thanks for letting me know, I didn't realise this existed (but suppose makes sense). Have pushed up a commit which uses this now.

Sources/App/Views/Search/SearchShow+View.swift
daveverwer: Fixed in 007bdeb1c053248f4869feebfe47d299b33c61cc

Sherlouk: Aha fair enough, cheers Dave

daveverwer:
This is an amazing feature, @Sherlouk. Thank you so much for all of your efforts with it.

❤️

Sherlouk:
teamwork makes the dream work

PR #1406, opened 2021-11-30
Add `last_active` filter to the search filters.

### LINKED ISSUE #1303: Add more metadata to package lists

daveverwer:
Supercedes #1390
Fixes #1303

### DIFF
--- a/Sources/App/Core/SearchFilter.swift
+++ b/Sources/App/Core/SearchFilter.swift
@@ -273,24 +273,6 @@ struct LicenseSearchFilter: SearchFilter {
 /// last_commit:<2020-07-01 - Last commit made on any day older than July 1st 2020
 /// ```
 struct LastCommitSearchFilter: SearchFilter {
-    static var parseDateFormatter: DateFormatter = {
-        let formatter = DateFormatter()
-        formatter.dateFormat = "yyyy-MM-dd"
-        formatter.locale = Locale(identifier: "en_US_POSIX")
-        formatter.timeZone = TimeZone(secondsFromGMT: 0)
-        
-        return formatter
-    }()
-    
-    static var viewDateFormatter: DateFormatter = {
-        let formatter = DateFormatter()
-        formatter.dateFormat = "d MMM yyyy"
-        formatter.locale = Locale(identifier: "en_US_POSIX")
-        formatter.timeZone = TimeZone(secondsFromGMT: 0)
-        
-        return formatter
-    }()
-    
     static var key: String = "last_commit"
     
     let comparison: SearchFilterComparison
@@ -298,7 +280,7 @@ struct LastCommitSearchFilter: SearchFilter {
     let value: String
     
     init(value: String, comparison: SearchFilterComparison) throws {
-        guard let date = Self.parseDateFormatter.date(from: value) else {
+        guard let date = DateFormatter.filterParseFormatter.date(from: value) else {
             throw SearchFilterError.invalidValueType
         }
         
@@ -319,7 +301,7 @@ struct LastCommitSearchFilter: SearchFilter {
         .init(
             key: "last commit",
             comparison: comparison,
-            value: Self.viewDateFormatter.string(from: date)
+            value: DateFormatter.filterDisplayFormatter.string(from: date)
         )
     }
 }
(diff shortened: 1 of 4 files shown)


daveverwer:
This is the most primitive implementation of a `last_activity` filter. Some things that struck me while doing this:\

* This isn't a huge amount of code, so it's not a huge deal but there's a lot of duplication here.
* We can certainly pull the date formatters out into shared code, but is there more we can do? Especially if we're expanding this system to more fields going forward.

Sherlouk (approved):

Sources/App/Core/SearchFilter.swift
```diff
@@ -338,32 +320,14 @@ struct LastCommitSearchFilter: SearchFilter {
...
-        formatter.dateFormat = "yyyy-MM-dd"
-        formatter.locale = Locale(identifier: "en_US_POSIX")
-        formatter.timeZone = TimeZone(secondsFromGMT: 0)
-
-        return formatter
-    }()
-
-    static var viewDateFormatter: DateFormatter = {
-        let formatter = DateFormatter()
-        formatter.dateFormat = "d MMM yyyy"
```
finestructure: Personally I find it a bit odd that the display format is different from the supported search syntax. I guess it does tell the user that the parsing was ok but it also suggests that they should use the display format (which wouldn't work, or would it?).

I think I'd have stuck with `yyyy-MM-dd` in both cases. It's the unambiguous "programmery" date format we're all comfortable with.

finestructure (approved):

Sources/App/Core/SearchFilter.swift
daveverwer: I'm not sure I agree here. The reason we have the YYYYMMDD format in the filter is for ease of parsing, but I don't think it's necessarily a good thing that we use that elsewhere. In fact, I think given that the formatted output here is using a month name is a great validation of what date is being parsed out.

I'd vote to keep it like this.

Sherlouk: I can definitely see your perspective. I changed the display format not really as a way of proving it worked but more of a validation to the user that they got the correct format. I'll be the first to admit I often get the MM and DD mixed up and it could return confusing results otherwise.

One potential option is to have some form of tooltip perhaps? So the display format is the raw input (YYYY-MM-DD) but if you hover over it then you see the formatter version as a validation?

finestructure: It's just that it leads to

CleanShot 2021-12-01 at 14 52 22@2x

or, slightly weirder

CleanShot 2021-12-01 at 14 53 06@2x

(where I input `test last_commit:>"1 Nov 2021"` and it gets truncated to `test last_commit:>`)

🤷‍♂️

daveverwer: > (where I input test last_commit:>"1 Nov 2021" and it gets truncated to test last_commit:>)

That's a great bug, but unrelated to this change I think.

I'd still vote to keep the formatting like it is here, but both of these issues are a little unrelated to this PR as the date filters already work like this.

I'm going to merge this PR. If you feel strongly that we should change it then let's log it as a bug and discuss it further in a separate issue.

daveverwer:
Deploying as 2.50.2.

Sources/App/Core/SearchFilter.swift
finestructure: > That's a great bug, but unrelated to this change I think.

Yes, it's unrelated. James and I discussed that briefly in the original PR where we noticed this behaviour.

I'm not too fussed, just wanted to point out the issue and had ticked the approve already :)

PR #1411, opened 2021-12-01
Minor tweaks to search filters (docs, metrics, cleanup)

Sherlouk:
We keep on grinding... I will make this a great feature 😅 

This PR:
* Breaks the huge file down into more manageable files
* Adds documentation to the FAQs section
* Adds support for >= and <= operators
* Adds some initial metrics as to understand how users use filters (if at all)


Sources/App/Core/SearchFilter/SearchFilterParser.swift
```diff
@@ -0,0 +1,85 @@
...
+            .filter { !$0.isEmpty }
+        
+        guard components.count == 2 else {
+            return nil
+        }
+        
+        // Operator
+        let comparison: (length: Int, value: SearchFilterComparison) = {
+            let value = components[1]
+            switch value {
```
finestructure, line 59: I've just stumbled across this while looking into the metrics. This looks odd: `value` is a closure parameter and then gets immediately shadowed by a local variable?

finestructure: Ah, never mind - it's not a closure, it's a tuple.

Package.swift
```diff
@@ -55,7 +55,7 @@ let package = Package(
             "OhhAuth",
             "SwiftSoup",
             .product(name: "SwiftPMPackageCollections", package: "SwiftPM")
-        ]),
+        ], exclude: [ "Core/SearchFilter/README.md" ]),
```
Sherlouk, line 58: Feedback from Dave:

Change this README some more, and make it user-facing in the FAQs

daveverwer (approved):
I'd argue that this is already a great feature. But you are correct that this makes it better! Thanks, James!

PR #1412, opened 2021-12-01
Search Filter Metrics

Sherlouk:
Based on https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/discussions/1403#discussioncomment-1728037

Adds four new metrics. The intention is to track the usage of each filter (and filters in general) as well as how many filters/terms are used in each search query. We already track the number of searches in order to build a percentage.

Not sure if the gauge is correctly implemented but the counters seem fine. Looking for some reviews to guide on this.

### DIFF
--- a/Sources/App/Core/AppMetrics.swift
+++ b/Sources/App/Core/AppMetrics.swift
@@ -105,6 +105,22 @@ enum AppMetrics {
     static var apiSearchGetTotal: PromCounter<Int, EmptyLabels>? {
         counter("spi_api_search_get_total", EmptyLabels.self)
     }
+    
+    static var apiSearchGetWithFilterTotal: PromCounter<Int, EmptyLabels>? {
+        counter("spi_api_search_get_with_filter_total", EmptyLabels.self)
+    }
+    
+    static func apiSearchGetWithFilterTotal(key: String) -> PromCounter<Int, EmptyLabels>? {
+        counter("spi_api_search_get_with_filter_\(key)_total", EmptyLabels.self)
+    }
+    
+    static var searchTermsCount: PromGauge<Int, EmptyLabels>? {
+        gauge("spi_search_terms_count", EmptyLabels.self)
+    }
+    
+    static var searchFiltersCount: PromGauge<Int, EmptyLabels>? {
+        gauge("spi_search_filters_count", EmptyLabels.self)
+    }
 
     static var buildCandidatesCount: PromGauge<Int, EmptyLabels>? {
         gauge("spi_build_candidates_count", EmptyLabels.self)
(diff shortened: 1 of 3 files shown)


Sources/App/Core/AppMetrics.swift
```diff
@@ -105,6 +105,22 @@ enum AppMetrics {
     static var apiSearchGetTotal: PromCounter<Int, EmptyLabels>? {
         counter("spi_api_search_get_total", EmptyLabels.self)
     }
+    
+    static var apiSearchGetWithFilterTotal: PromCounter<Int, EmptyLabels>? {
+        counter("spi_api_search_get_with_filter_total", EmptyLabels.self)
+    }
+    
+    static func apiSearchGetWithFilterTotal(key: String) -> PromCounter<Int, EmptyLabels>? {
+        counter("spi_api_search_get_with_filter_\(key)_total", EmptyLabels.self)
```
finestructure, line 114: `key` should be a label, and we should also probably constrain it to an `enum`. This will also make `spi_api_search_get_with_filter_total` superfluous: you can filter by label individually and against the total without labels.

I'll take a look and will open a PR on top of this one.

finestructure:
That's great, thanks James!

Sources/App/Core/AppMetrics.swift
Sherlouk: Making it a label in hindsight would've been a smart idea, agreed.

I do get where you're coming from with respect of using an enum for the key, i just wanted to try and make it so that you didn't have to think about metrics every time you add a new filter. With an enum, it would force us to update it each time a new metric is added

finestructure: I've just opened a PR with some minimal changes. I think there's more we could do here to make it type-safe and improve testability/maintenance a bit but that's all minor and not related to the labels themselves.

Sources/App/Core/Search.swift
```diff
@@ -277,6 +277,14 @@ enum Search {
         let page = page.clamped(to: 1...)
         let (sanitizedTerms, filters) = SearchFilterParser().split(terms: sanitize(terms))
         
+        // Metrics
+        if filters.isEmpty == false {
+            AppMetrics.apiSearchGetWithFilterTotal?.inc()
+        }
```
finestructure, line 283: I forgot to remove this - it is now superfluous.

Sherlouk:
All seems to make sense. Were the gauges used correctly here to try and establish an average rather than a sum?

finestructure:
They should be, yes!

Sherlouk:
Not had a formal approval, but given the small changes and the sense of happiness I'm going to merge this into the little feature branch I have going on.

PR #1413, opened 2021-12-01
Add `author` and `keyword` search filters

Sherlouk:
This pull request introduces two new search filters. `author` and `keyword`. Both of these filters support equals to, and not equals to operators. Values are case insensitive, and keywords supports partial matching.

Based on discussions in #1403

http://localhost:8080/search?query=author%3ASherlouk


Sherlouk:
@finestructure Am I missing something simpler with respect of doing a case insensitive comparison in this query? Or does my idea of adding a new field which is just a function on the existing values make sense?

finestructure:
I'd use either `ilike`, like we do for the package page query:

```swift
            .filter(Repository.self, \.$owner, .custom("ilike"), owner)
            .filter(Repository.self, \.$name, .custom("ilike"), repository)
```

or if that isn't possible for some reason do a `lower(...)` on both sides of the comparison.

Sherlouk:
Let me take a look into how the `where` query works and see if I can use the custom `ilike` comparison operator. Thanks for that pointer

Sherlouk:
Thanks for that tip Sven, I've got it working as expected now!

http://localhost:8080/search?query=fluent+author%3AvaPor (fluent author:vaPor)
http://localhost:8080/search?query=fluent+author%3A%21vaPor (fluent author:!vaPor)

Sources/App/Core/Search.swift
```diff
@@ -165,6 +165,7 @@ enum Search {
             .column(lastCommitDate)
             .column(lastActivityAt)
             .from(searchView)
+            .from(SQLFunction("CONCAT", args: keywords), as: keyword)
```
Sherlouk, line 168: The output of this, for future reference is `{keyword,keyword2,keyword3}`.

daveverwer (approved):
This is great! Thank you James!

PR #1416, opened 2021-12-03
Switch search filter to use binds instead of string literals

Sherlouk:
Who doesn't love a bit of SQL injection in the morning? (It's the afternoon James)

Technically, we have protection elsewhere which meant this couldn't ever happen... but I've switched over to using a bind instead of a string literal for a little bit of added protection.

Checked and does all still work, even with wildcards (%) in filters, but you can't execute your own commands now. Suppose that's a win?


finestructure (approved):

finestructure:
Thanks, James!

daveverwer:
Deploying as 2.50.3.

PR #1417, opened 2021-12-03
Add support for none/other license searches

Sherlouk:
Non-breaking tweak to filters here. Instead of handling `appStoreCompatible` directly, we use the LicenseKind enum which already exists.

By making this change, it also means we get access to the three other values: `incompatible` (this is technically a duplicate of the already functional `license:!compatible`), `none` (no license), and `other` (unknown license).

http://localhost:8080/search?query=license%3Aother
http://localhost:8080/search?query=license%3Anone

Things we couldn't do before:
```
license:!none // any license
license:none // no license
license:other // unknown license
license:!other license!none // any known license
```

edit:// correction. `license:!compatible` and `license:incompatible` are actually slightly different in a rather subtle and maybe mildly confusing way.

!compatible means that it is **not** one of the handful of known licenses that we have marked as compatible.
incompatible means that it is explicitly one of the handful of known licenses that we know is not compatible.

This means that for packages with a license of `none` or `other`, they are !compatible but not incompatible. Say that 3 times quickly.

I don't think this is a distinction we need to worry about?

### DIFF
--- a/Sources/App/Models/License.swift
+++ b/Sources/App/Models/License.swift
@@ -154,10 +154,10 @@ enum License: String, Codable, Equatable, CaseIterable {
         
         var userFacingString: String {
             switch self {
-            case .none: return License.none.shortName
-            case .other: return License.other.shortName
-            case .incompatibleWithAppStore: return "App Store incompatible"
-            case .compatibleWithAppStore: return "App Store compatible"
+            case .none: return "not defined"
+            case .other: return "unknown"
+            case .incompatibleWithAppStore: return "incompatible with the App Store"
+            case .compatibleWithAppStore: return "compatible with the App Store"
             }
         }
     }
--- a/Sources/App/Models/License.swift
+++ b/Sources/App/Models/License.swift
@@ -151,6 +151,15 @@ enum License: String, Codable, Equatable, CaseIterable {
         case other
         case incompatibleWithAppStore = "incompatible"
         case compatibleWithAppStore = "compatible"
+        
+        var userFacingString: String {
+            switch self {
+            case .none: return License.none.shortName
+            case .other: return License.other.shortName
+            case .incompatibleWithAppStore: return "App Store incompatible"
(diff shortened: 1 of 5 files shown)


Sources/App/Models/License.swift
```diff
@@ -151,6 +151,15 @@ enum License: String, Codable, Equatable, CaseIterable {
         case other
         case incompatibleWithAppStore = "incompatible"
         case compatibleWithAppStore = "compatible"
+        
+        var userFacingString: String {
+            switch self {
+            case .none: return License.none.shortName
+            case .other: return License.other.shortName
+            case .incompatibleWithAppStore: return "App Store incompatible"
```
daveverwer, line 159: This wording seems off. My first instinct was to have this say "incompatible with the App Store" but when it's in the filter it says "matches" and so:

"matches incompatible with the App Store" is weird

but so is:

"matches App Store incompatible"

It makes me think that "matches" might be better as "is". I think "is" would work everywhere "matches" works and would make this make more sense too.

Sources/App/Models/License.swift
```diff
@@ -151,6 +151,15 @@ enum License: String, Codable, Equatable, CaseIterable {
...
+            case .compatibleWithAppStore: return "App Store compatible"
```
daveverwer, line 160: same with "compatible with the App Store"

Sherlouk: Hadn't actually changed this one 😀 

I don't disagree with the feedback, just a limitation with how I've currently got it setup. Won't be too hard to tweak though so will update this PR in the morning with amendments.

daveverwer: Yea I wasn't meaning to make "is" specific for this case. I think it would work as a blanket replacement for "matches"

daveverwer (approved):
This is :chefs_kiss:. Thanks James!

PR #1418, opened 2021-12-04
Add `platform` search filter

Sherlouk:
⚠️  Schema change.
⚠️  Work in progress. Need support cleaning up the query to make it more legible, but it does appear to work as is.

This pull request introduces a `platform` search filter. For example `platform:ios` or `platform:!macos`.


Sources/App/Core/SearchFilter/Filters/PlatformSearchFilter.swift
```diff
@@ -0,0 +1,93 @@
...
+        self.comparison = comparison
+        self.value = platform
+    }
+    
+    func `where`(_ builder: SQLPredicateGroupBuilder) -> SQLPredicateGroupBuilder {
+        let exp = SQLBinaryExpression(
+            left: SQLIdentifier("platforms").cast(to: "jsonb"),
+            op: SQLRaw("@>"),
+            right: SQLBind("[{\"name\":\"\(value.rawValue)\"}]").cast(to: "jsonb")
+        )
```
Sherlouk, line 49: Okay so the query I got working was this:

```
SELECT sp, package_name
FROM "versions", array_to_json(supported_platforms) AS sp 
WHERE sp::jsonb @> '[{"name":"ios"}]'::jsonb
LIMIT 100;
```

This shows all versions which supports iOS (or so my testing makes me believe).

Unfortunately SQL Kit does not make it easy to deal with:
1. JSON contains operators
2. JSONB casting (which is required to make this work)
3. JSON array syntax
4. Inverted (NOT) group expressions

I've tried to write my own helpers to make this a bit more readable but... 🤷

Sources/App/Core/Search.swift
```diff
@@ -33,12 +33,16 @@ enum Search {
     static let lastActivityAt = SQLIdentifier("last_activity_at")
     static let license = SQLIdentifier("license")
     static let lastCommitDate = SQLIdentifier("last_commit_date")
+    static let platforms = SQLIdentifier("platforms")
     static let searchView = SQLIdentifier("search")
     static let summary = SQLIdentifier("summary")
+    static let supportedPlatforms = SQLIdentifier("supported_platforms")
+    static let swiftVersions = SQLIdentifier("swift_versions")
```
Sherlouk, line 40: I was very optimistic this evening 😂

Sources/App/Core/SearchFilter/Filters/PlatformSearchFilter.swift
finestructure: Yes, fluent doesn't support JSON specific operators etc yet.

I haven't tried or looked this up but something like this might/should work

```swift
        let exp: SQLBinaryExpression = SQLRaw(
            #"platforms::JSONB @> '[{"name":\#(bind: value.rawValue)'::JSONB"#
        )
```

As the query itself, it might be easier to unpack the names from the supported platforms array so that essentially you're testing against a `text[]` of platform names. That should also open up using Fluent again.

Just a quick comment, I haven't looked into this very thoroughly.

Sherlouk: > it might be easier to unpack the names from the supported platforms array

I assume what you mean here is that `platforms` should be a text array which is `["ios","macos"]` instead of `[{"name": "tvos", "version": "10.0"},{"name": "ios", "version": "10.0"}]`? This was actually my first idea, I just couldn't figure out what the SQL query looked like to make that work - is there a specific function(s) I should be looking at?

Sherlouk: Just to add, I did have a play around with your example code above to tidy up the query.  

It looks like you have to use `SQLQueryString` to get the binding interpolation, but this doesn't actually work since the SQLBind feature wraps the binding in apostrophes (`'ios'`) which isn't valid JSON. It needs to be `"ios"` for the query to work 😞

finestructure: I just realised `supported_platforms` is the wrong field anyway:

```swift
    // TODO: rename to minimumPlatformVersions?
    @Field(key: "supported_platforms")
    var supportedPlatforms: [Platform]
```

which is filled from

```swift
    version.supportedPlatforms = manifest.platforms?.compactMap(Platform.init(from:)) ?? []
```

i.e. the manifests minimum required platform version field. Our version compatibility is from the build records (1-N on a version). That's not going to be searchable without some intermediate aggregation, either into a materialised view or into the version.

I don't think this search (not Swift version support) is doable without that (nontrivial) groundwork.

Sherlouk: That explains a lot about why so many packages didn't have data I suppose, since it's coming from a different source. Had my suspicion but didn't see the correlation with the Build model.

Will close this PR then, completely off the mark and need to come back at it differently.

PR #1419, opened 2021-12-04
Add `dependencies` search filter 

Sherlouk:
After the absolute fail which was #1418, I thought I'd try and restore faith in myself.

⚠️  Schema change. Migration needed.

This PR adds a new search filter (there's a bit of a trend this week, don't know if anyone has noticed) which allows you to query the number of dependencies the project has - for example: `dependencies:<=2` will show packages with 2 or fewer dependencies.

If/when this PR goes through, I'm probably going to look into how we can reuse search filters rather than creating them from scratch each time. We have two date filters, and now two numeric filters, there is very little difference between the two of them. This can certainly be cleaned up.


Tests/AppTests/SearchTests.swift
```diff
@@ -21,6 +21,8 @@ import XCTVapor
 
 class SearchTests: AppTestCase {
     
+    let recordSQL = false
```
Sherlouk, line 24: This and the move to using inline snapshots is purely quality of life. I'm updating the query a lot at the moment as we bring more data in for search improvements - updating these tests is a bit of a chore. This lovely lil record boolean makes it a tonne easier.

Sherlouk:
Fun fact: largest project I've seen has 36 dependencies

Sources/App/Core/Search.swift
```diff
@@ -164,8 +166,20 @@ enum Search {
...
+                SQLFunction(
+                    "COALESCE",
+                    args: SQLFunction(
+                        "array_length",
+                        args: resolvedDependencies, SQLLiteral.numeric("1")
+                    ),
+                    SQLLiteral.numeric("0") // default to zero if NULL or empty array
+                ),
+                as: dependenciesCount
+            )
```
finestructure, line 182: We have a `coalesce` function in `SQLKit+functions.swift`:

```swift
func coalesce(_ args: SQLExpression...) -> SQLFunction {
    SQLFunction("COALESCE", args: args)
}
```

which should work here. (There's also a `concat` in that file, which could perhaps replace that usage on line 171, which I missed in an earlier PR.)

Tests/AppTests/SearchTests.swift
finestructure: I think

```swift
        SnapshotTesting.isRecording = true
```

should work for `_assertInlineSnapshot` as well. It's what we use for a temporary override with all the other snapshots without needing to add `record: true` to the individual calls (see `SnapshotTestCase.swift` - but it can be added anywhere).

Tests/AppTests/SearchTests.swift
```diff
@@ -36,46 +38,58 @@ class SearchTests: AppTestCase {
...
+        _assertInlineSnapshot(matching: renderSQL(b), as: .lines, record: recordSQL, with: """
+        SELECT 'package' AS "match_type", "keyword", "package_id", "package_name", "repo_name", "repo_owner", "score", "summary", "stars", "license", "last_commit_date", "last_activity_at", "dependencies_count" FROM "search", CONCAT("keywords") AS "keyword", COALESCE(array_length("resolved_dependencies", 1), 0) AS "dependencies_count" WHERE CONCAT_WS(' ', "package_name", COALESCE("summary", ''), "repo_name", "repo_owner") ~* $1 AND "package_name" IS NOT NULL AND "repo_owner" IS NOT NULL AND "repo_name" IS NOT NULL ORDER BY LOWER("package_name") = $2 DESC, "score" DESC, "package_name" ASC
+        """)
```
finestructure, line 43: Using

finestructure: Actually, scratch that. `renderSQL` is a method on `AppTestCase` because it's using `app.db` to instantiate an `SQLSerializer`. We'd need to make it `.sql(app.db)` to pass it into the strategy and also make `renderSQL` and `bind` static or global functions. At that point it's just simpler to render it out into a `String` as is.

Sources/App/Core/Search.swift
Sherlouk: Will clean this up, thanks for pointing out the helper!

Sources/App/Core/SearchFilter/SearchFilter.swift
```diff
@@ -70,7 +70,7 @@ enum SearchFilterComparison: String, Codable, Equatable {
         case .match: return "is"
         case .negativeMatch: return "is not"
         case .greaterThan: return "is greater than"
-        case .greaterThanOrEqual: return "is greather than or equal to"
+        case .greaterThanOrEqual: return "is greater than or equal to"
```
Sherlouk, line 73: 🙈

Sources/App/Migrations/039/UpdateSearch4.swift
```diff
@@ -0,0 +1,76 @@
...
+              r.keywords,
+              r.last_commit_date,
+              r.license,
+              r.name AS repo_name,
+              r.owner AS repo_owner,
+              r.stars,
+              r.last_activity_at,
+              r.summary,
+              v.package_name,
+              v.resolved_dependencies
```
finestructure, line 42: I think we need to have a think about what the plans are for searching through a packages dependencies. Or rather: given that we're only looking at the count at this time, I think we want to store a dependency count only at this time instead of copying the whole column, which can be quite big.

finestructure: On second thought, I think we should hold off on this filter until we've got some usage info regarding the filters in general. The others have all been "client side" filter logic but before we start making further schema changes we should evaluate where we are.

Would you mind parking it in the meantime, James?

PR #1420, opened 2021-12-06
Package update

finestructure:


Sources/App/Core/Extensions/SQLSelectBuilder+ext.swift
```diff
@@ -16,12 +16,6 @@ import SQLKit
 
 
 extension SQLSelectBuilder {
-    // sas 2020-06-05: workaround `direction: SQLExpression` signature in SQLKit
-    // (should be SQLDirection)
-    func orderBy(_ expression: SQLExpression, _ direction: SQLDirection = .ascending) -> Self {
-        return self.orderBy(SQLOrderBy(expression: expression, direction: direction))
-    }
-
```
finestructure, line 24: I often think it's maybe overkill to update packages every week but this actually saved us quite some time debugging. Because it was just three packages it was easy to see in the change notes that SQLKit had added some new overloads. It was source compatible but made our workaround call itself and crashed the tests with a stack overflow.

daveverwer: Wouldn't it be great if something could look at your Package.resolved diff and open up release notes on SPI for all packages 😱

daveverwer (approved):

Sources/App/Core/Extensions/SQLSelectBuilder+ext.swift
finestructure: That would be an amazing feature!

PR #1423, opened 2021-12-06
Tweaked the db-dump script for the new flexible PG server

daveverwer:


scripts/dump-db.sh
```diff
@@ -25,5 +24,5 @@ if [[ $REPLY =~ ^[Yy]$ ]]
 then
   # This command assumes the db will be available on $PORT on localhost, via an ssh tunnel:
   # For example: ssh -i <your private key> -L 7432:db:5432 -p 2222 root@173.255.229.82
-  pg_dump --no-owner -Fc -h localhost -p ${PORT} -U ${USER}@${HOST} ${DATABASE} > ${USER}-$(date +%Y-%m-%d).dump
+  pg_dump --no-owner -Fc -h localhost -p ${PORT} -U ${USER} ${DATABASE} > ${USER}-$(date +%Y-%m-%d).dump
```
finestructure, line 27: We shouldn't be dumping via the tunnel though - so `localhost` certainly isn't correct here. In fact we should remove this script to make sure we use the one in infrastructure!

daveverwer:
Replaced by #1427

PR #1424, opened 2021-12-07
Fix search filter operator typo

Sherlouk:
Closed #1419 off the back of Sven's review, but it contained a minor typo fix - so opening that up separately so we can get it through.


finestructure (approved):

finestructure:
Thanks James!

PR #1426, opened 2021-12-07
Add filter suggestions under the search box on the search results page

daveverwer:
Merge after #1428 

These are deliberately *only* on the search results page as the most common use of these filters will be to narrow down search results. The filters work from everywhere, of course, but for education, I'd say this is the best place.

Screenshot 2021-12-07 at 19 23 14

This obviously looks awful on mobile. I could tweak it slightly but it's never going to be great and I wonder if we should make this a desktop-only feature. There is one thing I considered that might help here, and which I'd love an opinion on. I added the suggestions with JavaScript as they rely on JavaScript to work, but the "Add filters to narrow search results (Learn more)." text doesn't. That could be added server-side, and also display on mobile as it'd fit nicely below the smaller search field, and just have the interactive suggestions appear on desktop only. What do you think?

Screenshot 2021-12-07 at 19 29 11@2x


Resources/Scripts/search_filter_suggestions.js
```diff
@@ -0,0 +1,102 @@
...
+  static formattedFilterDate() {
+    var thirtyDaysAgo = new Date()
+    thirtyDaysAgo.setDate(thirtyDaysAgo.getDate() - 30)
+
+    const year = thirtyDaysAgo.getFullYear()
+    const month = thirtyDaysAgo.getMonth() + 1 // Yes, JavaScript returns months as zero based.
+    const day = thirtyDaysAgo.getDate() // ... but not the day of the month. That's one based.
+
+    return `${year}-${String(month).padStart(2, '0')}-${String(day).padStart(2, '0')}`
+  }
```
daveverwer, line 101: This is some of the worst code I can remember writing, but JavaScript does not have freeform date formatting without introducing a dependency which I don't want to do. It does, and should continue to work, but oof... 😬

daveverwer:
> This obviously looks awful on mobile. I could tweak it slightly but it's never going to be great and I wonder if we should make this a desktop-only feature. There is one thing I considered that might help here, and which I'd love an opinion on. I added the suggestions with JavaScript as they rely on JavaScript to work, but the "Add filters to narrow search results (Learn more)." text doesn't. That could be added server-side, and also display on mobile as it'd fit nicely below the smaller search field, and just have the interactive suggestions appear on desktop only. What do you think?

I made a decision on this and implemented it. 😬 

The "Add filters to narrow search results (Learn more)." is now inserted server-side and so will always be visible with or without JavaScript and on all browser sizes. Then, the suggestions are appended but are hidden on mobile.

Screenshot 2021-12-07 at 21 31 22@2x

Screenshot 2021-12-07 at 21 31 48@2x

It's trivial to make these display on mobile, if that's what we decide we want.

daveverwer:
One thing that strikes me is that these filters in the snapshot tests are going to break them whenever time moves forward, and it has a habit of doing that... 😬

Resources/Scripts/search_filter_suggestions.js
finestructure: Missed opportunity to use `lpad` 😬

finestructure:
I think it'd be nice to have the tappable link on mobile because typing the `_` in particular is a bit annoying on a mobile keyboard. But I think it's fine to leave it out as well. People probably aren't triaging dependencies for their next big project on the phone :)

In a native app you'd probably put those as custom suggestions above the keyboard but I'm pretty sure there's no API for that for web text fields.

finestructure:
Regarding the snapshots: what if we use `Current.date()` on the Swift side to inject a

```html
<!-- searchFilterStartDate: 2021-11-08 -->
```

(or some other DOM construct) into the page that JS can read and use? That way we could fully control it without having to come up with JS dependency injection setup.

finestructure (approved):

daveverwer:
> In a native app you'd probably put those as custom suggestions above the keyboard but I'm pretty sure there's no API for that for web text fields.

You are correct

daveverwer:
Final tweaking. Fixes the snapshots issue and re-works the sentence so the Learn more link is at the end:

Screenshot 2021-12-08 at 08 50 58@2x

Also, no longer hidden on mobile:

Screenshot 2021-12-08 at 08 51 58@2x

Issue #1429, opened 2021-12-08
Add and populate `Package.platformCompatibility`


finestructure:
As discussed [here](https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/discussions/1403#discussioncomment-1766129), add a new field to `Package` where we roll up the platform compatibility from the build matrix.

finestructure:
Updating `Package.platformCompatibility` when we receive build reports will take care of keeping the field up-to-date going forward. Letting the service gradually update via incoming builds will not fill gaps for projects that aren't actively maintained - we need to backfill existing packages.

We have the following options:

1. run the some update sql as a one-off
2. backfill as part of the migration

Option 1 isn't great, because running ad hoc SQL isn't great. The good thing is the query can be run in small badges to limit impact and we can obviously test it on staging. I've run it locally on a copy and it should be fine.

Option 2 would be nice to tie it into the rollout but this would be a quite long-running migration (some minutes at least). Would be tricky to re-run and we obviously can't batch it.

I'd say we merge the PR with the schema change and the continuous update mechanism first and then run the manual SQL on staging for testing.

Before we deploy any of this to production we should make sure the search extension work as expected. That means between the first part of this landing in main and completing the whole task we shouldn't make any deployments. (Or maybe we should deploy from a branch on staging while we integrate everything - but that's a detail.)

The update SQL is as follows:

```sql
UPDATE packages p SET platform_compatibility = ARRAY(
	SELECT
		CASE
			WHEN b.platform LIKE 'macos-%' THEN 'macos'
			ELSE b.platform
		END
	FROM versions v
	JOIN builds b ON b.version_id = v.id
	WHERE v.package_id = p.id
	AND v.latest IS NOT NULL
	AND b.status = 'ok'
	GROUP BY b.platform
	HAVING count(*) > 0
),
updated_at = NOW()
WHERE p.id IN (
	SELECT id FROM packages
        -- adjust the following parameters as needed
	WHERE updated_at < '2021-12-08 12:27:05.82166+00'
	LIMIT 1
);
```

finestructure:
The query syntax for search using the new column is as follows:

```sql
select * from packages where platform_compatibility @> '{linux,macos}';
```

With the platform array it will be easy to support search for single or even multiple platforms, i.e. `platform:linux`, `platform:linux,macos`.

[...]

PR #1432, opened 2021-12-08
Search platform compatibility

finestructure:
⚠️ Schema change ⚠️

This adds a new column `platform_compatibility` to `packages`. The field is updated whenever the server receives a build report from one of the builders by running an update query. This will handle keeping the field up-to-date going forwards.

We *must not* merge this until we've decided what to do about `macosArm` (see https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/pull/1432/files#r764835140). As the PR stands right now we're adding a case we're not actually ever using.

See #1429 for post-merge/deployment SQL to update the newly added column.

Implements #1429

### DIFF
--- a/Sources/App/Models/Package.swift
+++ b/Sources/App/Models/Package.swift
@@ -15,6 +15,7 @@
 import Fluent
 import SemanticVersion
 import Vapor
+import SQLKit
 
 
 final class Package: Model, Content {
@@ -229,6 +230,34 @@ extension Package {
                 ? host + trunk
                 : host + "-" + trunk
         }
+    }
+}
+
+
+extension Package {
+    func updatePlatformCompatibility(on database: Database) -> EventLoopFuture<Void> {
+        guard let db = database as? SQLDatabase else {
+            return database.eventLoop.future(error: AppError.genericError(id, "Database must be an SQLDatabase ('as? SQLDatabase' must succeed)"))
+        }
+        return db.raw(
+            #"""
+            UPDATE packages p SET platform_compatibility = ARRAY(
+                SELECT
+                    CASE
+                        WHEN b.platform LIKE 'macos-%' THEN 'macos'
+                        ELSE b.platform
+                    END
+                FROM versions v
+                JOIN builds b ON b.version_id = v.id
+                WHERE v.package_id = p.id
+                AND v.latest IS NOT NULL
+                AND b.status = 'ok'
+                GROUP BY b.platform
+                HAVING count(*) > 0
+            )
+            WHERE p.id = \#(bind: id)
+            """#
+        ).run()
     }
 }
 
--- a/Sources/App/Models/Package.swift
+++ b/Sources/App/Models/Package.swift
@@ -36,7 +36,7 @@ final class Package: Model, Content {
     // data fields
 
     @Field(key: "platform_compatibility")
-    var platformCompatibility: [PlatformCompatibility]
+    var platformCompatibility: Set<PlatformCompatibility>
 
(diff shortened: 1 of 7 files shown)


Sources/App/Models/Package.swift
```diff
@@ -86,6 +92,19 @@ extension Package: Hashable {
 }
 
 
+extension Package {
+    enum PlatformCompatibility: String, Codable {
+        case ios
+        case macos
+        // TODO: decide if we want to include macosArm here
+        case macosArm
```
finestructure, line 100: If we keep `macosArm` we need to change the update SQL both in this PR and in the post-merge SQL. They currently map any `macos-%` to `macos`.

finestructure: FWIW, I'm leaning towards keeping the SQL simpler and dropping it. It's already not an important platform discriminator and it's going to be even less so in the future.

Sources/App/Models/Package.swift
```diff
@@ -215,6 +234,35 @@ extension Package {
...
+                FROM versions v
+                JOIN builds b ON b.version_id = v.id
+                WHERE v.package_id = p.id
+                AND v.latest IS NOT NULL
+                AND b.status = 'ok'
+                GROUP BY b.platform
+                HAVING count(*) > 0
+            ),
+            updated_at = NOW()
+            WHERE p.id = \#(bind: id)
```
finestructure, line 259: I've tried to write this update statement in Fluent but if it's possible at all it's going to be hard and potentially require upstream changes. But more importantly I believe there'd be some uncertainty whether that update runs on the server side or if it'd pulls records and send another update statement back.

Using raw SQL we _know_ we're doing it in the most efficient way. The downside is losing type-safety (see below).

Sources/App/Models/Package.swift
daveverwer: Certainly, for platform filters, we don't need it and it only makes the feature more complicated. I'd argue that at this stage, we should stop differentiating this across the whole system. If you're not compatible with *both* platforms, you're not compatible with macOS. Removing the whole feature is a discussion outside this PR, though.

Tests/AppTests/PackageTests.swift
```diff
@@ -354,6 +357,68 @@ final class PackageTests: AppTestCase {
...
+            // in combination with a plain TEXT[] backing field for
+            // platform_compatibility prevents us from relying on type safety.
+            // This test ensures that a newly added case in Build.Platform
+            // must also be handled in the updatePlatformCompatibility SQL
+            // statement.
+            // If it isn't, this test will fail with:
+            // invalid field: platform_compatibility type: Set<PlatformCompatibility> error: Unexpected data type: TEXT. Expected jsonb/json
+            // (which is a bit obscure but means that the content of
+            // platform_compatibility cannot be de-serialised into
+            // PlatformCompatibility)
```
finestructure, line 402: This is to deal with the lack of type-safety. We're essentially converting from `PackageShow.Model.PlatformCompatibility` to `Package.PlatformCompatibility` in the update SQL and can't rely on the compiler to check our transformation between the cases. But at least via this test we have that coverage, even if it's not at compile time.

finestructure:
Leaving this in draft for now because of the decision regarding `macosArm` but otherwise this is good to go.

Sources/App/Models/Package.swift
finestructure: Yep, agreed! I'll update the PR now.

daveverwer (approved):

PR #1435, opened 2021-12-08
Changed all RSS feeds to use a custom guid format that's truly unique and not URL-based

### LINKED ISSUE #1410: RSS feed GUIDs should be unique
Yes... a rather obvious bug 😅

Screenshot 2021-12-01 at 16 15 53@2x

The current `guid` is a link to the package and not the release, so RSS readers are marking items as read when they are not. We could use the `Version` database ID here, but the full release URL should also be truly unique.

The `link` element should also go to the same place.

daveverwer:
Fixes #1410


Sources/App/Core/RSS.swift
```diff
@@ -65,12 +65,16 @@ extension RSSFeed {
...
+        "\(repositoryOwner)/\(repositoryName)"
+    }
+
     var rssItem: Node<RSS.ChannelContext> {
         let link = SiteURL.package(.value(repositoryOwner),
                                    .value(repositoryName),
                                    .none).absoluteURL()
         return .item(
-            .guid(.text(link), .isPermaLink(true)),
+            .guid(.text(rssGuid), .isPermaLink(false)),
```
daveverwer, line 77: It's interesting that the default value for an omitted `isPermaLink ` is true so we need to explicitly set false here.

Tests/AppTests/RSSTests.swift
```diff
@@ -40,20 +44,21 @@ class RSSTests: SnapshotTestCase {
...
+                                               packageSummary: "packageSummary0",
+                                               createdAt: .t0).rssItem,
+                            RecentPackage.mock(id: .id1,
+                                               repositoryOwner: "repositoryOwner1",
+                                               repositoryName: "repositoryName1",
+                                               packageName: "packageName1",
+                                               packageSummary: "packageSummary1",
+                                               createdAt: .t1).rssItem
+                           ]
         )
```
daveverwer, line 60: I'm not sure there's any huge value in using the mock where every field is being specified, but it brings us up to date with using `.t0` and `.id0` helpers.

daveverwer:
This is ready to merge if you're happy with it 👍

finestructure (approved):

daveverwer:
Deploying as 2.51.1

PR #1444, opened 2021-12-10
Filter test updates

finestructure:
This removes the `resolveBinds:` parameter from `renderSQL` and adds some additional test integrating the search filters in the package search query.


Tests/AppTests/SearchFilterTests.swift
```diff
@@ -126,10 +126,10 @@ class SearchFilterTests: AppTestCase {
...
         let builder = SQLSelectBuilder(on: app.db as! SQLDatabase)
             .where(searchFilters: [filter])
-        let sql = renderSQL(builder, resolveBinds: true)
-        _assertInlineSnapshot(matching: sql, as: .lines, with: """
-        SELECT  WHERE ("stars" > '1')
-        """)
+        _assertInlineSnapshot(matching: renderSQL(builder), as: .lines, with: """
+            SELECT  WHERE ("stars" > $1)
+            """)
+        XCTAssertEqual(binds(builder), ["1"])
```
finestructure, line 132: These tests are somewhat redundant now that I added `test_packageMatchQuery_StarsSearchFilter` et al to `SearchTests`. Keeping them in for now but I have some follow-up changes for another day that'll address this.

Tests/AppTests/AppTestCase.swift
```diff
@@ -40,20 +40,14 @@ class AppTestCase: XCTestCase {
 
 
 extension AppTestCase {
-    func renderSQL(_ builder: SQLSelectBuilder?, resolveBinds: Bool = false) -> String {
-        renderSQL(builder?.query, resolveBinds: resolveBinds)
+    func renderSQL(_ builder: SQLSelectBuilder?) -> String {
+        renderSQL(builder?.query)
     }
```
finestructure, line 45: Removing the `resolveBinds:` parameter forces all tests to assert both the sql with the substitution parameters (`$1`, `$2`, ...) and the binds separately. A bit more work but the big advantage is that we can see and ensure that binds are always used and filled in as we expect.

daveverwer: Yea this is a great change. I can see how the original misunderstanding happened!

Tests/AppTests/AppTestCase.swift
```diff
@@ -64,9 +58,12 @@ extension AppTestCase {
...
         return serializer.binds.reduce(into: []) { result, bind in
-            if let bind = bind as? String { result.append(bind) }
-            if let bind = bind as? Date { result.append(DateFormatter.filterParseFormatter.string(from: bind)) }
-            if let bind = bind as? Int { result.append(String(bind)) }
+            switch bind {
+                case let bind as Date:
+                    result.append(DateFormatter.filterParseFormatter.string(from: bind))
+                default:
+                    result.append("\(bind)")
+            }
```
finestructure, line 66: There was a bug here where a new type (i.e. `Set<PlatformCompatibility>`) would be skipped from the bind variables. This lead me to debugging binds and ending down the rabbit hole with SQLBind and quoting...

Now we just default to `"\(bind)"` as a replacement which will always yield _something_ and we can then specialise it with a new case above.

daveverwer (approved):

PR #1446, opened 2021-12-10
Remove unused performance test

finestructure:
Also, there's a performance tests that's not actually working because we have no recorded baseline. I suspect this cannot work without an Xcode project to contain the perf baseline data - and it might be tricky to get that working reliably in CI where we may end up running on quite different machines. Skipping it for now until we've got something regarding perf tests in place (which we really should add).


daveverwer (approved):
The name of this PR is a little misleading now, but approved! 😂

PR #1450, opened 2021-12-13
Package update

finestructure:
Our SQLKit `UNION` extension has been upstreamed, although not quite with the same API that we proposed.


Sources/App/Core/Search.swift
```diff
@@ -255,14 +255,24 @@ enum Search {
...
+                                         filters: filters,
+                                         offset: offset,
+                                         limit: limit)
+
+            }).query
+        : packageMatchQueryBuilder(on: database,
+                                   terms: sanitizedTerms,
+                                   filters: filters,
+                                   offset: offset,
+                                   limit: limit).query
```
finestructure, line 275: I'm not really happy how this reads now but that's what the API reads like after review. It _is_ more expressive as it supports `... UNION ... UNION ALL ...`, which `db.unionAll(...)` doesn't but it _is_ making this a bit more awkward.

daveverwer: Yea I agree this is a little harder to read.

daveverwer (approved):

PR #1451, opened 2021-12-13
Added documentation for the Platform filter

### LINKED ISSUE #1448: Add platform filter documentation to FAQ

daveverwer:
Fixes #1448

### DIFF
--- a/Resources/Markdown/faq.md
+++ b/Resources/Markdown/faq.md
@@ -18,7 +18,7 @@ description: Frequently Asked Questions about the Swift Package Index
 - [Is there an API?](#api)
 - [Why are package versions missing?](#missing-versions)
 - [What is the SPI-Playgrounds app?](#spi-playgrounds-app)
-- [How can I filter my search results?](#search-filters)
+- [How can I filter search results?](#search-filters)
 
 ---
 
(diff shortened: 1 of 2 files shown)


Resources/Markdown/faq.md
```diff
@@ -146,50 +146,52 @@ If you don't have the SPI-Playgrounds app installed, you can [download it for ma
...
+You can search for packages created by a specific user or organisation by filtering on `author`. For example, a query of [`fluent author:vapor`](https://swiftpackageindex.com/search?query=fluent+author%3Avapor) matches any package containing the word "fluent" owned by the vapor organisation.
 
 Similarly, you can search for packages that contain keywords with a `keyword` filter. For example, [`keyword:accessibility`](https://swiftpackageindex.com/search?query=keyword%3Aaccessibility) matches any package having the "accessibility" keyword.
 
-#### Excluding results
+#### Querying Platform values
+
+You can search for packages that are compatible with a platform using a `platform` filter. Specify multiple platforms together with commas. For example, [`layout platform:ios,linux`](https://swiftpackageindex.com/search?query=layout+platform%3Aios%2Clinux) shows packages batching the word "layout" that are compatible with both iOS and Linux.
```
Sherlouk, line 191:
```suggestion
You can search for packages that are compatible with a platform using a `platform` filter. Specify multiple platforms together with commas. For example, [`layout platform:ios,linux`](https://swiftpackageindex.com/search?query=layout+platform%3Aios%2Clinux) shows packages matching the word "layout" that are compatible with both iOS and Linux.
```

daveverwer: Thanks James, nice catch!

Resources/Markdown/faq.md
```diff
@@ -146,50 +146,52 @@ If you don't have the SPI-Playgrounds app installed, you can [download it for ma
...
+You can search for packages created by a specific user or organisation by filtering on `author`. For example, a query of [`fluent author:vapor`](https://swiftpackageindex.com/search?query=fluent+author%3Avapor) matches any package containing the word "fluent" owned by the vapor organisation.
 
 Similarly, you can search for packages that contain keywords with a `keyword` filter. For example, [`keyword:accessibility`](https://swiftpackageindex.com/search?query=keyword%3Aaccessibility) matches any package having the "accessibility" keyword.
 
-#### Excluding results
+#### Querying Platform values
+
+You can search for packages that are compatible with a platform using a `platform` filter. Specify multiple platforms together with commas. For example, [`layout platform:ios,linux`](https://swiftpackageindex.com/search?query=layout+platform%3Aios%2Clinux) shows packages matching the word "layout" that are compatible with both iOS and Linux.
```
Sherlouk, line 191: This query clearly does work, but I wonder if it's the best example we could give? AutoLayout for Linux is an interesting concept 😅  I'm not sure the results we show are the best here.

finestructure (approved):

Resources/Markdown/faq.md
daveverwer: I'd be open to other queries as long as they don't repeat the same word as other examples, but the query isn't AutoLayout, it's "layout".

Sherlouk: I understand it's looking for "layout" and not "autolayout" but the results are mostly around AutoLayout which doesn't make a tonne of sense given the use cases. As I say, it's not wrong, just maybe not the best.

> don't repeat the same word as other examples

I was very intentional with that in my other examples 😛  Tried to give a variety of examples.

finestructure (approved):

Resources/Markdown/faq.md
daveverwer: Yep, I saw that and appreciated it. Any suggestions on what to use here instead of "layout"? I don't see it as a huge issue, but if you want to change it then you need to be the one suggesting a replacement 😂

Sherlouk: Perhaps "testing" for testing libraries which work on both iOS and Linux?

https://swiftpackageindex.com/search?query=testing+platform%3Aios%2Clinux

PR #1453, opened 2021-12-13
Increase the number of items in RSS feeds

### LINKED ISSUE #1443: Increase the number of items in `recent_releases` to 400
As discussed in #1366.

We may also want to increase `recent_packages` to 200 at the same time, although there's much less urgency on that one!

- [ ] Increase the number of rows in the view.
- [ ] Check that all those rows are output to the RSS feed.

daveverwer:
Fixes #1443 

⚠️ SCHEMA CHANGE ⚠️ 

I went to 500 instead of 400 because of round numbers!

Performance shouldn't be an issue because these are materialised. File size has obviously increased (releases is now ~750Kb) but the amount of traffic we get to the RSS feeds mean it should be inconsequential. If we need to reduce this, we should cache it with CloudFlare, but I think that would be overkill at this time.

### DIFF
--- a/Sources/App/Core/Constants.swift
+++ b/Sources/App/Core/Constants.swift
@@ -33,7 +33,7 @@ enum Constants {
     
     static let reIngestionDeadtime: TimeInterval = .minutes(90)
     
-    static let rssFeedMaxItemCount = 100
+    static let rssFeedMaxItemCount = 500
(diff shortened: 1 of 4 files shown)


finestructure (approved):

daveverwer:
Deploying as 2.53.0

Issue #1466, opened 2021-12-27
Package build report page gives a 404 for packages with a null package name


daveverwer:
From: https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/issues/1441#issuecomment-1000579168

Issue #1441, opened 2021-12-09
Missing Package name valentinradu/Firestarter


daveverwer:
I thought we fixed this, but we either didn't or this is a regression!

I went to add [Firestarter](https://swiftpackageindex.com/valentinradu/Firestarter) to the index tonight as it didn't show in search results only to find that it already existed in the `packages.json` file.

The package page is visible and everything looks good there, but the package is missing from search results:

- Search for [firestarter](https://swiftpackageindex.com/search?query=firestarter).
- Search for [valentinradu](https://swiftpackageindex.com/search?query=valentinradu).

The issue is that there's no package name on the `version` records, meaning the search view looks like this:

<img width="865" alt="Screenshot 2021-12-09 at 20 31 40@2x" src="https://user-images.githubusercontent.com/5180/145471286-39d556e2-155c-4165-9b72-64e0b103181d.png">

I think there are two issues here:

* The package name in the manifest is valid, and the JSON output from `dump-package` seems to be correct at a quick glance (see below).
* Packages that have no package name should still show up in search results. We should use the repository name if we don't have a package name like we do on the package page. Logged as #1442.

#### Package dump JSON

```json
{
  "cLanguageStandard" : null,
  "cxxLanguageStandard" : null,
  "dependencies" : [

  ],
  "name" : "Firestarter",
  "packageKind" : "root",
  "pkgConfig" : null,
  "platforms" : [
      ...
```

chris-araman:
I'm not sure if this is related, but my packages don't appear in search, either, although they appear to have been indexed:
https://swiftpackageindex.com/chris-araman/AsyncCloudKit
https://swiftpackageindex.com/chris-araman/CombineCloudKit

Names appear for both in `swift package dump-package`  and `xcodebuild -list -json` as expected.

Tangentially but possibly related, the "Full Compatibility Report" links yield 404 errors:
https://swiftpackageindex.com/chris-araman/AsyncCloudKit/builds
https://swiftpackageindex.com/chris-araman/CombineCloudKit/builds
https://swiftpackageindex.com/valentinradu/Firestarter/builds

finestructure:
Thanks for the additional into, @chris-araman ! I can see what the issue is: all these packages are referencing platform idioms like `macCatalyst` that we're not parsing yet:

```
[...]

PR #1467, opened 2021-12-27
Allow packages with NULL package names to appear in search results

### LINKED ISSUE #1442: Packages that have no package name should still show up in search results
See full details of this problem in #1441.

This ticket tracks the second part of this bug, that packages should not be excluded from search results just because they are missing a package name. We should use the repository name when we have no package name.

daveverwer:
Fixes #1442 

Does not address https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/issues/1441#issuecomment-1000579168 which is now logged as #1466.

Screenshot 2021-12-27 at 15 38 58@2x

Screenshot from a database where the Firestarter package name is NULL:

Screenshot 2021-12-27 at 15 39 33@2x


finestructure:
The screenshot is great to confirm that a nil package won't break the search display. I think we should add that case to the snapshot test `test_SearchShow`. I'm in the middle of changes which I can't easily stash right now but I can add that later on in case you're not around.

finestructure (approved):

daveverwer:
> The screenshot is great to confirm that a nil package won't break the search display. I think we should add that case to the snapshot test `test_SearchShow`. I'm in the middle of changes which I can't easily stash right now but I can add that later on in case you're not around.

Added an extra case to the snapshot.

PR #1468, opened 2021-12-27
Modify the build index and build show actions to cope with packages that don't have a package name

### LINKED ISSUE #1466: Package build report page gives a 404 for packages with a null package name

daveverwer:
Merge after #1467
Fixes #1466 

See my comments inline as I'm not sure what the best approach is.

### DIFF
--- a/Sources/App/Views/PackageController/Builds/BuildIndex+Model.swift
+++ b/Sources/App/Views/PackageController/Builds/BuildIndex+Model.swift
@@ -21,13 +21,13 @@ extension BuildIndex {
     struct Model {
         var owner: String
         var repositoryName: String
-        var packageName: String?
+        var packageName: String
         var completedBuildCount: Int
         var buildMatrix: BuildMatrix
 
         internal init(owner: String,
                       repositoryName: String,
-                      packageName: String?,
+                      packageName: String,
                       buildGroups: [BuildGroup]) {
             self.owner = owner
             self.repositoryName = repositoryName
@@ -42,7 +42,7 @@ extension BuildIndex {
 
             self.init(owner: packageInfo.repositoryOwner,
                       repositoryName: packageInfo.repositoryName,
-                      packageName: packageInfo.packageName,
+                      packageName: packageInfo.packageName ?? packageInfo.repositoryName,
                       buildGroups: buildGroups)
         }
 
--- a/Sources/App/Views/PackageController/Builds/BuildIndex+Model.swift
+++ b/Sources/App/Views/PackageController/Builds/BuildIndex+Model.swift
@@ -21,13 +21,13 @@ extension BuildIndex {
     struct Model {
         var owner: String
         var repositoryName: String
-        var packageName: String
+        var packageName: String?
         var completedBuildCount: Int
         var buildMatrix: BuildMatrix
 
         internal init(owner: String,
                       repositoryName: String,
-                      packageName: String,
+                      packageName: String?,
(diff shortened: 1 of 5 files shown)


Sources/App/Views/PackageController/Builds/BuildIndex+View.swift
```diff
@@ -27,7 +27,7 @@ enum BuildIndex {
         }
 
         override func pageTitle() -> String? {
-            "\(model.packageName) &ndash; Build Results"
+            "\(model.packageName ?? model.repositoryName) &ndash; Build Results"
```
daveverwer, line 30: It's obviously not ideal that we now leak logic for dealing with nil package names into the view. However, it also didn't feel right to populate a field named `packageName` with a repository name.

Sources/App/Views/PackageController/Builds/BuildShow+Model.swift
```diff
@@ -15,29 +15,28 @@
...
-                      packageName: packageName,
+                      packageName: result.version.packageName,
                       repositoryOwner: repositoryOwner,
                       repositoryName: repositoryName,
                       versionId: versionId)
         }
 
         internal init(buildInfo: BuildInfo,
-                      packageName: String,
+                      packageName: String?,
```
daveverwer, line 39: To fix it, we could keep the `packageName` field in this initialiser as non-optional, and deal with the substitution in one place on line 32 above.

This is certainly a cleaner implementation, but also doesn't feel perfect.

Sources/App/Views/PackageController/Builds/BuildIndex+Model.swift
```diff
@@ -37,14 +37,12 @@ extension BuildIndex {
         }
 
         init?(packageInfo: PackageController.BuildsRoute.PackageInfo, buildInfo: [PackageController.BuildsRoute.BuildInfo]) {
-            guard let packageName = packageInfo.packageName else { return nil }
-
```
finestructure, line 41: I kept existing behaviour here when I touch this last week but my intention was to fill it in with the owner here instead of letting it trickle into the views (and keep `packageName: String` instead of `String?`. I.e.

```swift
            self.init(owner: packageInfo.repositoryOwner,
                      repositoryName: packageInfo.repositoryName,
                      packageName: packageInfo.packageName ?? packageInfo.repositoryName,
                      buildGroups: buildGroups)
```

This would then also make the initialiser non-optional etc and all that.

Sources/App/Views/PackageController/Builds/BuildIndex+View.swift
finestructure: I think that's what you mean here, don't you? The reason I'd say that's ok is because `Model` isn't a db model but a transition view model where we prep the data for the view. We could rename the field but I don't think there's a better name, nor should it be confusing that the view's package name can fall back to the owner.

Sources/App/Views/PackageController/Builds/BuildShow+Model.swift
finestructure: I'd do the same here: keep it `String` and substitute in `repositoryOwner`.

daveverwer:
All switched around. Thanks Sven!

finestructure (approved):

PR #1469, opened 2021-12-28
Issue 1072 rework show query

### LINKED ISSUE #1072: Package page loads for packages with lots of releases are slow
There's a performance problem with package pages. Packages that have a large number of releases or products take quite a significant amount of time to load.

For example:

* [LeftPad](https://swiftpackageindex.com/daveverwer/LeftPad) - 3 releases - 170ms
* [Alamofire](https://swiftpackageindex.com/Alamofire/Alamofire) - 67 releases - 722ms
* [Vapor](https://swiftpackageindex.com/vapor/vapor) - 262 releases - 1,480ms
* [fastlane](https://swiftpackageindex.com/fastlane/fastlane) - 483 releases - 1,220ms

and

* [Soto](https://swiftpackageindex.com/soto-project/soto) - 265 libraries - 2,280ms
* [smoke-aws](https://swiftpackageindex.com/amzn/smoke-aws) - 32 libraries - 701ms

It's not a linear scale, as fastlane proves. But certainly, there is a correlation between page speed and the number of products/releases. 

After a quick chat with @finestructure, this isn't as simple as adding indexes! 😬 It's the deserialisation of all version objects that are used to generate the significant versions panel as well and the solution is not trivial.

However, since we only need the significant versions (stable, beta, branch) for the rest of the page, and everything else is just a count, and products are only ever displayed as a count, I wonder if we should limit our deserialisation to only the significant versions, and maybe de-normalise the number of releases, libraries, and executables into the products table and update it during analysis?

finestructure:
This breaks down the big "show route" query into individual queries with more targeted selections. The performance improvement is significant for "normal" packages like Alamofire and dramatic for packages with many libraries:

```
┌──────────────────────────────┐   ┌──────────────────────────────┐
│                              │   │                              │
│2.53.3                        │   │this branch:                  │
│                              │   │                              │
│Alamofire-package             │   │Alamofire-package             │
│-----------------             │   │-----------------             │
│Count:     100                │   │Count:     100                │
│Average:   0.121s             │   │Average:   0.066s             │
│Median:    0.119s             │   │Median:    0.065s             │
│Min:       0.112s             │   │Min:       0.062s             │
│Max:       0.169s             │   │Max:       0.082s             │
│Std dev:   0.008s             │   │Std dev:   0.003s             │
│90% Pctl:  0.127s             │   │90% Pctl:  0.070s             │
│                              │   │                              │
│soto-package                  │   │soto-package                  │
│------------                  │   │------------                  │
│Count:     100                │   │Count:     100                │
│Average:   2.652s             │   │Average:   0.096s             │
│Median:    2.647s             │   │Median:    0.096s             │
│Min:       2.606s             │   │Min:       0.091s             │
│Max:       2.751s             │   │Max:       0.110s             │
│Std dev:   0.028s             │   │Std dev:   0.003s             │
│90% Pctl:  2.689s             │   │90% Pctl:  0.098s             │
│                              │   │                              │
│                              │   │                              │
└──────────────────────────────┘   └──────────────────────────────┘
```

This is probably very hard to review at the code level. I've had to move lots of things around and there was no way to break this down into steps and keep it building/passing tests.

There is some chronology of the changes in the sequence of commits. I.e. it should be possible to follow the journey by looking at the steps as they each touch a part of the overall changes pretty much in isolation.

The change was functionally complete with commit f5f393bbafb40340e3721e2f50b115363eb2b8a3 when the tests all passed again. After that I removed code that was now unused as well as all the warning/deprecation reminders I'd left for myself.

The only functional change I've made was to require a default branch version for packages (see the comment in `test_show`). I believe this could be lifted but I think it'd be non-trivial undertaking and something I'd like to avoid.

It's probably easiest to simply run this branch locally and compare some packages to prod. Thanks to the test suite there shouldn't be any nasty surprises :)

Fixes #1072


Tests/AppTests/PackageController+routesTests.swift
```diff
@@ -17,107 +17,87 @@
...
 
-class PackageControllerTests: AppTestCase {
+class PackageController_routesTests: AppTestCase {
 
     func test_show() throws {
         // setup
         let pkg = try savePackage(on: app.db, "1")
         try Repository(package: pkg, name: "package", owner: "owner")
             .save(on: app.db).wait()
+        try Version(package: pkg, latest: .defaultBranch).save(on: app.db).wait()
```
finestructure, line 27: This is a new requirement: a package _must_ have a default branch version. This is true for all packages, i.e.

```sql
select p.url
from packages p
left join versions v on v.package_id = p.id
where v.id is null
```

returns no results on dev and prod.

I'd say it's ok to impose this. A package without a default version would be a package without any commits (or at least without any commits we can parse for some reason). 404ing on that would be ok even if it's a problem on our end, because we'd want to/need to fix that.

Alternatively, it should be possible to drop this requirement again but it'd require some digging into the mechanics of the underlying query and re-introduction of optionals in quite a few places.

daveverwer: I agree with the requirement completely, and can't see a situation where a package would not have a default branch. Is that even possible with git?

It's not clear to mr wheat would happen if we *did* somehow get a package in there with a default branch. What would the effect on that package in the index?

daveverwer (approved):

Tests/AppTests/PackageController+routesTests.swift
finestructure: No, that wouldn’t be possible with git. But it could theoretically happen if for instance our analyser crashed while getting the versions or if for some reason.

If that did happen, the package would not show up in any lists and would 404 if you navigated to its url.

finestructure:
Thanks for the tick! I’ll merge this now so we can let this run a bit on staging before actually deploying it.

PR #1474, opened 2021-12-31
Added `ExternalURL` to make linking to common external URLs more consistent

daveverwer:


Sources/App/Core/ExternalURL.swift
```diff
@@ -0,0 +1,31 @@
...
+
+    var description: String {
+        switch(self) {
+            case .projectBlog: return "https://blog.swiftpackageindex.com"
+            case .projectGitHub: return "https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server"
+            case .projectSponsorship: return "https://github.com/sponsors/SwiftPackageIndex"
+            case .raiseNewIssue: return "https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/issues/new/choose"
+        }
+    }
+}
```
finestructure, line 31:
```suggestion
enum ExternalURL: String, URLRepresentable {
    case projectBlog = "https://blog.swiftpackageindex.com"
    case projectGitHub = "https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server"
    case projectSponsorship = "https://github.com/sponsors/SwiftPackageIndex"
    case raiseNewIssue = "https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/issues/new/choose"

    var description: String { "\(rawValue)" }
}
```

finestructure (approved):

Sources/App/Core/ExternalURL.swift
daveverwer: Yea, I considered that but they didn't feel much like identifiers.

daveverwer: They're also not really descriptions, so it's not great either way.

PR #1479, opened 2022-01-06
Add search filter integration tests

finestructure:
Add some high level search filter integration tests.

### DIFF
--- a/Tests/AppTests/SearchTests.swift
+++ b/Tests/AppTests/SearchTests.swift
@@ -1063,6 +1063,14 @@ class SearchTests: AppTestCase {
             XCTAssertEqual(res.results.compactMap(\.packageResult?.repositoryName), ["1"])
         }
 
+        do {
+            // MUT
+            let res = try Search.fetch(app.db, ["test", "license:compatible"], page: 1, pageSize: 20).wait()
+
+            // validate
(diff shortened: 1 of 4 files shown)


Tests/AppTests/SearchTests.swift
```diff
@@ -903,6 +1116,79 @@ class SearchTests: AppTestCase {
...
+            .save(on: app.db).wait()
+        try Version(package: p2, packageName: "p2", reference: .branch("main"))
+            .save(on: app.db).wait()
+        try Search.refresh(on: app.db).wait()
+
+        // MUT
+        let res = try Search.fetch(app.db, ["test", "license:>mit"], page: 1, pageSize: 20).wait()
+
+        // validate
+        XCTAssertEqual(res.results.compactMap(\.packageResult?.repositoryName), [])
```
finestructure, line 1189: We may want to adjust this behaviour: right now it simply swallows the error and returns an empty list without any indication that the filter is invalid:

CleanShot 2022-01-06 at 08 59 17@2x

daveverwer: Oh yea. I think this should be logged as a bug. Maybe we need a set of operators that are valid for each type of filter field, or something like that.

daveverwer (approved):

PR #1486, opened 2022-01-13
Migrate ErrorMiddleware to async/await

finestructure:
As part of the whole `pg_dump` investigation I wanted to make sure the rollbar item is sent in a fire and forget event and it turned out to be way easier in async/await.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,3 +24,4 @@ Public/main.css.map
 
 .nova/
 /.vscode/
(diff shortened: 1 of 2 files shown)


finestructure:
Just marking this "draft" for now because I need to run more `pg_dump` tests and staging and don't want a deployment to kick in while I'm doing that :)

daveverwer (approved):

PR #1487, opened 2022-01-13
Report build runner on build reports

daveverwer:


Sources/App/Core/BuildRunner.swift
```diff
@@ -0,0 +1,41 @@
...
+    // Note: Runner IDs can be found under "Runners", here:
+    // https://gitlab.com/finestructure/swiftpackageindex-builder/-/settings/ci_cd
+
+    case linux1 = "WmqMUs6e"
+    case linux2 = "iE4SRjQi"
+    case mac0 = "pMpxem39"
+    case mac1 = "DVohN7oe"
+    case mac2 = "eFGZxpyH"
+
+    var description: String {
```
finestructure, line 32: Nit: might want to declare it as `CustomStringConvertible` conformer as well

finestructure (approved):

Sources/App/Core/BuildRunner.swift
daveverwer: Done

PR #1488, opened 2022-01-14
Bring back `NULL AS "keyword"` in package search subquery

finestructure:


Sources/App/Core/Search.swift
```diff
@@ -153,7 +153,7 @@ enum Search {
         let preamble = db
             .select()
             .column(.package)
-            .column(keyword)
+            .column(null, as: keyword)
```
finestructure, line 156: Bringing this back to re-establish "symmetry" between the search subqueries. Each subquery type has its column where it "sends" the data it finds and the other columns report back as `NULL`. I.e.

```
match_type | keyword | package_name | ... | repo_owner
package      NULL      foo                  bar
keyword      ios       NULL                 NULL
author       NULL      NULL                 bar
```

`package` being a slight exception in that it also needs the `repo_owner` field (which is the author field) to report back all of the data required for the package search result type.

What we're effectively doing is trying to create an enum in SQL such that we can `UNION ALL` different cases together and map them onto an `enum` case when we decode the rows into enums on the Swift side.

Writing this mostly as a reminder to myself and now that I've written it I think I'll need to find a place to add this as a comment around the search query constructs. It's quite complicated and easy to forget how this words.

daveverwer (approved):

PR #1496, opened 2022-01-18
Async ingest

finestructure:
We're getting a reproducible TSAN related failure:

```
Test Case 'IngestorTests.test_insertOrUpdateRepository' started at 2022-01-18 13:15:25.568
ThreadSanitizer:DEADLYSIGNAL
==8776==ERROR: ThreadSanitizer: SEGV on unknown address 0x7b680003ae20 (pc 0x7b680003ae20 bp 0x7b6000012d80 sp 0x7f7ca279de08 T8778)
==8776==The signal is caused by a READ memory access.
ThreadSanitizer:DEADLYSIGNAL
==8776==Hint: PC is at a non-executable region. Maybe a wild jump?
    #0 <null> <null> (0x7b680003ae20)

ThreadSanitizer can not provide additional info.
SUMMARY: ThreadSanitizer: SEGV (<unknown module>) 
==8776==ABORTING
error: terminated(1): /usr/bin/llvm-profdata merge -sparse -o /__w/SwiftPackageIndex-Server/SwiftPackageIndex-Server/.build/x86_64-unknown-linux-gnu/debug/codecov/default.profdata output:
    error: No input files specified. See llvm-profdata merge -help

make: *** [Makefile:36: test] Error 1
```

I can't tell if it's TSAN crashing or TSAN finding an issue. FWIW, it's passing with TSAN on macOS.

I've tried running this in Docker on arm64 but _every_ test fails with the following error:

```
error: terminated(66): /workspaces/spi-server/.build/aarch64-unknown-linux-gnu/debug/SPI-ServerPackageTests.xctest --dump-tests-json output:
    FATAL: ThreadSanitizer CHECK failed: /home/build-user/llvm-project/compiler-rt/lib/tsan/rtl/tsan_platform_linux.cpp:296 "((personality(old_personality | ADDR_NO_RANDOMIZE))) != ((-1))" (0xffffffffffffffff, 0xffffffffffffffff)
        #0 __tsan::TsanCheckFailed(char const*, int, char const*, unsigned long long, unsigned long long) /home/build-user/llvm-project/compiler-rt/lib/tsan/rtl/tsan_rtl_report.cpp:47 (SPI-ServerPackageTests.xctest+0x2fc3e0)
        #1 __sanitizer::CheckFailed(char const*, int, char const*, unsigned long long, unsigned long long) /home/build-user/llvm-project/compiler-rt/lib/sanitizer_common/sanitizer_termination.cpp:78 (SPI-ServerPackageTests.xctest+0x2767a0)
        #2 __tsan::InitializePlatform() /home/build-user/llvm-project/compiler-rt/lib/tsan/rtl/tsan_platform_linux.cpp:296 (SPI-ServerPackageTests.xctest+0x30328c)
        #3 __tsan::Initialize(__tsan::ThreadState*) /home/build-user/llvm-project/compiler-rt/lib/tsan/rtl/tsan_rtl.cpp:401 (SPI-ServerPackageTests.xctest+0x2ed508)
        #4 _dl_rtld_di_serinfo ??:? (ld-linux-aarch64.so.1+0xe9c0)
        #5 <null> <null> (ld-linux-aarch64.so.1+0x1140)
```

Need to try and repro this on an x86_64 machine.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -42,7 +42,10 @@ jobs:
       - name: Install unzip
         run: apt-get update && apt-get install -y unzip
       - name: Run tests
-        run: cp .env.testing.template .env.testing && make test
+        # temporarily run without TSAN due to false positives on Linux
+        # https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/pull/1496
+        # Review with Swift 5.6
+        run: cp .env.testing.template .env.testing && make test-without-tsan
         env:
           COLLECTION_SIGNING_PRIVATE_KEY: ${{ secrets.COLLECTION_SIGNING_PRIVATE_KEY }}
           DATABASE_HOST: postgres
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -42,10 +42,7 @@ jobs:
       - name: Install unzip
         run: apt-get update && apt-get install -y unzip
       - name: Run tests
-        # temporarily run without TSAN due to false positives on Linux
-        # https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/pull/1496
-        # Review with Swift 5.6
-        run: cp .env.testing.template .env.testing && make test-without-tsan
+        run: cp .env.testing.template .env.testing && make test
         env:
           COLLECTION_SIGNING_PRIVATE_KEY: ${{ secrets.COLLECTION_SIGNING_PRIVATE_KEY }}
           DATABASE_HOST: postgres
--- a/Makefile
+++ b/Makefile
@@ -36,6 +36,10 @@ test:
 	env RUN_IMAGE_SNAPSHOT_TESTS=true \
 	swift test --enable-code-coverage --disable-automatic-resolution --sanitize=thread
 
+test-without-tsan:
+	env RUN_IMAGE_SNAPSHOT_TESTS=true \
+	swift test --enable-code-coverage --disable-automatic-resolution
+
 test-fast:
 	@echo Skipping image snapshot tests
 	@echo Running without --sanitize=thread
--- a/Makefile
+++ b/Makefile
@@ -36,10 +36,6 @@ test:
 	env RUN_IMAGE_SNAPSHOT_TESTS=true \
 	swift test --enable-code-coverage --disable-automatic-resolution --sanitize=thread
 
-test-without-tsan:
-	env RUN_IMAGE_SNAPSHOT_TESTS=true \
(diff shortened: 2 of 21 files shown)


daveverwer (approved):

finestructure:
Interesting. I can reproduce the TSAN crash when I run the test in docker but it passes when I run it on a Linux host directly.

finestructure:
Seems to be a TSAN issue: https://bugs.swift.org/browse/SR-15444

I've made a little progress eliminating a couple of warnings but I'm unsure we can silence them all reliable, especially if we move to convert more code to a/a.

I'll investigate if we can disable TSAN individual tests. If that doesn't work, which I fear is going to be the case, we've got two options:

1. hold off merging this and converting more code for the time being
2. disable TSAN on Linux but make sure we run it during development on macOS (I have it enabled always now)

1. would be unfortunate, and it might be up to us to push to get this fixed. So by holding off, we'll be robbing us of our best chance to get this fixed.
2. would work but it assumes that TSAN issues are cross-platform. Conceptually I think that's true - it shouldn't be library implementation specific whether our code has data races.

finestructure:
The last few runs passed without TSAN errors. I'm not sure whether we can trust it though. I'll rebase and run it a few more times to see how reliable it is.

finestructure:
From running these tests many times on `lb1` I get the feeling that TSAN is sensitive to whether `await` calls are intermingled with `try EFL.wait()` calls, and how things are ordered. I've seen cases where I could turn a reliably failing TSAN into a success by inserting a call `await dummy()`, where `dummy` was simply:

```
func dummy() async { }
```

My concern is that if it's passing right now, we just happen to be in a semi-stable state.

Also, I've seen consistent warnings in `ErrorMiddlewareTests` when running on `lb1` (both directly on machine as well as in a docker container), even though that async/await change has been merged a week ago.

finestructure:
Ok, I _think_ I've figured it out. We need to make sure we have no `ELF.wait()` calls in async tests. They seem to be screwing up TSAN, making it unreliable. It doesn't always fail but seems to be increasing the likelihood of a false positive.

Leaving TSAN active for now, let's see how it goes.

finestructure:
After a few successful runs it now errored out again: https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/runs/4883987770?check_suite_focus=true

Hopefully this will get more stable as we remove more and more `.wait()` calls (if there are in fact the issue) but for now I'll disable TSAN.

finestructure:
I'm going to revert this, we're having a huge performance regression:

CleanShot 2022-01-21 at 12 19 57@2x

PR #1504, opened 2022-01-24
Async ingest

finestructure:
The `.and` conversions I made were [definitely supposed to be concurrent](https://forums.swift.org/t/vapor-update-concurrent/53274/8) and timings from real-world ingestion testing shows that the `mapAsync` loops should be concurrent as well:

Ingesting 50 packages timings:

release2.55.9: 4s
this branch: 3.2s
async ingest before these changes: 56s

I'll deploy to staging from this branch to confirm.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -42,7 +42,10 @@ jobs:
       - name: Install unzip
         run: apt-get update && apt-get install -y unzip
       - name: Run tests
-        run: cp .env.testing.template .env.testing && make test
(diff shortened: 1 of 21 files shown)


daveverwer (approved):

finestructure:
Ok, it was indeed just a misreading of the original implementation. I tested it and it came out as serial but that was probably an artefact of testing it from within an XCTest rather than by running the full command. Converting all the async Tasks I was doubtful about in to concurrent Tasks gives us identical performance to the ELF version:

CleanShot 2022-01-24 at 15 27 01@2x

PR #1519, opened 2022-01-30
Copy dependency button prototype

finestructure:
I've been creating a few packages recently and one of the annoying bits about adding a dependency has been to piece together the dependency declaration for the `Package.swift` file. We've got the info for it on our package page and I think it'd be really useful to expose a button next to the significant versions we're tracking to quickly copy a declaration.

This is a hacky _prototype_ with some obvious shortcomings but I wanted to see if this is something we could do quickly.

CleanShot 2022-01-30 at 17 09 46@2x

Obviously we wouldn't show the full text (I'm just showing it here because I'm not sure how to hide it while still allowing copying). Also, for some reason the `master` branch doesn't get a button although it has the correct `id` attribute. Must have something to do with it being the last element.

The button name perhaps isn't ideal but if we can add a tooltip it should work. It might be clear enough even so – just clicking it to try and then inspecting the clipboard would give people the idea. Maybe even just "Copy" would do.

Also, the parsing out of the dependency line from the url is a bit hacky but it might fly given that it's just a utility (and perhaps not long lived, once Xcode gains better support for package management).

I'm suspect this could easily be cleaned up if we wanted to add it. I'd personally find it really useful.


daveverwer (approved):
I think this is ready for review now @finestructure

Resources/Scripts/panel.js
```diff
@@ -0,0 +1,62 @@
...
+          panelElement.classList.remove('hidden')
+          event.preventDefault()
+
+          const panelUnderlayElement = document.createElement('div')
+          document.body.appendChild(panelUnderlayElement)
+          panelUnderlayElement.id = 'panel_underlay'
+          panelUnderlayElement.addEventListener('click', (event) => {
+            this.hidePanelElement(panelElement)
+          })
+        })
```
finestructure, line 51: Is it worth adding a plausible event based on some unique button id here so we can tell if/how these buttons are used? Or is that perhaps happening elsewhere in the callstack?

finestructure:
I can't attach comments to the snapshot but it looks like the build index page has regressed:

CleanShot 2022-02-11 at 03 23 51@2x

finestructure:
There's also a change in alignment on the package page:

CleanShot 2022-02-11 at 03 35 03@2x

and some vertical spacing change around the matrix (probably just the same root cause as on the build index page):

CleanShot 2022-02-11 at 03 33 35@2x

finestructure:
Other than that looks and works great!

Resources/Scripts/panel.js
daveverwer: I intentionally didn't add plausible events for opening or closing the panel, only clicking the buttons to copy the snippets or copy the URL. I think those are the key events rather than opening/closing the panel.

daveverwer:
> I can't attach comments to the snapshot but it looks like the build index page has regressed:
> 
> CleanShot 2022-02-11 at 03 23 51@2x
> 
> Only on mobile though - desktop shows up in the diff but it seems to be only a tiny difference in the search bar:
> 
> CleanShot 2022-02-11 at 03 28 58@2x

Yea this was an intentional change in 668217a0662aa8c04f6827439c7850e2ac2f6967. The gap (mobile only) before was way too big and I snuck this change in, which I shouldn't have 😂 I'll back this change out now and look at that gap in a separate PR.

Resources/Scripts/panel.js
finestructure: Ah cool, I actually meant the buttons 👍

Issue #1182, opened 2021-06-18
Prevent `METRICS_PUSHGATEWAY_URL ` environment error in Debug builds


daveverwer:
```
[ WARNING ] Environment variable not set: METRICS_PUSHGATEWAY_URL [component: server] Swift/ErrorType.swift:200: Fatal error: Error raised at top level: App.AppError.envVariableNotSet("METRICS_PUSHGATEWAY_URL")
```

Sherlouk:
❤️  every time this error catches me out

finestructure:
What's the dev use case where this warning actually shows up? I was just looking into this and realised I always run via Xcode and don't get this:

<img width="556" alt="CleanShot 2021-07-05 at 08 08 19@2x" src="https://user-images.githubusercontent.com/65520/124424960-2e6a7c80-dd68-11eb-9199-63dde8ae7b96.png">

Are you actually building a docker image and running it via compose? If so, my initial approach to detect a debug build might not work (I think the image is built in release mode). We also can't use `Environment.detect`, because `dev` is staging - so we have no local environment to suppress this on unless we introduce a new one and make sure it's set everywhere. I.e. it's easy to end up in the same situation as before if the env isn't correctly set.

daveverwer:
I don't get it in Xcode anymore, but I certainly have in the past, and it's certainly not when building with docker. Strange!

@selichter and @Sherlouk what are we doing wrong?

Sherlouk:
Aye, I think this must have been fixed by another PR because I'm not getting it when I build from `main`.

finestructure:
I'll go ahead and close this as magically fixed 😬  Please re-open if it appears again!

finestructure:
One thing to note about this one: this is a warning message and it does not crash for me:

```
❯ swift run Run ingest --limit 1
[0/0] Build complete!
[ INFO ] Ingesting (limit: 1) ... [component: ingest]
[ WARNING ] AppMetrics.push failed with error: connection reset (error set): Connection refused (errno: 61) [component: ingest]
```

I see it does/did for you, Dave:

```
[ WARNING ] Environment variable not set: METRICS_PUSHGATEWAY_URL [component: server]
Swift/ErrorType.swift:200: Fatal error: Error raised at top level: App.AppError.envVariableNotSet("METRICS_PUSHGATEWAY_URL")
make: *** [ingest] Trace/BPT trap: 5
```

[...]

PR #1522, opened 2022-01-31
Don't silence errors

finestructure:
```
❯ env swift run Run ingest --limit 1
[0/0] Build complete!
[ INFO ] Ingesting (limit: 1) ... [component: ingest]
[ WARNING ] Environment variable not set: METRICS_PUSHGATEWAY_URL [component: ingest]
~/P/S/spi-server on  log-metrics-push-errors
❯ echo $status
0
```

As mentioned in #1182

### DIFF
--- a/Sources/App/Commands/Analyze.swift
+++ b/Sources/App/Commands/Analyze.swift
@@ -55,10 +55,14 @@ struct AnalyzeCommand: Command {
 
         try Self.trimCheckouts()
 
-        try AppMetrics.push(client: client,
-                            logger: logger,
-                            jobName: "analyze")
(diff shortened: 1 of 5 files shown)


daveverwer (approved):
Approved but still pending the change to `analyze` we discussed this morning so I won't merge.

finestructure:
I've applied the same error handling to the ELF commands as the newer async/await ones: errors arising in `AppMetrics.push` will be logged as warnings but _not_ propagated up. I.e. `make analyze` will behave the same as `make ingest` does and exit cleanly.

Fixes #1182

PR #1523, opened 2022-02-02
Required reference

finestructure:


Sources/App/Migrations/044/UpdateVersionReferenceNotNullable.swift
```diff
@@ -0,0 +1,44 @@
...
+
+struct UpdateVersionReferenceNotNullable: AsyncMigration {
+    func prepare(on database: Database) async throws {
+        try await database.transaction { tx in
+            guard let db = tx as? SQLDatabase else {
+                fatalError("Database must be an SQLDatabase ('as? SQLDatabase' must succeed)")
+            }
+            try await db.raw(
+                #"DELETE FROM "versions" WHERE "reference" IS NULL"#
+            ).run()
```
finestructure, line 26: We don't currently have any rows `WHERE "reference" IS NULL` but there's technically nothing to prevent that. We therefore have a decision to make with respect to the reliability of the migration:

- alter the table and fail if there are rows that can't meet the new requirement
- ensure the migration will always succeed by fulfilling the precondition of no reference being NULL

In other cases where there's a clear default value (e.g. "score"), we run an update statement before the `ALTER TABLE` statement to set the field to 0 in case there are `NULL` values and we set a default on the column.

Here, there's not clear default. We could set `.branch("main")` but it's not a great default. I think it's better to reject saves without a reference entirely.

That also means we have no value to fill in for `NULL` fields (which we don't actually have). So instead I put in a `DELETE` which will effectively "run empty". The point is really to make the migration locally consistent and guarantee success even if somehow a version without a reference ended up in the table.

There is no harm deleting a version, especially one without a reference, which would be ignored by any of our processing, because if it was actually valid it would be recreated properly by the analysis stage.

Sources/App/Models/Version.swift
```diff
@@ -59,7 +59,7 @@ final class Version: Model, Content {
     var publishedAt: Date?
 
     @Field(key: "reference")
-    var reference: Reference?
+    var reference: Reference
```
finestructure, line 62: The are some other fields in `Version` that could be made non-optional at the same time, essentially the ones we assign when we create a `Version` in `getIncomingVersions`:

```swift
            return try Version(package: package.model,
                               commit: revInfo.commit,
                               commitDate: revInfo.date,
                               reference: ref,
                               url: url)
```

But making them non-optional would make using `Version` in tests a bit more awkward, because we'd need to fill these in everywhere or create a convenience initialiser with some sort of defaults.

Therefore I left it to `reference`, because it's the one we actually use the most and which has the most impact on cleaning up our code.

finestructure: Actually, I'll give it a try to see what the impact is. It's quite tempting to clean up at least `commit` and `commitDate`, which are closely related to `reference`.

finestructure: Turns out it's not too awkward if we introduce a convenience init that we use in tests only where we stub in some defaults for these newly required fields.

Sources/App/Migrations/044/UpdateVersionReferenceNotNullable.swift
daveverwer: Sounds good.

daveverwer (approved):
My PR is ready but I bet this will conflict with it so I'm going to merge this and then fix up the merge conflicts before putting mine in for review.

PR #1524, opened 2022-02-02
Added a real-time build monitor page

daveverwer:
There's no public link to this yet. I'd like to get it live on production for a little while before we add that, but this should be a pretty good start to getting more visibility on what the build system is doing.

### Screenshots

Screenshot 2022-02-02 at 14 27 35@2x
Screenshot 2022-02-02 at 14 28 01@2x
Screenshot 2022-02-02 at 14 28 28@2x
Screenshot 2022-02-02 at 14 28 44@2x


Sources/App/Models/Version.swift
```diff
@@ -132,6 +132,14 @@ final class Version: Model, Content {
...
         case preRelease = "pre_release"
         case release
+
+        var cssClass: String {
+            switch self {
+                case .preRelease: return "beta"
+                case .defaultBranch: return "branch"
+                case .release: return "stable"
+            }
+        }
```
finestructure, line 142: Nit: View-related stuff should probably live in a little extension near the view code under `App/Views` somewhere rather than in the model file itself.

Sources/App/Views/BuildMonitorController/BuildMonitorIndex+Model.swift
```diff
@@ -0,0 +1,145 @@
...
+        init?(buildResult: BuildResult) {
+            guard let id = buildResult.build.id,
+                  let createdAt = buildResult.build.createdAt
+            else { return nil }
+
+            let runner: BuildRunner? = {
+                guard let runnerId = buildResult.build.runnerId
+                else { return nil }
+                return BuildRunner(rawValue: runnerId)
+            }()
```
finestructure, line 63:
```suggestion
            let runner = buildResult.build.runnerId.flatMap(BuildRunner.init(rawValue:))
```

Sources/App/Views/PackageController/Builds/BuildShow+View.swift
```diff
@@ -103,14 +103,6 @@ private extension Build.Status {
 
     // There should only be "ok" or "failed" builds visible on this page (others don't have any details to display) so these other cases are only covered for completeness.
```
finestructure, line 104: I believe this comment came with `var text: String`, so it should perhaps move to `description` or go away, depending on whether `description` is really identical.

finestructure: (The comment makes it sound like `text` was rather specific to this page but it might be ok to generalise.)

Sources/App/Models/Version.swift
daveverwer: Ah, I meant to comment this myself when I opened the PR. This conversion is needed from two places, and didn't feel like it deserved a whole file just for this.

Tests/AppTests/BuildMonitorIndexModelTests.swift
```diff
@@ -0,0 +1,151 @@
...
+        let model = try XCTUnwrap(BuildMonitorIndex.Model(buildResult: buildResult))
+
+        XCTAssertEqual(model.packageName, "LeftPad from Repository.name")
+    }
+
+    func test_init_from_Build_with_no_package_name() throws {
+        do {
+            let package = try savePackage(on: app.db, "https://github.com/daveverwer/LeftPad")
+            let version = try Version(package: package
+                                      // Deliberately missing a `packageName`
```
finestructure, line 85: Perhaps self-document?
```suggestion
                                      packageName: nil
```

finestructure:
Looks great!

Tests/AppTests/BuildMonitorIndexModelTests.swift
daveverwer: We can add the `packageName: nil` in addition to the comment, if you like, but I think the comment is important. It wouldn't be clear that that line specifically is important over and above all the other setup.

Sources/App/Models/Version.swift
finestructure: I'd still move it out of the models though and keep any view related code in that hierarchy 🤷

Sources/App/Views/PackageController/Builds/BuildShow+View.swift
daveverwer: I left this comment as it also applies to the two other properties directly below this. All three previously ignored the `infrastructureError`. `triggered`, and `timeout` cases, and now only two do, but the comment is still important.

daveverwer: I also didn't duplicate this comment in the new implementation as the new implementation no longer ignores the other cases.

Tests/AppTests/BuildMonitorIndexModelTests.swift
finestructure: Well, I thought the test name `test_init_from_Build_without_repository_name` would help with that :)

daveverwer: All parts of the puzzle for sure, but I know that if/when I ever need to come back to this test it'll help me understand it.

daveverwer: I'll add the `packageName: nil` though

Sources/App/Views/PackageController/Builds/BuildShow+View.swift
finestructure: Ah, I see, I hadn't expanded the review snippet :+1:

Sources/App/Models/Version.swift
daveverwer: I moved it. There was no sensible place for these to live under Views so I made a new ModelExtensions directory.

PR #1537, opened 2022-02-07
Update package depedencies

finestructure:


.github/workflows/ci.yml
```diff
@@ -17,9 +17,12 @@ name: Continuous Integration
 on:
   workflow_dispatch:
   push:
+  pull_request_review:
+    types: [submitted]
 
 jobs:
   test-linux:
+    if: github.event_name == 'push' || github.event_name == 'workflow_dispatch' || (github.event.ref == 'updatePackageDepedencies' && github.event.review.state == 'approved')
```
finestructure, line 25: While the addition of `workflow_dispatch` now allows us to manually trigger CI runs on any branch, for some weird reasons their results are not accepted as a check.

What this here should do if it works as intended is run CI on approval of the `updatePackageDepedencies` (which is the auto-generated package update branch).

We can add the same for the package list nightly once this has proven to work.

PR #1564, opened 2022-02-16
Fix // parsing in packageDependencyCodeSnippet

### LINKED ISSUE #1563: Invalid Package URL
**Please describe the bug**

Okay so basically I spent like 25 minutes debugging why Xcode was being stupid until I realised that the URL I copied from SPI was actually invalid 😂 

1. Navigate to a package
2. Click on 'Use this package' button
3. See that URLs for versions (not the default branch) are missing one slash on the scheme of the URL (aka its https:/github.com instead of https://github.com thus making it an invalid URL)

**Screenshots**

Captured 2022-02-15 at 9 59 27

Chrome

- Platform: Desktop
- Browser: Chrome

**Additional context**

https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/blob/5004b010f2e1428f02fd1f2d0d15d348e2f02554/Tests/AppTests/__Snapshots__/WebpageSnapshotTests/test_PackageShowView_few_keywords.1.txt#L90

You can see this issue in the snapshot tests here.

finestructure:
Fixes #1563

### DIFF
--- a/Sources/App/Views/PackageController/PackageShow+Model.swift
+++ b/Sources/App/Views/PackageController/PackageShow+Model.swift
@@ -443,7 +443,8 @@ extension PackageShow.Model {
         if isSemVer {
             // url: "https://github.com/Alamofire/Alamofire/releases/tag/5.5.0"
             //  .package(url: "https://github.com/Alamofire/Alamofire.git", from: "5.5.0"),
-            let url = url.split(separator: "/").dropLast(3).joined(separator: "/") + ".git"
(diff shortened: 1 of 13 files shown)


finestructure:
I'll add an issue for us to actually construct this from data upstream and add it to me model properly rather than parsing it from bits, which I should have done in the first place.

daveverwer (approved):
Thanks for spotting and reporting this @Sherlouk!

PR #1567, opened 2022-02-17
Replace Rollup with Esbuild

daveverwer:
Using [esbuild](https://esbuild.github.io) is much faster in development, but more importantly, as the build tool is now a native binary there should be _far_ less spurious Dependabot alerts.

The size of the compiled JavaScript is up by 3k with this change (a 2% increase), but we're still only at ~147k which is still very small in 2022.

I also moved all of the front-end files out of the Resources directory with this PR. They were mixed in with markdown files, SVGs, certificates. That Resources directory had become a bit of a dumping ground. They now live in a new top-level directory `FrontEnd` and we shouldn't add anything in there that's doesn't contribute to the front-end build process.

The build process itself doesn't change as part of this, it's still just a case of running `yarn build` or `yarn serve`.

All snapshots changed but I couldn't see even a pixel different on any of them even using Kaleidoscope's Difference view.


finestructure:
Wow, I had to really squint to see those snapshot differences :)

CleanShot 2022-02-17 at 11 56 51@2x
CleanShot 2022-02-17 at 11 57 36@2x
CleanShot 2022-02-17 at 11 57 54@2x

finestructure (approved):

daveverwer:
My eyes aren't good enough to see that :)

daveverwer:
Deploying as 2.57.7

PR #1568, opened 2022-02-17
Widen keyword search

### LINKED ISSUE #1193: Widen keyword search
Via https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/pull/1190#discussion_r658497753

The initial keyword search implementation is doing simple `"keyword" = 'test'` filtering with LIMIT 1 when searching for `test`.

We'll want to also match `unit-testing`, changing that to `"keyword" like '%test%'` with some limit on the results, say `LIMIT 3`. When we do that we'll also need to `DISTINCT` the results (because the `UNNEST` is generating lots of duplicates).

More importantly, we'll need to define a sort order of partial keyword matches. I'm thinking something like

1. exact match
2. partial matches `%test%` ordered by number of occurrences, alphabetically

finestructure:
CleanShot 2022-02-17 at 15 58 57@2x

Fixes #1193


Sources/App/Core/Search.swift
```diff
@@ -181,28 +183,31 @@ enum Search {
...
+            .column(SQLFunction("LEVENSHTEIN", args: keyword, SQLBind(mergedTerms)),
+                    as: levenshteinDist)
             .from(searchView)
             .from(SQLFunction("UNNEST", args: keywords), as: keyword)
-            .where(keyword, .equal, mergedTerms)
-            .limit(1)
-        // TODO: increase limit when we do % matching
+            .where(keyword, ilike, SQLBind(searchPattern))
+            .orderBy(levenshteinDist)
+            .limit(10)
```
finestructure, line 210: I picked 10 because it seemed like a sensible figure but I didn't check what the distribution of keyword counts actually looks like. I'd like to have _something_ here but it doesn't have to be 10.

Sources/App/Migrations/045/CreateExtensionFuzzyStrMatch.swift
```diff
@@ -0,0 +1,36 @@
+// Copyright 2020-2021 Dave Verwer, Sven A. Schmidt, and other contributors.
```
finestructure, line 1: Just noticed this, we need to bump this in all files, don't we?

Tests/AppTests/SearchTests.swift
```diff
@@ -52,8 +52,8 @@ class SearchTests: AppTestCase {
             filters: [try AuthorSearchFilter(expression: .init(operator: .is, value: "foo"))]
         )
 
-        _assertInlineSnapshot(matching: renderSQL(b), as: .lines, with: """
-              SELECT 'package' AS "match_type", NULL AS "keyword", "package_id", "package_name", "repo_name", "repo_owner", "score", "summary", "stars", "license", "last_commit_date", "last_activity_at" FROM "search", CONCAT("keywords") AS "keyword" WHERE CONCAT_WS(' ', "package_name", COALESCE("summary", ''), "repo_name", "repo_owner") ~* $1 AND "repo_owner" IS NOT NULL AND "repo_name" IS NOT NULL AND ("repo_owner" ILIKE $2) ORDER BY LOWER("package_name") = $3 DESC, "score" DESC, "package_name" ASC
+        XCTAssertEqual(renderSQL(b), """
```
finestructure, line 55: I've changed these back to `XCTAssertEqual` because with the recent Kaleidoscope update that lets you diff XCTest output it's actually now easier again to see changes _that_ way.

Sources/App/Core/Search.swift
daveverwer: In my testing (just using SQL) yesterday, 15 felt good but we can easily change it once it's live.

We may even want to implement something like we have on the package page where they all come back and are hidden by some javascript so they can be expanded if people want to see everything.

daveverwer (approved):

finestructure:
Making sure we only pad the keyword search string with `%` if it's non-empty now:

CleanShot 2022-02-18 at 03 49 58@2x

PR #1569, opened 2022-02-18
Widen author search

finestructure:
CleanShot 2022-02-18 at 03 59 56@2x

PR #1570, opened 2022-02-18
Extend package search by searching keywords

### LINKED ISSUE #1194: Match search against `package.keywords`
Via https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/pull/1190#discussion_r658510334

The initial keyword search implementation is _not_ matching against a package's keywords.

One simple way to do this is by appending the topics to the search string:

```sql
WHERE CONCAT_WS(' ', "package_name", COALESCE("summary", ''), "name", "owner", ARRAY_TO_STRING(keywords, ' ')) ~* 'foo'
```

I _think_ we'll want to do that. In practise it might pull in a lot of results for certain terms, like for instance `test`, but I suppose that's a good thing. You can sort of do a keyword search without actually picking the keyword result from the search results.

Follow up to #1179

finestructure:
Merge after #1569 

Fixes #1194

This is how this affects results for example:

CleanShot 2022-02-18 at 04 13 01@2x

CleanShot 2022-02-18 at 04 14 59@2x

One downside with this is that it's not obvious how "SwifterSwift" earned its position - there's no mention of `test` in the displayed result. It'll only be somewhat obvious on click-through that the result is "on topic".

On the other hand "ViewInspector", which has `unit testing` in its description gets pushed further down, because its `score` (probably driven mainly by stars) is lower.

Not sure how big a problem that is. I don't think displaying all keywords in the results is a solution. When we showed them at first it looked really busy. Maybe should show matching keywords in the result to make the matching clearer.


daveverwer:
I think this is an important feature and will be essential for high-quality search, but I agree we need to think about the display in some way. Let me play with some options for this early next week.

daveverwer (approved):

daveverwer:
Deploying as 2.58.2

PR #1574, opened 2022-02-21
Enable author and keyword search results on prod

### LINKED ISSUE #1320: Final steps for adding author and keyword search
We’re so close to this feature! @finestructure and I chatted through the last couple of bits this morning and here’s a list:

* Search should return multiple matches for authors from a query:
	* We spoke about using some kind of string distance function for ordering of these results. From looking at the documentation, the “Levenshtein” method that Postgres supports might be a good primary ordering, maybe with a number of packages as a secondary sort order.
* Search should return multiple matches for keywords from a query:
	* Ordered by number of packages with that tag attached
	* If we can pull the number out as well, we should. We can display it next to the tag. 

There’s also #1194 but I’m not sure I understand that issue. Aren’t the package’s keywords where we’re pulling this single keyword result from? If it doesn’t match against keywords from a package, where are the keywords coming from?

finestructure:
Fixes #1320

### DIFF
--- a/Sources/App/Core/Search+Result.swift
+++ b/Sources/App/Core/Search+Result.swift
@@ -22,11 +22,6 @@ extension Search {
         case package(PackageResult)
 
         init?(_ record: DBRecord) {
-            // don't show non-package results on production yet
-            if Environment.current == .production && !record.isPackage {
-                return nil
-            }
-            // -- end --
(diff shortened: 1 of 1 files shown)


daveverwer (approved):

daveverwer:
Fixes #1320

PR #1577, opened 2022-02-21
Display matching keywords in package search results

daveverwer:
Merges into #1570 

Where a keyword contributes to a package being shown in search results, that keyword (and only that keyword) is now shown in the package list item. The exception to this is where the number of keywords being shown is clipped, and we'll need to increase this from 15 before this is complete, but that's not part of this PR.

I'm also not very happy with how these tags display. The border on the left of the package rows looks strange with the tags also having that same left border. They were never really designed to go inside each other. 😬

Screenshot 2022-02-21 at 19 41 12@2x
Screenshot 2022-02-21 at 19 41 38@2x


daveverwer:
I had an idea to help with the keyword display problem and I quite like it. See screenshots below but adding "Matching keywords:" makes it much clearer that this is not *all* keywords and also helps to solve the border inside a border problem as it pushes the keyword display further to the right away from where the eye notices it.

Screenshot 2022-02-21 at 20 01 10@2x

Screenshot 2022-02-21 at 20 01 27@2x

finestructure (approved):
Great!

Issue #1462, opened 2021-12-20
Add a true `staging` environment setting rather than depending on `development`


daveverwer:
Currently, we rely on the default `production` and `development` environments supplied with Vapor:

https://github.com/vapor/vapor/blob/73d57959a96ceca719558e70c320b2f729d2b25b/Sources/Vapor/Environment/Environment.swift#L90-L96

This has worked well enough for now but has always been a little limiting and now there's a concrete benefit from adding a third environment in #1182, let's consider adding a dedicated staging environment.

finestructure:
I'm not quite sure what the intention here was. `development` *is* staging. We've got the following three environments:

- production → swiftpackageindex.com
- development → staging.swiftpackageindex.com
- testing → tests or local env

Since #1182 was fixed without this, should we close this?

daveverwer:
> `development` _is_ `staging`

That's my point. It's odd to me that those two environments are the same.

In previous projects I have created/worked on there have been four environments (development, staging, production, testing) and it's useful to be able to differentiate between development (where you might not want to do something like report metrics, as we saw in that issue).

finestructure:
Ah, I see. I've always called that dev, hence `spi_dev` etc everywhere. If anything I'd have called it `dev.swiftpackageindex.com` 🤷‍♂️🙂

finestructure:
Closing this for now.

PR #1578, opened 2022-02-22
Assert query plan

finestructure:
This is now working and the results are stable across runners (but _not_ stable over time as staging content changes - but that's part of what we're testing).

I ran the test with an expected cost set low enough to trigger a failure both locally and in the CI to compare:

CI:
```
    ✖ test_packageMatchQuery, failed - Total cost of 266.48 above threshold of 70.0 (incl variation)
2477
```

M1 MacBook Pro:
```
/Users/sas/Projects/SPI/spi-server/Tests/AppTests/QueryPerformanceTests.swift:41: error: -[AppTests.QueryPerformanceTests test_packageMatchQuery] : failed - Total cost of 266.48 above threshold of 70.0 (incl variation)
```


.github/workflows/ci.yml
```diff
@@ -50,11 +50,29 @@ jobs:
...
+          DATABASE_HOST=${{ secrets.STAGING_DATABASE_HOST }}
+          DATABASE_NAME=spi_dev
+          DATABASE_PASSWORD=${{ secrets.STAGING_DATABASE_PASSWORD }}
+          DATABASE_PORT=5432
+          DATABASE_USE_TLS=true
+          DATABASE_USERNAME=spi_dev
+          " >> .env.staging
+
+          make test-without-tsan
+          make test-query-performance
```
finestructure, line 73: Instead of running the two tests back to back we could also add another job specifically to run the query performance tests. I've put them together here because build time is ~8 mins vs test time of ~1 min.

I think if we decide to do this, we _should_ set up another job and schedule it to run daily only, in which case the extra build time doesn't matter really.

finestructure: I'll clean this up once we decide to go forward and restore the original job to what it was before.

Makefile
```diff
@@ -45,6 +45,13 @@ test-without-tsan: xcbeautify
 	   swift test --enable-code-coverage --disable-automatic-resolution \
 	2>&1 | ./xcbeautify
 
+test-query-performance: xcbeautify
+	set -o pipefail \
+	&& env RUN_QUERY_PERFORMANCE_TESTS=true \
+	   swift test --enable-code-coverage --disable-automatic-resolution \
+	   --filter QueryPerformanceTests \
+	2>&1 | ./xcbeautify
+
```
finestructure, line 54: If we set this up as a daily job we could get rid of this extra target entirely. It wouldn't matter if we _also_ run all the other tests. We'd just invoke

```
env RUN_QUERY_PERFORMANCE_TESTS=true make test-without-tsan
```

Tests/AppTests/Util.swift
```diff
@@ -26,8 +26,13 @@ private var _schemaCreated = false
 
 func setup(_ environment: Environment, resetDb: Bool = true) throws -> Application {
     let app = Application(.testing)
+    let host = try configure(app)
+
+    // Ensure `.testing` refers to "postgres" or "localhost"
+    precondition(host == "localhost" || host == "postgres", ".testing must be a local db, was: \(host)")
```
finestructure, line 32: This is very important to note: while testing this PR I discovered that once an `Application` is instantiated via `Application(.testing)`, it is _not_ possible to instantiate another `Application(.staging)` and have it resolve to a different database.

When I did so, the query performance tests kept failing, because even though `app` was instantiated via `Application(.staging)`, the database host the test was connecting to was still `postgres` (i.e. the test db docker container).

That _also_ means that had I for some reason run the query performance tests first, thereby instantiating `Application(.staging)` first and then within the same test run run _another_ test that tries to set up the test db... 💣 I would have torn down staging...

There are now a couple of safe-guards in place against this (this `precondition` and the `RUN_QUERY_PERFORMANCE_TESTS=true` env variable). We can go further and pull this test out into its own target entirely to add another layer of protection.

daveverwer: As long as we can never run the main set of tests against the staging database 😅

daveverwer (approved):

Tests/AppTests/Util.swift
finestructure: Yeah, I'll make it so it can't be done accidentally :)

I'll tidy this up over the coming days and baseline the relevant queries.

daveverwer: It's probably not a good fit for doing at the same time but this discussion also reminded me of #1462 in this area.

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,42 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+name: Query Performance Test
+
+on:
+  workflow_dispatch:
+  push:
```
finestructure, line 19: To do: add `schedule` instead of `push` after testing is complete

finestructure:
I've now rebased this branch on `main`. I set the baseline specifically based on a7a82a700e5533d8722c3c589ce2333d0b686719 (merge after 2.58.1) to capture the baseline before the latest search changes merged in via #1570 and it's caught the first regressions:

```
    ✔ test_Search_authorMatchQuery (0.891 seconds)
    ✔ test_Search_keywordMatchQuery (0.057 seconds)
    ✔ test_Search_packageMatchQuery (0.051 seconds)
    ✔ test_Search_query_authorFilter (0.057 seconds)
    ✖ test_Search_query_keywordFilter, failed - Total cost of 3921.07 above threshold of 3900.0 (incl variation)
    ✖ test_Search_query_lastActicityFilter, failed - Total cost of 3921.02 above threshold of 3900.0 (incl variation)
    ✔ test_Search_query_licenseFilter (0.077 seconds)
    ✖ test_Search_query_noFilter, failed - Total cost of 3909.57 above threshold of 3900.0 (incl variation)
    ✖ test_Search_query_platformFilter, failed - Total cost of 3800.42 above threshold of 3800.0 (incl variation)
    ✖ test_Search_query_starsFilter, failed - Total cost of 3826.71 above threshold of 3800.0 (incl variation)
```

.github/workflows/nightly.yml
finestructure: This runs faster than the regular CI and I think it makes more sense to run it in parallel in fact. If there are changes being introduced that cause a performance regression we'll want to find out while the PR is up and not have to figure out what it was a day later.

Tests/AppTests/Util.swift
finestructure: I suspect this was related to https://github.com/vapor/vapor/issues/2755, which is now fixed. But it’s good to have the safeguard in place anyway.

PR #1580, opened 2022-02-22
Make the overflowing keyword list generic so it can be used elsewhere

daveverwer:
No functional changes here, but it will allow us to make any list of anything collapsable

Screenshot 2022-02-22 at 16 14 42@2x
.

### DIFF
--- a/Sources/App/Views/Plot+Extensions.swift
+++ b/Sources/App/Views/Plot+Extensions.swift
@@ -52,7 +52,7 @@ extension Node where Context: HTML.BodyContext {
     }
 
     static func spiOverflowingList(_ nodes: Node<HTML.ListContext>...) -> Self {
-        .element(named: "spi-overflowing-ul", nodes:[ .ul(.group(nodes)) ])
+        .element(named: "spi-overflowing-list", nodes:[ .ul(.group(nodes)) ])
     }
 
     static func spinner() -> Self {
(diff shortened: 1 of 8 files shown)


Sources/App/Views/Plot+Extensions.swift
```diff
@@ -51,6 +51,22 @@ extension Node where Context: HTML.BodyContext {
         .element(named: "tab-bar", nodes: nodes)
     }
 
+    static func spiOverflowingList(overflowMessage: String,
+                                   overflowHeight: Int,
+                                   _ nodes: Node<HTML.ListContext>...) -> Self {
+        // Note: The `overflowHeight` is a magic number that needs some experimentation each
+        // time this tag is used. It's the exact size in pixels of the collapsed element. If
+        // incorrect, the "show more" button will show up unnecessarily as it may overflow by
+        // just a single invisible pixel.
```
finestructure, line 60: 👌 

CleanShot 2022-02-22 at 17 18 20@2x


https://github.com/SwiftPackageIndex/Pipelines/blob/295e2bdde86abdf8d95fd999a5fab0a7f05fc2eb/Sources/HomeView/SidebarView.swift#L367

finestructure (approved):

Sources/App/Views/Plot+Extensions.swift
daveverwer: Magic numbers are fine as long as you explain them... Right? 😂

PR #1586, opened 2022-02-23
Potential fix for #1584

### LINKED ISSUE #1584: Visual Bug with Code Snippets
**Please describe the bug**

I'm finding that on one of my machines there's a bit of a weird visual bug with code snippets (see screenshot below).

Screenshot 2022-02-23 at 3 12 22 pm

This happens to all code snippets, both in our static pages (Add a Package) and dynamic pages like READMEs.

Relates slightly to this comment: https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/issues/1572#issuecomment-1046311481

1. Navigate to a page which contains a code snippet, example: https://swiftpackageindex.com/add-a-package
2. Depending on your device, you might see the weird visual issue on the code snippet.

If I turn off `overflow: scroll` for the `<pre>` tag via "Inspect Element" then it goes back to what I would call the expected behaviour.

Screenshot 2022-02-23 at 3 14 20 pm

- Platform: Desktop
- Browser: Chrome (Version 92.0.4515.107)

**Other information**

On affected machines, it happens 100% of the time. However, the number of devices it impacts appears to be low.
Of all my machines, only one is showing it - and given how Chrome syncs across devices essentially everything is the same on these two machines which is confusing me.

daveverwer:
Fixes #1584 

I'm not sure this is a good change but maybe you can pull it locally @Sherlouk and test it?

### DIFF
--- a/FrontEnd/styles/build_logs.scss
+++ b/FrontEnd/styles/build_logs.scss
@@ -20,5 +20,4 @@
 #build_log {
   white-space: pre-wrap;
   height: 40vh;
-  overflow: scroll;
 }
--- a/FrontEnd/styles/build_logs.scss
+++ b/FrontEnd/styles/build_logs.scss
@@ -20,4 +20,5 @@
 #build_log {
   white-space: pre-wrap;
   height: 40vh;
+  overflow: scroll;
 }
--- a/FrontEnd/styles/base.scss
+++ b/FrontEnd/styles/base.scss
@@ -77,7 +77,7 @@ pre {
   font-size: 14px;
   border-radius: 4px;
   padding: 15px;
-  overflow-x: scroll;
+  overflow: auto;
 }
 
 code {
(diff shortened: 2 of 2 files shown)


FrontEnd/styles/build_logs.scss
```diff
@@ -20,4 +20,5 @@
 #build_log {
   white-space: pre-wrap;
   height: 40vh;
+  overflow: scroll;
```
daveverwer, line 23: We need `overflow` on both directions here as the logs always scroll both ways.

Sherlouk:
Screenshot 2022-02-24 at 9 23 14 am

As may have been expected, it fixes the scrollbar on the one axis we've disabled scrolling for.

I've locally tried using `overflow: auto;` and this appears to fix the visual issue when scrollbars are not needed, but allows both axis to be scrollable when necessary. Perhaps worth trying?

daveverwer:
See the thread in #1584 for details on why we can close this.

daveverwer:
Re-opened as the change does make things slightly better for people with non-Apple pointing devices so we should merge this. There's no downside for other people.

finestructure (approved):

PR #1599, opened 2022-03-01
Issue 1585 widened search

### LINKED ISSUE #1585: Widened searches impacted search filters
Search for `keyword:app` doesn't seem to filter _only_ on packages with that keyword.

Noticed this for keyword filters, might affect others as well. I'm a bit worried that this didn't trigger a test failure. That means we need to extend coverage a bit.

https://staging.swiftpackageindex.com/search?query=a+keyword%3Aapp

CleanShot 2022-02-23 at 17 34 34@2x

finestructure:
Fixes #1585


Sources/App/Core/SearchFilter/Filters/KeywordSearchFilter.swift
```diff
@@ -33,8 +35,13 @@ struct KeywordSearchFilter: SearchFilterProtocol {
...
                 .caseInsensitiveLike : .notCaseInsensitiveLike,
-            bindableValue: .value("%\(expression.value)%"),
+            bindableValue: .value(expression.value),
             displayValue: expression.value
         )
     }
+
+    var sqlIdentifier: SQLExpression {
+        // override default identifier
+        arrayToString(Search.keywords, delimiter: " ")
```
finestructure, line 45: This actually only works for single keywords, which is why our test passes. I.e.

```sql
select ARRAY_TO_STRING(ARRAY['foo'], ' ') ILIKE 'foo';
```
→ TRUE

```
select ARRAY_TO_STRING(ARRAY['foo', 'bar'], ' ') ILIKE 'foo';
```
→ FALSE

which makes sense.

finestructure: I've updated `test_keywordSearchFilter` to expose this in 76101d78c72faf2696cf073717dca44202eb5798

daveverwer (approved):

Sources/App/Core/SearchFilter/Filters/KeywordSearchFilter.swift
```diff
@@ -33,8 +35,16 @@ struct KeywordSearchFilter: SearchFilterProtocol {
...
         )
     }
+
+    var sqlIdentifier: SQLExpression {
+        predicate.sqlBind
+    }
+
+    var sqlBind: SQLExpression {
+        any(Search.keywords)
+    }
```
finestructure, line 49: Right now I've fixed this by making `sqlIdentifier` and `sqlBind` return "flipped" expressions. I.e. since both are `SQLExpression`s, it doesn't really make a difference which one is the bind value and which one the column expression, apart from them being misnamed.

It's probably best if we merge this as is for the moment and I'll follow up with a PR that renames both those protocol getters, to reduce the noise in this PR.

finestructure:
Costs have improved a bit. Not dramatically but noticeably.

CleanShot 2022-03-01 at 15 56 46@2x

PR #1603, opened 2022-03-02
Fix incremental build regression

### LINKED ISSUE #1602: Incremental build & run regression
Incremental (probably also affecting full builds but we don't notice it as much) builds have slowed down in recent weeks. It seems to be tied to the changes in `Search.swift`.

### Test process

- Run `time swift test --disable-automatic-resolution --filter SearchTests.test_DBRecord_packageURL` once to trigger initial build
- Add dummy line `static let foo = "bar"` to `Search` (just below `static let schema = "search"`)
- Run `time swift test --disable-automatic-resolution --filter SearchTests.test_DBRecord_packageURL` again

### Incremental

2.58.6	34s
2.58.2	34s	<- further regression (match against package keywords – Search.swift changes?)
2.58.1	13s
2.58.0	13s	<- further regression (widen author search – Search.swift changes?)
1d9ad8	10s	<- first regression (widen keyword search – Search.swift changes?)
2.57.7	6s
2.57.6	6s
2.57.4	6s
2.57.0	7s
2.56.0	7s
2.54.0	8s
2.51.0	7s

NB: this test is impossible to run with Xcode, at least with the code as it was at the time, because it insists to update packages when switching branches. Xcode does not have an option like `--disable-automatic-resolution` to my knowledge.

NB: Some older revisions like 2.51.0 fail to build with Swift 5.5.3 because they had a section in `XCTestCase+runAsyncTest.swift` which doesn't actually compile with newer Swift versions:

```swift
#if compiler(>=5.5.2)
    // review and remove with Swift 5.5.2, this workaround should be unnecessary by then
    // https://twitter.com/0xtim/status/1460894705294745607?s=21
    @available(*, deprecated)
#endif
```

Just remove that section to test these revisions with 5.5.3.

finestructure:
This brings incremental builds back from ~34s to ~6s:

```
❯ time swift test --disable-automatic-resolution --filter SearchTests.test_DBRecord_packageURL
[24/24] Build complete!
Test Suite 'Selected tests' started at 2022-03-02 12:05:00.924
Test Suite 'SPI-ServerPackageTests.xctest' started at 2022-03-02 12:05:00.924
Test Suite 'SearchTests' started at 2022-03-02 12:05:00.924
Test Case '-[AppTests.SearchTests test_DBRecord_packageURL]' started.
Test Case '-[AppTests.SearchTests test_DBRecord_packageURL]' passed (0.103 seconds).
Test Suite 'SearchTests' passed at 2022-03-02 12:05:01.027.
	 Executed 1 test, with 0 failures (0 unexpected) in 0.103 (0.103) seconds
Test Suite 'SPI-ServerPackageTests.xctest' passed at 2022-03-02 12:05:01.027.
	 Executed 1 test, with 0 failures (0 unexpected) in 0.103 (0.103) seconds
Test Suite 'Selected tests' passed at 2022-03-02 12:05:01.027.
	 Executed 1 test, with 0 failures (0 unexpected) in 0.103 (0.104) seconds

________________________________________________________
Executed in    5.54 secs    fish           external
   usr time    8.53 secs    0.07 millis    8.53 secs
   sys time    2.61 secs    1.51 millis    2.61 secs
```

Fixes #1602

### DIFF
--- a/Sources/App/Core/Search.swift
+++ b/Sources/App/Core/Search.swift
@@ -19,7 +19,7 @@ import Vapor
 
 enum Search {
     static let schema = "search"
-    
+
     // identifiers
     static let author = SQLIdentifier("author")
     static let keyword = SQLIdentifier("keyword")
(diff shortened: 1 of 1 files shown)


daveverwer (approved):
Great sleuthing, @finestructure. The solution is 🤮 but as long as it speeds everything back up then that's good.

Do you think the compiler team would be interested in this? It's a reproducible (since it's open-source) erxample of a non-trivial performance issue.

Issue #1561, opened 2022-02-15
Make the build matrix link to the full compatibility report.


daveverwer:
This might also be an opportunity to switch it to a css-grid layout.

PR #1605, opened 2022-03-03
Matrix HTML and CSS refactor

daveverwer:
Switched from Flexbox to Grid for the overall layout, which means much less hackery!

In preparation for #1561


Sources/App/Views/PackageController/PackageShow+Types.swift
```diff
@@ -152,16 +152,11 @@ extension PackageShow.Model {
             self.results = namedResult.results
         }
         
-        var label: Node<HTML.BodyContext> {
+        var labelParagraphNode: Node<HTML.BodyContext> {
             guard !references.isEmpty else { return .empty }
-            return .div(
-                .class("row_label"),
-                .div( // Note: It may look like there is a completely useless div here, but it's needed. I promise.
-                    .div(
```
daveverwer, line 160: Removing two empty "unused" divs that were there only to allow the CSS to manipulate the layout is also good!

finestructure (approved):

PR #1606, opened 2022-03-04
Redirect package page 404s to Github

finestructure:
We've briefly discussed this in the context of the [ReleaseNotes](https://github.com/SwiftPackageIndex/ReleaseNotes) tool: what if we re-directed non-existing packages to Github URLs.

This is just throwing this together quickly to try and discuss.

The reason I'm picking this up again is in the context of package dependencies:

## Package URL resolution

One issue we'll need to address is how to resolve URLs we find in `Package.resolved` and the "product dependency tree". We _could_ look up each one against the index and replace it with the package id if we have a match and keep the URL only if we don't.

This is going to cause quite a number of lookups for some of the packages. (To do: Double check the actual impact.)

That raises the question how we're actually going to use these links to packages. Presumably, we'll want to allow links to package pages. So how about we simply either convert or re-interpret `https://github.com/foo/bar.git` as `https://swiftpackageindex.com/foo/bar` and add a redirect handler to the original URL if we don't have the package?

If we do this we don't really need to look up any dependency URLs at all.

There's one obvious downside with this: it assumes that any URL we don't host (at the package level) is a github.com and not any other source code hosting provider. (Unless of course we try them subsequently in priority order...)

However, if all we ever do is link to package pages from dependency URLs then substituting them with package IDs is actually detrimental. We'd only be doing a lookup for a package ID and then look it up _again_ to convert it to a URL when following a dependency.

Unless of course we don't convert them to package IDs but URLs. But then why do a look up in the first place? If all we want is the URL, we might as well optimistically guess it.

---

In the context of the ReleasNotes tool we could leverage the same mechanism: substitute `SwiftPackageIndex.com` for all `GitHub.com` url without checking the server first. While our release notes aren't as fully functional as Github's yet, they're also not a mile away either. And you can click through to Github's release note from each of them.

The biggest problem with substituting SPI urls was the handling of 404s, which this would solve.


Sources/App/Controllers/PackageController+routes.swift
```diff
@@ -18,20 +18,33 @@ import Vapor
...
+            return try await ShowRoute
+                .query(on: req.db, owner: owner, repository: repository)
+                .map {
+                    PackageShow.View(path: req.url.path,
+                                     model: $0.model, packageSchema: $0.schema)
+                        .document()
+                }.get()
+        } catch let error as AbortError where error.status == .notFound {
+            throw Abort.redirect(to: "https://github.com/\(owner)/\(repository)",
+                                 type: .permanent)
```
finestructure, line 46: Rather than blindly re-direct to Github here we could (should) check if it exists first.

Sources/App/Core/ErrorMiddleware.swift
```diff
@@ -24,9 +24,8 @@ public final class ErrorMiddleware: AsyncMiddleware {
     public func respond(to req: Request, chainingTo next: AsyncResponder) async throws -> Response {
         do {
             return try await next.respond(to: req)
-        } catch {
-            let abortError = error as? AbortError ?? Abort(.internalServerError)
-            let statusCode = abortError.status.code
+        } catch let error as AbortError where error.status.code >= 400 {
```
finestructure, line 27: This part we'll definitely want to add: right now we're showing our _error page_ when we're doing a re-direct.

I haven't found any redirects in our current Swift code, so that shouldn't be an actual issue yet. I'm not sure if JS redirects would be affected by this but I've also not ever seen our error page with a 301 so if it is an actual issue it's quite rare.

Sources/App/Controllers/PackageController+routes.swift
daveverwer: I'll review this properly later, but while I notice this specifically, we want to be very careful with permanent redirects. this tells Google that this URL is not and never will be valid. This should be a temporary redierct.

daveverwer: But, I'm not sure a server-side redirect is the best choice here as this might be a good opportunity to ask people to add the package, accompanied with a link to the GitHub page.

We need to think about this a bit before merging this one. I'm writing today, but I'll have a think about it over the weekend.

finestructure: Yeah, this wasn't for merging, it was for discussion!

daveverwer:
> That raises the question how we're actually going to use these links to packages. Presumably, we'll want to allow links to package pages. So how about we simply either convert or re-interpret `https://github.com/foo/bar.git` as `https://swiftpackageindex.com/foo/bar` and add a redirect handler to the original URL if we don't have the package?

At least initially, yes, the only purpose of resolving the dependencies to package index packages would be so we can link to SPI package pages for each dependency that we have in our index.

Dependent packages that we don’t yet (or can’t) have in our index could either be linked back to the original repository or not linked at all. It should only be a day or two for us to ingest any dependent packages, so this is not a huge deal.

> If we do this we don't really need to look up any dependency URLs at all.
> 
> There's one obvious downside with this: it assumes that any URL we don't host (at the package level) is a github.com and not any other source code hosting provider. (Unless of course we try them subsequently in priority order...)

We are very dependent on GitHub in a number of ways (some [captured here](https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/discussions/1343), others captured in source comments, and many not written down at all 😅). Of course, we should try and keep that number of ways to a minimum, but we already have some serious dependencies on GitHub in things like relying on GitHub to sanitise README files, etc… 

Part of me thinks that we could come up with a mechanism for a fairly generic matching algorithm based on the owner and repo name parsing that we already do. Assuming all providers roughly match the scheme of `domain/owner/repo` which is certainly true for GitLab and BitBucket which would be the first (and maybe the only? Possibly custom hosted git might be something people might want?) two alternative platforms we’d want to support.

We already parse out owner and repo and for now we’d just assume everything is GitHub. If/when we expand to other systems, this doesn’t feel like a huge problem.

> 
> However, if all we ever do is link to package pages from dependency URLs then substituting them with package IDs is actually detrimental. We'd only be doing a lookup for a package ID and then look it up _again_ to convert it to a URL when following a dependency.

Agreed, but if we did that lookup we could also grab other metadata along with the URL. If you see the mock-up I did for the dependency page it included things like the package summary and stars, etc… We could compromise on that, of course, but there is some value to that extra lookup/fetch.

> In the context of the ReleasNotes tool we could leverage the same mechanism: substitute `SwiftPackageIndex.com` for all `GitHub.com` url without checking the server first. While our release notes aren't as fully functional as Github's yet, they're also not a mile away either. And you can click through to Github's release note from each of them.
> 
> The biggest problem with substituting SPI urls was the handling of 404s, which this would solve.

Yes, and this is interesting, but I think we should use these 404s (which should be rare) to try and find those packages that we don’t yet have in the index.

Rather than a server-side redirect, we could show a page a little like this:

My Sketches - 2022-03-07 13 31 28

I know my handwriting is the best, but in case anyone needs a translation the title would be something along the lines of “owner/repo not found” then with a sentence below explaining that it’s not in the index, then two buttons: “Add this Package” and “View on GitHub”.

Yes, it’s one more click over and above the automatic redirect, but it should be rare and would help expand our coverage by encouraging people to add packages in the cases where it is shown.

finestructure:
Yeah, although I’d say we push this to the end of the queue. The main reason I brought this up was to inform the decision on how to deal with the data processing from the dependency data sources. Now that that’s sorted I was going to park this unless there’s some interim change we’d want to merge.

Like for instance we could initially just redirect and then insert the landing page later.

daveverwer:
I'll take this over and finish it then as I'd like us to take advantage of this change in the releases tool. I can do the work on it though.

finestructure:
Added a page stub:

CleanShot 2022-03-08 at 16 11 11@2x

finestructure:
Just two minor nits regarding the page content:

- I think it'd be nice to show the actual url we'd be redirecting to, or at least the owner/repo slug
- maybe add the GH icon to the blue button?

Sources/App/Core/ExternalURL.swift
```diff
@@ -19,13 +19,15 @@ enum ExternalURL: URLRepresentable {
...
     case raiseNewIssue
+    case raiseNewPackageListIssue
 
     var description: String {
         switch(self) {
             case .projectBlog: return "https://blog.swiftpackageindex.com"
             case .projectGitHub: return "https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server"
             case .projectSponsorship: return "https://github.com/sponsors/SwiftPackageIndex"
             case .raiseNewIssue: return "https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/issues/new/choose"
+            case .raiseNewPackageListIssue: return "https://github.com/SwiftPackageIndex/PackageList/issues/new/choose"
```
finestructure, line 30: Should we perhaps link straight to "[add package](https://github.com/SwiftPackageIndex/PackageList/issues/new?assignees=&labels=Add+Package&template=add_package.yml&title=Add+%3CPackage%3E)" ?

```swift
case .addNewPackage: return "https://github.com/SwiftPackageIndex/PackageList/issues/new?assignees=&labels=Add+Package&template=add_package.yml&title=Add+%3CPackage%3E"
```

finestructure: We might even be able to pre-populate the url field via a query parameter. That'd be a _really_ convenient way to get new packages added.

daveverwer: Done

daveverwer:
Over to you @finestructure for a review and pondering on that last test issue we discussed.

daveverwer:
Looks like there is an issue compiling that foundation networking on Linux, too 😬

finestructure:
> Looks like there is an issue compiling that foundation networking on Linux, too 😬

Looks like FoundationNetworking's URLSession doesn't support a/a yet :/

finestructure:
Getting consistent test failures in CI but they pass locally with the same base image (although it's the arm64 variant, so not _exactly_ the same).

finestructure:
Marking as draft as I've got some additional changes to merge in first.

finestructure:
I still struggle to reproduce the CI failures 1:1 locally in a dev container. Trying to debug this has spawned #1630.

I've started re-doing the changes on a new branch `redirect` in order to bisect where the breakage starts while leaving this branch here as is.

As it stands, I've got a "known good" (in the sense that [5 CI runs all succeeded](https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/actions/runs/2012033964)) at revision 2ba7d6157a0faaa450029a16d56b48045ad2e4f4 

The next commit aea39074b87ea411bc0c59f9cba2982583933af9 after that has errors (although I yet have to double check what exactly they are, xcbeautify is obscuring it).

What's baffling is that this here is the extent of the changes in that commit:

```swift
    func test_show_checkingGitHubRepository_notFound() throws {
        Current.fetchHTTPStatusCode = { _ in .notFound }

        // MUT
        try app.test(.GET, "/unknown/package", afterResponse: { response in
            XCTAssertEqual(response.status, .notFound)
        })
    }

    func test_show_checkingGitHubRepository_found() throws {
        Current.fetchHTTPStatusCode = { _ in .ok }

        // MUT
        try app.test(.GET, "/unknown/package", afterResponse: { response in
            XCTAssertEqual(response.status, .notFound)
        })
    }
```

The only thing slightly "unusual" is that we're replacing an `async throws` in `Current.fetchHTTPStatusCode`. But that's not unique. We're doing the same for `fetchMetadata`, `fetchLicense`, and `fetchReadme` already.

finestructure:
> although I yet have to double check what exactly they are, xcbeautify is obscuring it

Actually, I'd done that already: https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/runs/5612358152?check_suite_focus=true

The failure is

```
Assertion failed: PostgresConnection deinitialized before being closed
```

daveverwer:
> > although I yet have to double check what exactly they are, xcbeautify is obscuring it
> 
> Actually, I'd done that already: https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/runs/5612358152?check_suite_focus=true
> 
> The failure is
> 
> ```
> Assertion failed: PostgresConnection deinitialized before being closed
> ```

So again, this is likely not related to the specific changes in that commit (as you imply) but to a database connection problem during testing, as we've been seeing with the benchmarking tests?

finestructure:
> > > although I yet have to double check what exactly they are, xcbeautify is obscuring it
> > 
> > 
> > Actually, I'd done that already: https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/runs/5612358152?check_suite_focus=true
> > The failure is
> > ```
> > Assertion failed: PostgresConnection deinitialized before being closed
> > ```
> 
> So again, this is likely not related to the specific changes in that commit (as you imply) but to a database connection problem during testing, as we've been seeing with the benchmarking tests?

Wel... I can't see how it could be really but best I can tell it _is_ this commit. I've reverted it again and now the tests pass in CI https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/actions/runs/2014179786

I'm going to try and pick this apart further, I can't think of anything else I can do.

finestructure:
Sheer test count doesn't seem to be the issue: https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/actions/runs/2015768647

finestructure:
Ok, I've actually managed to reproduce the error `Assertion failed: PostgresConnection deinitialized before being closed` on macOS now. Still not sure _why_ it's happening but at least the turn-around via CI should be eliminated.

finestructure:
Well, once and never again...

finestructure:
Ok, so there seems to be a race condition at work which leads to premature shutdown (`Application.shutdown() was not called before Application deinitialized.`) when we mock out `Current.fetchHTTPStatusCode` with a closure that _runs too fast_.

The tests can be made to pass reliably in CI if we mock as follows:

```swift
Current.fetchHTTPStatusCode = { _, _ in
    usleep(100_000) // 100ms
    return .notFound
}
```

10ms is not enough, and so I've timed this to ~how long the HTTP request (which made this work earlier) took in CI.

I'll finish this PR up now and then look into how we can get this properly fixed.

finestructure:
It's also working with URLSession, so going back to that for now.

daveverwer:
I think this takes the biscuit as the hardest pull request that you never intended to merge 😅

It's good, though. I just gave it a final test on production and it works really well. It's definitely the easiest way to add a package, which is kinda messed up as it's another whole wing on that Rube Goldberg machine! 😬

We should update the Package update tool to link to SPI pages now.

PR #1610, opened 2022-03-07
Make the compatibility matrix sections link to the build report.

### LINKED ISSUE #1561: Make the build matrix link to the full compatibility report.

daveverwer:
Fixes #1561 

* Moved link to full build report from (1) to (2)
* Made both of the matrix sections (platform and Swift version) links to the same report.

It looks like there are huge changes to the snapshots, but I double checked and the only visual changes apart from the moving of the link. It just makes a huge difference because the whole page shifts upwards.

Screenshot 2022-03-07 at 16 42 22@2x


finestructure:
I'm getting a bit of a weird wrapping behaviour:

CleanShot 2022-03-08 at 05 17 29@2x

I've run `make build-front-end`. It's resize-independent, i.e. it stays like this at any window width.

It's because I'm zooming out one step to get smaller fonts.

Also, should we maybe just call the link "All Builds"? Given the placement, that's really enough, isn't it?

finestructure (approved):

daveverwer:
> I'm getting a bit of a weird wrapping behaviour
> It's because I'm zooming out one step to get smaller fonts.

Yea it was only on smaller zooms and *only* in Safari. All the bugs are Safari-only these days 😢 I played with it and found a workaround, but then changed the copy instead which had the same effect.

> Also, should we maybe just call the link "All Builds"? Given the placement, that's really enough, isn't it?

All Builds felt a little too short, but I changed it to "Full Build Results" which is slightly smaller and also fixes the Safari layout issue.

Issue #1560, opened 2022-02-15
Make the link to the original package repository more obvious


daveverwer:
Now that the link to the package which was under the package title is gone:

- [x] Add an icon and maybe some custom styling to the View on GitHub link and move it to the top of the list.
- [x] Maybe play with bringing versions up above some list items.
- [x] Possibly add an icon for "Try in a Playground"

<img width="1366" alt="My Sketches - 2022-02-15 19 24 54" src="https://user-images.githubusercontent.com/5180/154134471-9b7c8410-6ad2-404e-8c1f-fab47c3954bc.png">

PR #1612, opened 2022-03-07
Moved link to "More packages from this author" under the package title

### LINKED ISSUE #1562: Add the repo owner and repo name under the package name
I find myself wanting to see the repo and owner name again. It shouldn't be the full link like it was before, just the repo and owner.

daveverwer:
Merge after #1611 
Fixes #1562

Moved the "more packages" link from (1) to (2)

Screenshot 2022-03-07 at 17 47 33@2x

Yes, this isn't the link to the repo and doesn't include the repo name 😅 but I still think it's going to be better overall #1560 is complete. I don't know how you've got on with it @finestructure, but in the time since we removed the repo link from under the package name I got used to using the "View on GitHub" link quite quickly, so I think this is probably a step forward.

I also think this is a much more logical place for a "more packages from this author" link. I remember wanting to put it here originally but I was unable to since the repo link was there.


finestructure:
Mmm, couldn't we just have a link

Quick / Nimble

below the package name? I feel it's too tempting to click on that and then end up on the package owner's page on Github which is probably not what people want in the vast majority of the cases. If we wanted to preserve that, we could make it two links

[Quick] / [Nimble]

I've adjust to clicking in the sidebar to get to the Github page but what I really miss is seeing the "owner / repo" spelled out. There are often packages with rather generic names and seeing the "url slug" in the familiar format is really helpful.

pointfreeco/swift-parsing

I know it's both there with this change but there's a lot of recognition in it being spelled out like you're used to.

Also, there are cases where package name and repo name are different. And because we're substituting in author names, we're "breaking" recognition. For example:



I think this would be more helpful/recognisable as

# Navigator

[flopshot/Navigator](https://github.com/flopshot/Navigator)

daveverwer:
I was so sure you'd like this! 😅

On the author name/author username issue, I feel quite strongly that we should use the owner's readable name here. It matches with the Package name (we don't use repository name, and that feels like the correct decision).

If you want the owner/repo displayed somewhere that's fine though, let me see where I can squeeze it in.

finestructure:
Complete brainfart on my end. I thought this was about the old package url link that we recently dropped 🤦 

Completely agree on moving the author link over to the header! 👌

finestructure (approved):

PR #1621, opened 2022-03-16
Fix Plot warnings

finestructure:


daveverwer (approved):

finestructure:
John came up with a really clever workaround to fix that warning without source breakage 👌

PR #1638, opened 2022-03-23
Async set up

finestructure:
Bit of a/a work and a small fix that fell out of that PR of desperation

### DIFF
--- a/Tests/AppTests/Util.swift
+++ b/Tests/AppTests/Util.swift
@@ -36,10 +36,10 @@ func setup(_ environment: Environment, resetDb: Bool = true) async throws -> App
 
     if !_schemaCreated {
         // ensure we create the schema when running the first test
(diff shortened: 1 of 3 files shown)


Tests/AppTests/Util.swift
```diff
@@ -24,8 +24,8 @@ import XCTest
 
 private var _schemaCreated = false
 
-func setup(_ environment: Environment, resetDb: Bool = true) throws -> Application {
-    let app = Application(.testing)
+func setup(_ environment: Environment, resetDb: Bool = true) async throws -> Application {
+    let app = Application(environment)
```
finestructure, line 28: This was a potential bug but we only ever call it with `.testing` as a parameter.

finestructure:
I converted `setUp` and `tearDown` to a/a because at some point I was hoping it might help with the shutdown issues.

PR #1642, opened 2022-03-24
Adds a site breadcrumb bar

### LINKED ISSUE #1613: Site-wide navigation and breadcrumbs
This came up during the discussions around the new sidebar. We need a better structure for site-wide navigation. We currently have too many situations like this:

Screenshot 2022-03-08 at 17 15 30@2x

Where we have the "back" navigation built into the page text rather than being navigation. Or this:

Screenshot 2022-03-08 at 17 16 10@2x

Where we have dedicated navigation, but it's bespoke for each page.

We need some kind of dedicated navigation with a breadcrumb trail, and for a package page it should look like this:

My Sketches - 2022-03-08 15 45 35 2

The breadcrumbs would be across the top of the screen under the site header and would be the Organisation/User display name followed by the package name. Then, under the package title, we can bring back the owner's username and repo name.

Then, for sub-pages of Package pages, you'd see something like this:

My Sketches - 2022-03-08 15 45 35 3

Each section of the breadcrumb apart from the last would be a link back up the tree, or to package/author page etc...

daveverwer:
Fixes #1613

Breadcrumbs are now enabled site-wide, except for the home page (intentional). The last breadcrumb in a set will not be a link, partly for visual differentiation and partly because we can't easily get the link for the Markdown pages! 😬 

Screenshots:

Screenshot 2022-04-05 at 15 58 38@2x

Screenshot 2022-04-05 at 15 58 53@2x

Screenshot 2022-04-05 at 15 59 21@2x

Screenshot 2022-04-05 at 15 59 05@2x


finestructure:
Love it! ❤️

finestructure (approved):

PR #1646, opened 2022-03-26
Trigger 5.6 builds

finestructure:
This triggers Swift 5.6 builds and displays Swift 5.6 build results in the build matrix and on the build details page.

Deployment sequence:

- [x] install Xcode 13.3 on builders
- [x] ensure tags are set up correctly
- [x] merge [builder update](https://gitlab.com/finestructure/swiftpackageindex-builder/-/merge_requests/100)
- [ ] merge this PR


finestructure:
Technically we could deploy this before installing Xcode on the builders if we wanted to. 5.6 builds would be triggered and fail with an "infrastructure error". These are retried on our usual schedule.

daveverwer (approved):

PR #1664, opened 2022-04-01
Login in again in case the previous steps were slow

finestructure:
Should fix these build errors in Gitlab:

```
Done in 1.61s.
Step 1/15 : FROM registry.gitlab.com/finestructure/spi-base:0.7.0 as build
Head "https://registry.gitlab.com/v2/finestructure/spi-base/manifests/0.7.0": unauthorized: HTTP Basic: Access denied
```


finestructure:
I suspect we've been seeing these in the last couple of days because the 5.6 load is taxing our Linux builders (which also build our deploy images)

PR #1671, opened 2022-04-03
Enable GZIP compression for all responses

Sherlouk:
As per the documentation (https://docs.vapor.codes/4.0/server/#response-compression) this is turned off by default, but does provide (in testing) a decrease in response times and data sizes. Hopefully this will help reduce our bandwidth, reducing costs, as well as making the experience better for our users.

We will want to closely monitor these changes, it may be hard to evaluate perfectly in staging due to the low usage.

### DIFF
--- a/Sources/App/configure.swift
+++ b/Sources/App/configure.swift
@@ -21,6 +21,10 @@ public func configure(_ app: Application) throws -> String {
     app.logger.component = "server"
     Current.setLogger(app.logger)
 
+    // Enable GZIP (de-)compression on all requests and responses
+    app.http.server.configuration.responseCompression = .enabled
+    app.http.server.configuration.requestDecompression = .enabled
+    
     app.middleware.use(FileMiddleware(publicDirectory: app.directory.publicDirectory))
     app.middleware.use(ErrorMiddleware())
 


daveverwer:
OK so I deployed this to staging and I'm not sure it's working. Here's a network response from production (where this is not deployed):

Screenshot 2022-04-04 at 11 05 17@2x

and the same headers from staging where this branch is deployed:

Screenshot 2022-04-04 at 11 05 40@2x

As proof that this is generated by the right branch, check the commit that's deployed:

Screenshot 2022-04-04 at 11 06 56@2x

Which is the same commit as `main` right now:

Screenshot 2022-04-04 at 11 07 21@2x

daveverwer:
I didn't know about `br` content encoding before, but it does seem to be compressed from this article:

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Encoding#syntax

I won't look into this more for now, but I think we need to double check if this is needed and/or working.

daveverwer:
Figured it out. Response (and revert) incoming.

daveverwer:
So Cloudflare is handling this for us, and if we zip on our side we actually slow down the response as Cloudflare is then unzipping our response and then recompressing it using Brotli.

Looking at the response data from the raw IP on production. There is no compression:

Screenshot 2022-04-04 at 11 18 32@2x

and on staging, there is gzip compression (deflate):

Screenshot 2022-04-04 at 11 19 54@2x

But by the time they both come through Cloudflare, they're both `br` compressed.

PR #1675, opened 2022-04-04
Disable server-side compression and add an explanation of why

daveverwer:


Sherlouk (approved):
Told you there was a smart reason not to do this 😂

finestructure (approved):

PR #1684, opened 2022-04-08
Debug test race condition

finestructure:
We don't want to merge this after all as it would make our day to day test execution worse (after finding that macOS tests apparently need a delay as well, at least in some cases).

What we currently have on `main` seems to be working well enough and we need to get this fixed upstream (or be told what we're doing wrong with our setup).

I've started drafting an issue but it's not an easy one to pull together. I fear it's going to be hard to get eyes on this unless we manage to derive a less SPI specific reproducer.

I've isolated the repro into a single test file without any dependencies other than the `PackageController.ShowRoute` query. Unfortunately that's a big one. We'll have to see if we can reproduce this in a new Vapor project.

### DIFF
--- a/Sources/App/Core/ErrorMiddleware.swift
+++ b/Sources/App/Core/ErrorMiddleware.swift
@@ -29,8 +29,9 @@ public final class ErrorMiddleware: AsyncMiddleware {
             let isCritical = (statusCode >= 500)
 
             if isCritical {
+                let current = Current
                 Task.detached {
-                    try await Current.reportError(req.client, .critical, error).get()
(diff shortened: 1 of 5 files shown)


daveverwer (approved):

finestructure:
This has been superseded by #1704, which works around the race condition in our tests.

There's some useful code in this PR dealing with TSan issue that we may want to preserve. I'll close this PR and move the branch into the attic for now.

finestructure:
Actually, I think there's some valuable cleanup in here that we'll probably want to merge soon. I'll follow up with a new branch and PR at some point.

PR #1689, opened 2022-04-11
Fix flickering snapshot test

### LINKED ISSUE #1681: Flickering snapshot test
I first noticed this yesterday and thought this commit 483f888874a50e1dc2bdcdc5ed3c3d464c6ce42d might fix the issue but it appears this is not due to the number of tags.

It's the test `test_PackageShowView_few_keywords` that's sometimes failing. Out of 10 runs of that test by itself (by clicking on the test diamond in Xcode) it failed 4 times for me just now. It obviously won't happen in CI because we don't run the image snapshot tests there.

CleanShot 2022-04-07 at 10 09 43@2x

It must be something in the Javascript as the HTML snapshot is unaffected.

daveverwer:
Fixes #1681

Should be more reliable now! The detection of overflow was sometimes happening before the DOM had finished loading and painting.


finestructure (approved):

daveverwer:
Deploying as 2.63.1

PR #1699, opened 2022-04-17
[1640] Add "Published <date> - Last updated: <date>" to bottom of package page

### LINKED ISSUE #1640: SEO optimisation - Last Updated
We are specifying some dates to Google via the JSON-LD data, but it's still not showing a date against our package pages.

According to [this page](https://developers.google.com/search/docs/advanced/appearance/publication-dates) we should also specify `datePublished` in the JSON-LD, and also somewhere on the page put the words "Last updated:" with a fully specified date (not "x days ago") to get Google to report this data.

jeroenleenarts:
Fixes #1640

### DIFF
--- a/Sources/App/Views/PackageController/PackageShow+StructuredDataSchema.swift
+++ b/Sources/App/Views/PackageController/PackageShow+StructuredDataSchema.swift
@@ -20,7 +20,7 @@ extension PackageShow {
         enum CodingKeys: String, CodingKey {
             case context = "@context", type = "@type"
             case identifier, name, description, license, version,
-                 codeRepository, url, dateCreated, dateModified,
+                 codeRepository, url, datePublished, dateModified,
                  sourceOrganization, programmingLanguage, keywords
         }
         
@@ -34,8 +34,8 @@ extension PackageShow {
         let version: String?
         let codeRepository: String
         let url: String
-        let dateCreated: Date?
-        let dateModified: Date?
+        let datePublished: Date
+        let dateModified: Date
         let sourceOrganization: OrganisationSchema
         let programmingLanguage: ComputerLanguageSchema
         let keywords: [String]
@@ -48,8 +48,8 @@ extension PackageShow {
             licenseUrl: String?,
             version: String?,
             repositoryUrl: String,
-            dateCreated: Date?,
-            dateModified: Date?,
+            datePublished: Date,
+            dateModified: Date,
             keywords: [String]
         ) {
             self.identifier = "\(repositoryOwner)/\(repositoryName)"
@@ -59,7 +59,7 @@ extension PackageShow {
             self.version = version
             self.codeRepository = repositoryUrl
             self.url = SiteURL.package(.value(repositoryOwner), .value(repositoryName), .none).absoluteURL()
-            self.dateCreated = dateCreated
+            self.datePublished = datePublished
             self.dateModified = dateModified
             self.sourceOrganization = OrganisationSchema(legalName: organisationName ?? repositoryOwner)
             self.programmingLanguage = ComputerLanguageSchema(name: "Swift", url: "https://swift.org/")
@@ -71,7 +71,9 @@ extension PackageShow {
             let repository = result.repository
             guard
                 let repositoryOwner = repository.owner,
-                let repositoryName = repository.name
+                let repositoryName = repository.name,
+                let datePublished = repository.firstCommitDate,
+                let dateModified = repository.lastCommitDate
             else {
                 return nil
             }
@@ -88,8 +90,8 @@ extension PackageShow {
                     releaseVersion: result.releaseVersion,
                     preReleaseVersion: result.preReleaseVersion).stable?.link.label,
                 repositoryUrl: package.url.droppingGitExtension,
-                dateCreated: repository.firstCommitDate,
-                dateModified: repository.lastCommitDate,
+                datePublished: datePublished,
+                dateModified: dateModified,
                 keywords: repository.keywords
             )
         }
--- a/Sources/App/Views/PackageController/PackageShow+View.swift
+++ b/Sources/App/Views/PackageController/PackageShow+View.swift
@@ -271,7 +271,7 @@ enum PackageShow {
             .unwrap(packageSchema) {
                 .section(
                     .hr(),
-                    .p(.small(.text("Published \(Self.dateFormatter.string(from:$0.datePublished)) - Last updated: \(Self.dateFormatter.string(from:$0.dateModified))")))
+                    .p(.small(.text("First published: \(Self.dateFormatter.string(from:$0.datePublished)) - Last updated: \(Self.dateFormatter.string(from:$0.dateModified))")))
                 )
             }
             
--- a/Sources/App/Views/PackageController/PackageShow+View.swift
+++ b/Sources/App/Views/PackageController/PackageShow+View.swift
@@ -20,6 +20,13 @@ enum PackageShow {
     
     class View: PublicPage {
         
+        private static let dateFormatter: DateFormatter = {
+            let formatter = DateFormatter()
+            formatter.dateStyle = .medium
+            
+            return formatter
+        }()
+        
         let model: Model
         let packageSchema: PackageSchema?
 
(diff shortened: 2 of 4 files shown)


Sources/App/Views/PackageController/PackageShow+View.swift
```diff
@@ -256,6 +268,16 @@ enum PackageShow {
                 )
             )
         }
+        func sidebarPublishedAndUpdatedSection() -> Node<HTML.BodyContext> {
+            .unwrap(packageSchema) {
+                .section(
+                    .class("sidebar_package_published"),
+                    .div(.small(.text("Published \(Self.dateFormatter.string(from:$0.dateCreated!))"))),
```
jeroenleenarts, line 275: Force unwrap is no good.

Sources/App/Views/PackageController/PackageShow+View.swift
```diff
@@ -256,6 +268,16 @@ enum PackageShow {
...
+                    .div(.small(.text("Last updated: \(Self.dateFormatter.string(from:$0.dateModified!))")))
```
jeroenleenarts, line 276: Force unwrap

Sources/App/Views/PackageController/PackageShow+StructuredDataSchema.swift
```diff
@@ -20,7 +20,7 @@ extension PackageShow {
         enum CodingKeys: String, CodingKey {
             case context = "@context", type = "@type"
             case identifier, name, description, license, version,
-                 codeRepository, url, dateCreated, dateModified,
+                 codeRepository, url, dateCreated = "datePublished", dateModified,
```
finestructure, line 23: I think we should update the underlying attribute to `datePublished` instead of rewriting the key here. We're already mapping once from `firstCommitDate` → `dateCreated` and then encoding that as `datePublished` will get confusing in the future :)

Sources/App/Views/PackageController/PackageShow+View.swift
finestructure:
```swift
                    .unwrap($0.datePublished, {
                        .div(.small(.text("Published \(Self.dateFormatter.string(from: $0))")))
                    }),
                    .unwrap($0.dateModified, {
                        .div(.small(.text("Last updated: \(Self.dateFormatter.string(from: $0))")))
                    })
```

should work but I think we can actually make those dates non-optional in the model and bail in `init?` should they be nil (which is practise they wouldn't):

```swift
        init?(result: PackageController.PackageResult) {
            let package = result.package
            let repository = result.repository
            guard
                let repositoryOwner = repository.owner,
                let repositoryName = repository.name,
                let datePublished = repository.firstCommitDate,
                let dateModified = repository.lastCommitDate,
                let dateLastActive = repository.lastActivityAt
            else {
                return nil
            }
            
            self.init(
                repositoryOwner: repositoryOwner,
                repositoryName: repositoryName,
                organisationName: repository.ownerName,
                summary: repository.summary,
                licenseUrl: repository.licenseUrl,
                version: PackageShow.releaseInfo(
                    packageUrl: package.url,
                    defaultBranchVersion: result.defaultBranchVersion,
                    releaseVersion: result.releaseVersion,
                    preReleaseVersion: result.preReleaseVersion).stable?.link.label,
                repositoryUrl: package.url.droppingGitExtension,
                datePublished: datePublished,
                dateModified: dateModified,
                dateLastActive: dateLastActive,
                keywords: repository.keywords
            )
        }
```

finestructure: (Throwing `dateLastActive` in there as well for symmetry.)

finestructure: Actually, on that note - `dateLastActive` doesn't seem to be used anywhere. My understanding of #1640 was that this should replace `dateModified`?

So I think we should drop that property and instead assign `lastActivityAt` to `dateModified` instead of using `lastCommitDate`.

But I'll let @daveverwer confirm that - I just wanted to give it a pass because I know Dave isn't around today.

finestructure (changes requested):
Thanks a lot Jeroen! Looking good, just added a few bits and bobs inline.

Sources/App/Views/PackageController/PackageShow+View.swift
finestructure: I've gone ahead and tried the `init?` change that I proposed. It can be made to work but it means we have to update some unrelated tests, because we're now more strict about populating `firstCommitDate` and `lastCommitDate` in our mocks.

I'll add a PR against this PR for discussion.

Sources/App/Views/PackageController/PackageShow+View.swift
```diff
@@ -140,7 +148,11 @@ enum PackageShow {
                     .hr(
                         .class("minor")
                     ),
-                    sidebarInfoForPackageAuthors()
+                    sidebarInfoForPackageAuthors(),
+                    .hr(
+                        .class("minor")
+                    ),
+                    sidebarPublishedAndUpdatedSection()
```
daveverwer, line 155: I like the experimentation with giving a couple of spots for this to be positioned, but since this information is effectively just for Google I think we should put it under the readme and remove this one.

Of course, once it's no longer in the sidebar it'll need a more appropriate name, too. We could even call it something like `visibleMetadataSection`so it's obvious what the purpose of it is so it stops us from adding more to it in the future (unless we need more visible metadata later). We should also probably add a comment to it to explain its purpose, maybe even with a link to the Google documentation?

Sources/App/Views/PackageController/PackageShow+View.swift
daveverwer: Yea I think it's extremely unlikely that either of these fields will ever be nil but if they are it's fine to omit it with the `unwrap`.

daveverwer: In terms of the markup here, the `section` tag is a good idea, but let's replace the `div` tag with a `p` so it gives a little padding above the footer. I'd also keep both fields on one line, maybe separated with a dash, and keep the formatting with the colon consistent. I'd also probably put a `hr` tag in there to separate it from the bottom of the README content.

Something like this:

Screenshot 2022-04-19 at 08 41 31@2x

Screenshot 2022-04-19 at 08 42 29@2x

I think we'll also get away without any section specific CSS, so the class can be omitted too.

Sources/App/Views/PackageController/PackageShow+View.swift
jeroenleenarts: Should this content be within the `article` tag or not. Also Google indicates we should…

> feature it prominently.

What does that even mean. 🤣

daveverwer: Yea Google is always vague to attempt to stop gaming their algorithm. I think it'll be fine at the bottom but we'll know for sure because this change is going to make a visible change to search results, like this:

Screenshot 2022-04-19 at 12 47 59@2x

So if that doesn't appear after a week or two of this being deployed, we'll need to move it.

Sources/App/Views/PackageController/PackageShow+View.swift
```diff
@@ -256,6 +267,15 @@ enum PackageShow {
                 )
             )
         }
+        func visibleMetadataSection() -> Node<HTML.BodyContext> {
+            .unwrap(packageSchema) {
+                .section(
+                    .hr(),
+                    .p(.small(.text("Published \(Self.dateFormatter.string(from:$0.datePublished)) - Last updated: \(Self.dateFormatter.string(from:$0.dateModified))")))
```
daveverwer, line 274:
```suggestion
                    .p(.small(.text("First published: \(Self.dateFormatter.string(from:$0.datePublished)) - Last updated: \(Self.dateFormatter.string(from:$0.dateModified))")))
```

daveverwer (approved):

finestructure:
Opened follow-up #1705 to address test failures and `lastActivityAt` usage.

Thanks again Jeoren!

finestructure:
Merged via #1705 but since we rebased a couple of times this isn't being picked up automatically!

PR #1702, opened 2022-04-19
Move image-based snapshot testing to Percy

### LINKED ISSUE #1695: Migrate image snapshot tests to Percy
I didn’t re-open #1117 or add this to #1355 as I wasn’t sure this was going to happen, but I’ve been working on setting this up for a little while and thanks to the generosity of BrowserStack, we’re now able to use Percy for our snapshots. 🚀

I’ve pursued this because there are several advantages to having our snapshots outside the repository and generated using a service:

* Consistency - By having Percy take the snapshots rather than relying on one of our machines, we gain a consistent environment that’s not dependent on what version of macOS or Safari we’re running.
* Consistency - By taking snapshots via a CI process, we can never forget to build the front end files.
* Cross-browser testing - We upload a serialisation of the DOM to Percy (no images) and they reconstruct that inside several browsers to capture image snapshots. By adding Chrome to our snapshot testing we can capture any breaking changes for 95% of our visitors (70% Safari and 25% Chrome).
* Easier, more flexible image assets - Currently we have to run all our images through a manual process of SVG Export -> Convert to Base64 -> Add to [images.scss](https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/blob/main/FrontEnd/styles/images.scss) so that they render in our snapshots. Using Percy means we’d be able to include regular images. 😅
* Easier diffing - We get to more easily see diffs during review. This process will give us Kaleidoscope-style visual diffs alongside our PRs. 
* Repository size - Not a reason to do this by itself, but our checked out repository is north of 2Gb now and I’m fairly certain that the majority of that is in snapshots.

We would keep `swift-snapshot-testing`, but it would be responsible for taking the HTML snapshots which we would still commit to the repository. Those snapshots then form the basis of our Percy testing.

The process of taking the snapshots in CI will be:

* Check out the repository
* Diff the branch with `main` to determine whether any of the HTML snapshots have changed, or if the CSS/JavaScript files have changed. Exit on finding no changes.
* Run the front-end build process.
* Run Percy against the contents of the `WebpageSnapshotTests` directory.

daveverwer:
Fixes #1695


.github/workflows/snapshots.yml
```diff
@@ -0,0 +1,45 @@
...
+        run: make build-front-end
+      - name: Copy Front-End
+        run: cp -a ./Public/* ./Tests/AppTests/__Snapshots__/WebpageSnapshotTests
+      - name: Exclude NoImageSnapshot Tests
+        run: rm ./Tests/AppTests/__Snapshots__/WebpageSnapshotTests/*_noImageSnapshots*.html
+      - id: diff-includes-html-files
+        name: Diff for HTML Changes
+        run: |
+          filecount=$(git diff origin/main --shortstat **.html | awk '{print $1}')
+          echo "::set-output name=matching-files::${filecount:-0}"
```
finestructure, line 40: As discussed, here's my take if you like it better:

```
bash-3.2$ git diff --name-only origin/main | grep .html$ | wc -l | bc
29
bash-3.2$ filecount=$(git diff origin/main --shortstat **.html | awk '{print $1}')
bash-3.2$ echo ${filecount:-0}
29
```

The `${...:-0}` is to convert the empty string when there's not diff to `0`.

finestructure: For reference:

```
bash-3.2$ git diff origin/main --shortstat **.html
 29 files changed, 7361 insertions(+)
```

daveverwer: Much better without the external file, too! 👍

.github/workflows/snapshots.yml
```diff
@@ -0,0 +1,43 @@
...
+      - name: Copy Front-End
+        run: cp -a ./Public/* ./Tests/AppTests/__Snapshots__/WebpageSnapshotTests
+      - name: Diff for HTML Changes
+        id: front-end-changes
+        run: |
+          filecount=$(git diff origin/main --shortstat "**.html" "**.css" "**.js" | awk '{print $1}')
+          echo "::set-output name=matching-files::${filecount:-0}"
+      - name: Run Percy
+        if: steps.front-end-changes.outputs.matching-files != '0'
+        run: npx @percy/cli snapshot ./Tests/AppTests/__Snapshots__/WebpageSnapshotTests
```
Sherlouk, line 41: minor: I'd probably keep the verbose flag in for now while we're still learning and fixing issues. Don't think it actively hurts us to have on.

.percy.yml
```diff
@@ -0,0 +1,9 @@
+version: 2
+static:
+  include: '**/*.{html,htm}'
+  exclude: '**/*noImageSnapshots*.html'
+snapshot:
+  widths:
+    - 375
```
Sherlouk, line 7: question: Do we have stats on widths which are used? I know 414pt is typically more common nowadays than 375pt but the difference is relatively negligible and it's probably best to test the worst case.

Sherlouk (approved):
Looking good!

finestructure (approved):

.github/workflows/snapshots.yml
```diff
@@ -32,12 +32,12 @@ jobs:
       - name: Copy Front-End
         run: cp -a ./Public/* ./Tests/AppTests/__Snapshots__/WebpageSnapshotTests
       - name: Diff for HTML Changes
-        id: diff-includes-html-files
+        id: front-end-changes
         run: |
           filecount=$(git diff origin/main --shortstat "**.html" "**.css" "**.js" | awk '{print $1}')
```
Sherlouk, line 37: major: Does this need to be .scss? We don't commit the CSS files into the branch so I don't think `git diff` will ever spot those changes?

.github/workflows/snapshots.yml
daveverwer: Good idea. 👍

.percy.yml
daveverwer: The iPhone X, XS (etc...) and iPhone 13 mini are all still 375, and the regular iPhone 13 is only 390, so I think this value is a sensible one.

.github/workflows/snapshots.yml
daveverwer: Really well spotted! 😅

PR #1705, opened 2022-04-20
Add Published and Updated dates to package pages

finestructure:
Supersedes #1699

### DIFF
--- a/Sources/App/Views/PackageController/PackageShow+StructuredDataSchema.swift
+++ b/Sources/App/Views/PackageController/PackageShow+StructuredDataSchema.swift
@@ -20,7 +20,7 @@ extension PackageShow {
         enum CodingKeys: String, CodingKey {
             case context = "@context", type = "@type"
             case identifier, name, description, license, version,
-                 codeRepository, url, dateCreated, dateModified,
+                 codeRepository, url, datePublished, dateModified,
                  sourceOrganization, programmingLanguage, keywords
         }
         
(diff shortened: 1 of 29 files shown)


daveverwer (approved):

finestructure:
Looks like the snapshots are still failing for some reason 🤔

finestructure:
Passing locally with those snapshots, might be flickers

daveverwer:
Yea I've been looking into this and I can't figure it out. It's not the image snapshots as they don't run on CI. The tests definitely pass locally.

I wish we could see the diff from the CI on the server 😬

Sherlouk:
> I wish we could see the diff from the CI on the server 😬

Out of scope for this PR, but we could probably start uploading the log file from the test run?

It should include the full XCTAssert output which would help to narrow it down?

daveverwer:
It's very odd. I just re-generated snapshots locally with a trivial change to a non-visible element in the HTML and those same three tests still fail.

Maybe @Sherlouk's idea should be in-scope for this PR as I'm out of ideas on how else to see what's wrong.

daveverwer:
@finestructure do you still have a local Linux environment to run the tests set up? Could you run this in that environment?

finestructure:
Fun: Foundation date formatter difference between platforms...

CleanShot 2022-04-20 at 15 13 59@2x


```
  6> fmt.locale.identifier
$R2: String = "en_GB"
```

```
  6> fmt.locale.identifier
$R2: String = "en_US"
```

PR #1711, opened 2022-04-20
Remove the hardcoded height on a README

### LINKED ISSUE #1708: Fix very long package page screenshots in Percy
5e5025a7576d4596b931d6d2e6dec109e39950bfbc2fc299b07a8e93256afc88

daveverwer:
Fixes #1708

Previously, to ensure that a package page could be reloaded and not reset the scroll position (as the README is loaded dynamically), we padded the README container to 10,000px until the README loaded.

That broke the snapshot testing now that Percy is snapshotting a full-page, rather than what did previously with a maximum height.

This PR caches README heights per package at runtime to remove the problem. In order to do that, several of our custom JS tags needed converting to Stimulus controllers.

😅

### DIFF
--- a/Tests/AppTests/__Snapshots__/WebpageSnapshotTests/test_PackageShowView.1.html
+++ b/Tests/AppTests/__Snapshots__/WebpageSnapshotTests/test_PackageShowView.1.html
@@ -320,13 +320,15 @@
             </section>
           </section>
         </article>
-        <nav data-controller="tab-bar" data-action="popstate@window->tab-bar#setTabFromLocation">
-          <ul>
-            <li id="readme" class="active" data-tab-bar-target="tab" data-action="click->tab-bar#updateTab">README</li>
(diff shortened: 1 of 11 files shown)


Sherlouk:
Snapshots look much healthier! Have noticed (unsure if intentional) but we've lost the padding on the left of the tab items (README/Releases).

daveverwer:
> Snapshots look much healthier! Have noticed (unsure if intentional) but we've lost the padding on the left of the tab items (README/Releases).

Yea this is intentional. I wouldn't have made this change by itself, but I think it looks better now that a bigger change is in progress.

Sources/App/Views/Plot+Extensions.swift
```diff
@@ -228,3 +248,14 @@ extension Attribute where Context == HTML.InputContext {
         .data(named: "gramm", value: String(isEnabled))
     }
 }
+
+extension Attribute where Context == HTML.BodyContext {
+}
```
finestructure, line 253: Only saw this after merging - there's some "dead" code here.

PR #1718, opened 2022-04-28
Prevent empty strings from being shown as homepage URLs

### LINKED ISSUE #1716: Package homepage buttons are showing for packages without a homepage set
For example, on: https://swiftpackageindex.com/mxcl/PromiseKit

Screenshot 2022-04-26 at 16 39 03@2x

I think this is a `NULL`/`EMPTY` string issue. For example, looking at Max's packages:

Screenshot 2022-04-26 at 16 41 14@2x

PromiseKit and [AppUpdater](https://swiftpackageindex.com/mxcl/AppUpdater) have the button, but [StreamReader](https://swiftpackageindex.com/mxcl/StreamReader) doesn't.

Sherlouk:
Fixes #1716

### DIFF
--- a/Sources/App/Commands/Ingest.swift
+++ b/Sources/App/Commands/Ingest.swift
@@ -221,7 +221,7 @@ func insertOrUpdateRepository(on database: Database,
         .first() ?? Repository(packageId: pkgId)
     repo.defaultBranch = repository.defaultBranch
     repo.forks = repository.forkCount
-    repo.homepageUrl = repository.homepageUrl
+    repo.homepageUrl = repository.homepageUrl?.trimmed
     repo.isArchived = repository.isArchived
     repo.isInOrganization = repository.isInOrganization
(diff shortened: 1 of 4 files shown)


Sherlouk:
one mo

daveverwer:
Thanks James! Even testing our other untested functions! 🙇

Sherlouk:
they were plenty tested implicitly by other tests but might as well cover the 3 extensions in that grouping while I was there 😛

daveverwer:
I can't see it possibly being related to anything in this PR. That check also failed with #1717, I noticed. Something we need to look at.

finestructure:
The query performance tests is partly dependent on the db and its content. Costs can change depending on how effective indexes are or which choices the planner makes depending on table statistics.

I'll merge this now and add an issue to follow up on query performance.

finestructure:
#1721

PR #1723, opened 2022-05-03
Docc proxy

finestructure:
To do:

- [x] add tests for routes
- [x] check if `http://\(bucket)...` could/should be `https://\(bucket)...`
- [x] what kind of tests can/should we add for `DocumentationPageProcessor`? Anything unit testable? Snapshot testable?
- [x] properly handle aws bucket 4xxs

CleanShot 2022-05-03 at 12 23 32@2x


Sources/App/Controllers/PackageController+routes.swift
```diff
@@ -55,6 +55,115 @@ struct PackageController {
...
+        case data
+        case js
+    }
+
+    func awsURL(owner: String, repository: String, reference: String, fragment: Fragment, path: String) throws -> URI {
+        guard let bucket = Current.awsDocsBucket() else {
+            throw AppError.envVariableNotSet("AWS_DOCS_BUCKET")
+        }
+
+        return URI(string: "http://\(bucket).s3.us-east-2.amazonaws.com/\(owner)/\(repository)/\(reference)/\(fragment)/\(path)")
```
finestructure, line 70: We should probably make the AWS region configurable at some point but I didn't want to add another config variable right now.

Sources/App/Controllers/PackageController+routes.swift
```diff
@@ -55,6 +56,76 @@ struct PackageController {
...
+            throw Abort(.notFound)
+        }
+
+        let path = req.parameters.getCatchall().joined(separator: "/")
+
+        let url = try Self.awsDocumentationURL(owner: owner, repository: repository, reference: reference, fragment: fragment, path: path)
+        let res = try await Current.fetchDocumentation(req.client, url)
+        guard (200..<399).contains(res.status.code) else {
+            // Convert anything that isn't a 2xx or 3xx into a 404
+            throw Abort(.notFound)
```
finestructure, line 95: This is now showing our 404 instead of an AWS 404 (or other error page) with our docs header on top. That's better but eventually we'll probably want to redirect to a page that explains that there's a problem with doc generation like we do for build errors.

Tests/AppTests/PackageController+routesTests.swift
```diff
@@ -197,6 +197,180 @@ class PackageController_routesTests: AppTestCase {
...
+            XCTAssertEqual($0.status, .ok)
+            XCTAssertEqual($0.content.contentType?.description, "text/html; charset=utf-8")
+            XCTAssertTrue(
+                $0.body.asString().contains("<p>/owner/package/1.2.3/documentation/</p>"),
+                "was: \($0.body.asString())"
+            )
+            // Assert body includes the docc.css stylesheet link (as a test that our proxy header injection works)
+            XCTAssertTrue($0.body.asString()
+                    .contains(#"<link rel="stylesheet" href="/docc.css?test">"#),
+                          "was: \($0.body.asString())")
```
finestructure, line 248: I've added a very crude high-level document processing test here.

Sources/App/Views/ResourceReloadIdentifier.swift
```diff
@@ -0,0 +1,47 @@
...
+    // This method is only called in a local development environment, so all paths
+    // can be relative to this source file.
+    private static func modificationDate(forLocalResource resource: String) -> Date {
+        let relativePathToPublic = "../../../Public/"
+        let url = URL(fileURLWithPath: relativePathToPublic + resource,
+                      relativeTo: URL(fileURLWithPath: #file))
+
+        // Assume the file has been modified *now* if the file can't be found.
+        guard let attributes = try? Foundation.FileManager.default.attributesOfItem(atPath: url.path)
+        else { return Date() }
```
daveverwer, line 41: `Current.`

Sources/App/Views/ResourceReloadIdentifier.swift
```diff
@@ -0,0 +1,47 @@
...
+
+    // This method is only called in a local development environment, so all paths
+    // can be relative to this source file.
+    private static func modificationDate(forLocalResource resource: String) -> Date {
+        let relativePathToPublic = "../../../Public/"
+        let url = URL(fileURLWithPath: relativePathToPublic + resource,
+                      relativeTo: URL(fileURLWithPath: #file))
+
+        // Assume the file has been modified *now* if the file can't be found.
+        guard let attributes = try? Foundation.FileManager.default.attributesOfItem(atPath: url.path)
```
daveverwer, line 40: `Current.`

daveverwer (approved):

Sources/App/Views/ResourceReloadIdentifier.swift
```diff
@@ -0,0 +1,47 @@
...
+            return String(Int(latestModificationDate.timeIntervalSince1970))
+        }
+    }
+
+    // This method is only called in a local development environment, so all paths
+    // can be relative to this source file.
+    private static func modificationDate(forLocalResource resource: String) -> Date {
+        let relativePathToPublic = "../../../Public/"
+        let url = URL(fileURLWithPath: relativePathToPublic + resource,
+                      relativeTo: URL(fileURLWithPath: #file))
```
Sherlouk, line 37: Would it be possible/better to use `DirectoryConfiguration.detect().publicDirectory` to prevent it being tied to the file location?

Sources/App/Views/ResourceReloadIdentifier.swift
```diff
@@ -0,0 +1,47 @@
...
+
+        // Assume the file has been modified *now* if the file can't be found.
+        guard let attributes = try? Foundation.FileManager.default.attributesOfItem(atPath: url.path)
+        else { return Date() }
+
+        // Also assume the file is modified now if the attribute doesn't exist.
+        let modificationDate = attributes[FileAttributeKey.modificationDate] as? Date
+        return modificationDate ?? Date()
```
Sherlouk, line 45: `Current.`

daveverwer: Thanks James! No need to mark all of these, the two I marked were to remind me to do the rest.

Sources/App/Views/ResourceReloadIdentifier.swift
daveverwer: This worked out great. Thanks James!

daveverwer:
Slight bug with the Percy intregration here in that it's not picking up a *new* snapshot test. That'll need fixing before we merge this.

Sherlouk:
> Slight bug with the Percy integration here in that it's not picking up a _new_ snapshot test

Is this with our funky diff command not spotting the change?

daveverwer:
> I’m not sure what to make of the snapshot diff. It looks like it’s flagging the readme diff which we merged earlier when there shouldn’t be any diff. (I’ve just retriggered a new build by rebasing, so it’ll perhaps sort itself out.)

I didn't see it at the time but all I see now is what I expected to see, one new snapshot.

finestructure:
Must have been the rebase then 👍

PR #1724, opened 2022-05-03
Add `Analyze` namespace around analysis functions

finestructure:
It's been bothering me for quite a while that all our analysis (and reconciliation, ingestion, etc) code are free, top-level functions and types. (I've actually had a branch that "fixes" this for reconciliation and ingestion on the side.)

When I wanted to add a new type `PackageInfo` to `Analyze.swift` for the documentation targets just now it conflicted with the same type in `Views`.

This looks like a big diff but is structurally pretty simple. Everything moves inside `enum Analyze { ... }` and all `func`s become `static func` to preserve semantics.

### DIFF
--- a/Sources/App/Commands/Analyze.swift
+++ b/Sources/App/Commands/Analyze.swift
@@ -18,56 +18,60 @@ import Vapor
 import ShellOut
 
 
-struct AnalyzeCommand: Command {
-    let defaultLimit = 1
-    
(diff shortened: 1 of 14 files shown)


finestructure:
The diff is best viewed locally in Kaleidoscope, because it deals with the whitespace only changes much better.

daveverwer (approved):

PR #1726, opened 2022-05-04
Diff against main to determine if Percy needs to run

### LINKED ISSUE #1725: Snapshots workflow doesn't run when pushing multiple changes at once
From https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/pull/1723#issuecomment-1116717189

> Actually I wonder if it's because you pushed up multiple commits at once. I think our current command expects you push every commit up. At the moment it's comparing the last commit, to the commit before it, and there is no changes to HTML files in that last commit.

You're right, this is the problem. I'm not sure there's going to be an easy way to check how many commits there have been between the current push and the previous push so I think the best way will be to go back to our fallback plan of diffing against `main`. For me, pushing multiple commits up is something I do *all* the time.

If the diff between a branch and main contains any of the files or patterns we're checking for, Percy runs. It will mean that every push to a branch after a single HTML change will run Percy, but it will work in every case.

daveverwer:
Fixes #1725

### DIFF
--- a/.github/workflows/snapshots.yml
+++ b/.github/workflows/snapshots.yml
@@ -26,7 +26,7 @@ jobs:
       - name: Checkout code
         uses: actions/checkout@v3
         with:
-          fetch-depth: 2
+          fetch-depth: 0
       - name: Build Front-End
         run: make build-front-end
       - name: Copy Front-End
--- a/.github/workflows/snapshots.yml
+++ b/.github/workflows/snapshots.yml
@@ -34,7 +34,7 @@ jobs:
       - name: Diff for Front-End Changes
         id: front-end-changes
         run: |
-          filecount=$(git diff HEAD^ --shortstat "**.html" "**.scss" "**.js" ".percy.yml" ".github/workflows/snapshots.yml" | awk '{print $1}')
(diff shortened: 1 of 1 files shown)


finestructure (approved):

daveverwer:
Hmm. I tested this locally and it worked well, but the workflow apparently found changes when running in CI.

If I have this branch checked out and run:

```
git diff origin/main --shortstat "**.html" "**.scss" "**.js" ".percy.yml" ".github/workflows/snapshots.yml"
```

I get no output, which is correct.

daveverwer:
> I get no output, which is correct.

Wait. Ignore me. I got no output before I also changed the workflow file which is included in that command! All working as expected!

Sherlouk:
Isn't this what the script was initially, but run into issues when run on `main`?

daveverwer:
If I remember rightly, it wasn't due to a problem. It was more that we could save some Percy quota with the other version.

daveverwer:
and reduce the fetch depth, too.

daveverwer:
But if there's a better way to do it. I'm all ears!

I don't know how we'd check how many commits were just pushed, though.

Sherlouk:
Probably doesn't need to be a completely different workflow, just change that one script based on the current ref?

PR #1727, opened 2022-05-05
Front-end changes to the DocC proxy feature

daveverwer:


Tests/AppTests/Snapshotting+html.swift
```diff
@@ -17,6 +17,12 @@
 import Plot
 import SnapshotTesting
 
+extension Snapshotting where Value == String, Format == String {
+    public static var html: Snapshotting {
+        Snapshotting<String, String>.init(pathExtension: "html", diffing: .lines) { $0 }
```
daveverwer, line 22: @finestructure to check if this is necessary.

Tests/AppTests/Snapshotting+html.swift
```diff
@@ -17,6 +17,12 @@
...
+    }
+}
```
finestructure, line 24: This is a little more concise, that's how `.lines` is defined (with an extension `.txt`):

```suggestion
extension Snapshotting where Value == String, Format == String {
  public static let html = Snapshotting(pathExtension: "html", diffing: .lines)
}
```

PR #1729, opened 2022-05-09
Amended Snapshots workflow with context sensitive comparison

daveverwer:
From https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/pull/1726#issuecomment-1117300136.

The Snapshots workflow now looks for snapshot-affecting files against `origin/main` when on a feature branch and `HEAD^` when running on `main`.

The effect of this is to run Percy if any snapshot-affecting files have been modified in *any* commit on a feature branch, but when merging into main, only run Percy if the merge commit contains snapshot-affecting files.

This still needs testing on `main` but needs merging for that to happen!

### DIFF
--- a/.github/workflows/snapshots.yml
+++ b/.github/workflows/snapshots.yml
@@ -31,10 +31,20 @@ jobs:
         run: make build-front-end
       - name: Copy Front-End
         run: cp -a ./Public/* ./Tests/AppTests/__Snapshots__/WebpageSnapshotTests
+      - name: Determine Comparison Git Reference
+        id: determine-comparison-ref
+        run: |
+          branch=$(git branch --show-current)
+          echo $branch
+          if [ $branch == "main" ]; then
+            echo "::set-output name=comparison-ref::HEAD^"
+          else
+            echo "::set-output name=comparison-ref::origin/main"
+          fi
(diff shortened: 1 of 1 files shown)


finestructure (approved):

Sherlouk:
but Dave! this introduces the 11 runs issue! 😂

daveverwer:
> but Dave! this introduces the 11 runs issue! 😂

Yep, but I don't think there's a better way! 😬

Sherlouk:
Could we use something like `${{ github.event.before }}` to get the last pushed up commit and diff to that?

See https://stackoverflow.com/a/61861763/17582231 for more context.

So the before commit is in replace of `origin/main` - it's whichever was the last commit GitHub ran an action for.

daveverwer:
> Could we use something like `${{ github.event.before }}` to get the last pushed up commit and diff to that?
> 
> See https://stackoverflow.com/a/61861763/17582231 for more context.
> 
> So the before commit is in replace of `origin/main` - it's whichever was the last commit GitHub ran an action for.

Thanks James. This is something we can look at when we start to run out of quota for sure, but my appetite for tweaking this further isn't there right now 😬

PR #1731, opened 2022-05-10
Save SPI manifest to versions table

finestructure:
From this we can construct DocC urls, initially by loading the `SPIManifest.Manifest` from the field and calling `allDocumentationTargets()` on the resulting manifest.

Once we have clearer usage patterns we can decide to pull out info from the manifest into dedicated fields.


Package.swift
```diff
@@ -25,7 +25,6 @@ let package = Package(
         .library(name: "DependencyResolution", targets: ["DependencyResolution"])
     ],
     dependencies: [
-        // 💧 A server-side Swift web framework.
```
daveverwer, line 28: This comment was load-bearing. 😂

daveverwer (approved):

Issue #1721, opened 2022-05-02
Follow up on query performance failures


finestructure:
Query performance checks started failing on or around #1717 

See if it's query or db related and fix or bump costs.

finestructure:
FWIW, it's passing again: https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/actions/workflows/query-performance.yml

<img width="827" alt="CleanShot 2022-05-03 at 11 18 29@2x" src="https://user-images.githubusercontent.com/65520/166430230-e337a6a0-8b88-4727-9be0-4035f1e3998d.png">

Closing this for now.

PR #1734, opened 2022-05-16
Update perf test expactations

finestructure:
Differences are minor, and on some runs the cost from the plans also goes down quite a bit (but it's not stable). For instance:

```
ℹ️ TEST:        test_Search_query_starsFilter()
ℹ️ COST:        3776.28
ℹ️ EXPECTED:    3940.0 ± 150.0
ℹ️ ACTUAL TIME: 72.933ms
```

and on re-run:

```
ℹ️ TEST:        test_Search_query_starsFilter()
ℹ️ COST:        3962.07
ℹ️ EXPECTED:    3940.0 ± 150.0
ℹ️ ACTUAL TIME: 36.325ms
```

Nothing to worry about, just recording the observation.

The point of the test is to get alerted to noticeable increases over time (i.e. having to adjust these costs by 5-10% frequently, which we don't) or to sudden 1.5-2x increases in cost.


finestructure:
Addresses #1721

PR #1735, opened 2022-05-16
Added documentation targets to the Package page sidebar

daveverwer:
Unstyled so far but with snapshot tests for when that happens. There's more to come to make these links actually *work*, but this felt like a good stopping point for a single PR.

### DIFF
--- a/Sources/App/Views/PackageController/PackageShow+Model.swift
+++ b/Sources/App/Views/PackageController/PackageShow+Model.swift
@@ -44,6 +44,7 @@ extension PackageShow {
         var score: Int?
         var isArchived: Bool
         var homepageUrl: String?
+        var documentationTargets: [String]?
         
         internal init(packageId: Package.Id,
                       repositoryOwner: String,
@@ -67,7 +68,8 @@ extension PackageShow {
                       url: String,
                       score: Int? = nil,
                       isArchived: Bool,
-                      homepageUrl: String? = nil) {
+                      homepageUrl: String? = nil,
+                      documentationTargets: [String]? = nil) {
             self.packageId = packageId
             self.repositoryOwner = repositoryOwner
             self.repositoryOwnerName = repositoryOwnerName
@@ -91,6 +93,7 @@ extension PackageShow {
             self.score = score
             self.isArchived = isArchived
             self.homepageUrl = homepageUrl
+            self.documentationTargets = documentationTargets
         }
         
         init?(result: PackageController.PackageResult,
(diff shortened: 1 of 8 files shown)


Tests/AppTests/PackageShowModelTests.swift
```diff
@@ -45,6 +46,35 @@ class PackageShowModelTests: SnapshotTestCase {
...
+                                      latest: .defaultBranch,
+                                      packageName: "Package",
+                                      reference: .branch("main"),
+                                      spiManifest: try Manifest(yml:
+                                         """
+                                         version: 1
+                                         builder:
+                                           configs:
+                                           - documentation_targets: [Target1, Target2]
+                                         """))
```
daveverwer, line 64: This test feels like it's also re-testing the yml parsing of the Manifest, but it was just the quickest and easiest way to get targets into a config.

Sources/App/Views/PackageController/PackageShow+Model.swift
```diff
@@ -154,6 +158,10 @@ extension PackageShow.Model {
     var gitHubRepositoryUrl: String {
         "https://github.com/\(repositoryOwner)/\(repositoryName)"
     }
+
+    func relativeDocumentationURL(target: String) -> String {
+        return "/\(repositoryOwner)/\(repositoryName)/documentation/\(target.lowercased())"
```
finestructure, line 163: Running this locally gives me a 404 when following the documentation url. I think we either need to include the reference fragment (`"/\(repositoryOwner)/\(repositoryName)/\(reference)/documentation/\(target.lowercased())"`) for now or add a route handler/converter/redirect to resolve this to the correct S3 bucket url.

Tests/AppTests/PackageShowModelTests.swift
finestructure: Yeah, I don't think there's much merit to this test. The bit that's relevant - whether the value is properly assigned in the `init` - is tested at a high level in the snapshot test.

The purpose of the `test_init_no_packageName` above was to test a pathological case, which is different.

I'd say we drop it.

Sources/App/Views/PackageController/PackageShow+Model.swift
daveverwer: Apologies! This is intentional, but I didn't explain it well enough in the PR. The fact they don't work is what this comment was referring to:

> There's more to come to make these links actually work

But this felt like enough for a single PR. Making them work will be in a follow-up PR.

Tests/AppTests/PackageShowModelTests.swift
daveverwer: Removed.

finestructure (approved):

PR #1755, opened 2022-05-23
Upgrade build machines to use Xcode 13.4.0/Swift 5.6

### LINKED ISSUE #1749: Upgrade Mac builders to Xcode 13.4/Swift 5.6.1

daveverwer:
Fixes #1749

We should upgrade our build infrastructure to match the latest Xcode and Swift patch.

This change is not urgent at all and probably shouldn't be done until we're fully deployed with DocC, but I was doing builder maintenance this morning and got a start on it by installing 13.3.1 on all Mac build machines.

Checklist:

- [x] Mac build machines have Xcode 13.4.0 installed
- [x] Builder upgraded to target Xcode 13.4.0 on Mac and 5.6.1 on Linux ([MR!116](https://gitlab.com/finestructure/swiftpackageindex-builder/-/merge_requests/116))
- [ ] Xcode 13.3.0 removed from Mac build machines (post-deploy)
- [ ] More?

We won't need to delete build records as we certainly shouldn't re-process everything 5.6 for this. I'll need your input @finestructure on whether anything needs changing with the DB as the `swift_version` field on existing records specifically refers to Swift 5.6.0, and I believe the change in `SwiftVersion+Build.swift` will change that.

### DIFF
--- a/Sources/App/Core/SwiftVersion+Build.swift
+++ b/Sources/App/Core/SwiftVersion+Build.swift
@@ -19,7 +19,7 @@ extension SwiftVersion {
     static let v5_3: Self = .init(5, 3, 3)
     static let v5_4: Self = .init(5, 4, 0)
     static let v5_5: Self = .init(5, 5, 2)
-    static let v5_6: Self = .init(5, 6, 0)
+    static let v5_6: Self = .init(5, 6, 1)
 
     /// Currently supported swift versions for building
     static var allActive: [Self] {
@@ -37,7 +37,7 @@ extension SwiftVersion {
             case .v5_5:
                 return "Xcode 13.2.1"
             case .v5_6:
-                return "Xcode 13.3.0"
+                return "Xcode 13.3.1"
             default:
                 return nil
         }
--- a/Sources/App/Core/SwiftVersion+Build.swift
+++ b/Sources/App/Core/SwiftVersion+Build.swift
@@ -36,7 +36,7 @@ extension SwiftVersion {
             case .v5_5:
                 return "Xcode 13.2.1"
             case .v5_6:
-                return "Xcode 13.3.1"
+                return "Xcode 13.4.0"
             default:
                 return nil
         }
--- a/Sources/App/Core/SwiftVersion+Build.swift
+++ b/Sources/App/Core/SwiftVersion+Build.swift
@@ -36,7 +36,7 @@ extension SwiftVersion {
             case .v5_5:
                 return "Xcode 13.2.1"
             case .v5_6:
-                return "Xcode 13.4.0"
+                return "Xcode 13.4.1"
             default:
                 return nil
         }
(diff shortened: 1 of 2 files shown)


Sources/App/Core/SwiftVersion+Build.swift
```diff
@@ -19,15 +19,14 @@ extension SwiftVersion {
...
-    static let v5_6: Self = .init(5, 6, 0)
+    static let v5_6: Self = .init(5, 6, 1)
 
     /// Currently supported swift versions for building
     static var allActive: [Self] {
         [.v5_3, .v5_4, .v5_5, .v5_6]
     }
 
     var xcodeVersion: String? {
-        // Match with https://gitlab.com/finestructure/swiftpackageindex-builder/-/blob/main/Sources/BuilderCore/SwiftVersion.swift#L41
```
daveverwer, line 30: I believe this is no longer necessary. Certainly, this file reference is no longer valid and the builder project now pulls in SwiftVersion from this project as a dependency, so I believe this comment is no longer needed.

However, I noticed that the builder still references an old version of this project as a dependency that does not include Swift 5.6, which might need looking at.

finestructure: I haven't investigated this further but I'm pretty sure this still required but the reference is in the SPIManifest project not in builder anymore.

daveverwer:
There has now been a new Xcode release since I opened this. It's the same version of Swift, but as I was doing Mac builder maintenance this morning I thought we might as well be on the latest when we merge this.

This PR and MR [!116](https://gitlab.com/finestructure/swiftpackageindex-builder/-/merge_requests/116) on GitLab have both been updated.

Sources/App/Core/SwiftVersion+Build.swift
finestructure: Bumping the versions in the Server project is mostly irrelevant or for display only. For instance, the Xcode version is only shown on a build detail page. It's purely for display. We could even consider dropping the version details entirely or at least truncate it to x.y format to avoid having to deal with it.

The Swift version is only relevant in in its x.y part as well. We're actually triggering `5.6` when triggering a build and it's the builder who picks whatever latest Swift/Xcode version in that bracket it's got.

Here, we can't quite drop the patch version because our type has it. But it doesn't really matter what the patch version is. It gets saved to the `builds` table when we trigger but it's not actually even the true trigger version. I.e. if we didn't update the server and trigger 5.6.0 but the builder has 5.6.2 we'd save 5.6.0 in builds but the build was actually run with 5.6.2.

The true type should really be a Swift version without a patch field.

finestructure (approved):

Sources/App/Core/SwiftVersion+Build.swift
daveverwer: I agree with your philosophy here but for as long as we're keeping patch versions in there they should be correct. Something to think about for a future change, though.

finestructure: Sure, I just mentioned it to clarify that this isn't operationally important or risky and we can deploy these kinds of changes freely on the server side.

PR #1760, opened 2022-05-24
Show a "Targets" bar as part of the DocC generated header

### LINKED ISSUE #1743: Style DocC documentation links
The documentation links need styling up, both for single links and for packages with multiple documentation targets.

Screenshot 2022-05-18 at 15 51 06@2x

daveverwer:
Fixes #1743 

When there is > 1 documentation target:

Screenshot 2022-05-24 at 20 37 18@2x
Screenshot 2022-05-24 at 20 37 44@2x

When there is exactly 1 documentation target:

Screenshot 2022-05-24 at 20 38 23@2x
Screenshot 2022-05-24 at 20 38 32@2x

### DIFF
--- a/Sources/App/Views/DocumentationPageProcessor.swift
+++ b/Sources/App/Views/DocumentationPageProcessor.swift
@@ -22,16 +22,22 @@ struct DocumentationPageProcessor {
     let repositoryOwnerName: String
     let repositoryName: String
     let packageName: String
+    let reference: String
+    let targets: [String]
 
     init?(repositoryOwner: String,
           repositoryOwnerName: String,
           repositoryName: String,
           packageName: String,
+          targets: [String],
+          reference: String,
           rawHtml: String) {
         self.repositoryOwner = repositoryOwner
         self.repositoryOwnerName = repositoryOwnerName
         self.repositoryName = repositoryName
         self.packageName = packageName
+        self.reference = reference
+        self.targets = targets
 
         do {
             document = try SwiftSoup.parse(rawHtml)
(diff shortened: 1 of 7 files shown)


daveverwer:
Looks like the bar isn't rendering correctly with Percy. Could you check it out locally and make sure it works, please, @finestructure?

finestructure:
Looks great:

CleanShot 2022-05-25 at 09 24 09@2x

Sources/App/Views/DocumentationPageProcessor.swift
```diff
@@ -22,16 +22,22 @@ struct DocumentationPageProcessor {
...
     let packageName: String
+    let reference: String
+    let targets: [String]
 
     init?(repositoryOwner: String,
           repositoryOwnerName: String,
           repositoryName: String,
           packageName: String,
+          targets: [String],
+          reference: String,
```
finestructure, line 33: nit: reverse order to other two additions
```suggestion
          reference: String,
          targets: [String],
```

Sources/App/Views/PackageController/PackageShow+Model.swift
```diff
@@ -176,6 +176,7 @@ extension PackageShow.Model {
         "https://github.com/\(repositoryOwner)/\(repositoryName)"
     }
 
+    // Note: When this gets merged back with the refactored SiteURL, note that it's duplicated in `DocumentationPageProcessor`.
```
finestructure, line 179: Probably better than to just make it a static func on `SiteURL` for now. That way we'll address it automatically.

I'll take a look.

finestructure: Never mind, didn't see the additional parameters `repositoryOwner` and `repositoryName` being used.

daveverwer:
> Looks great:

Thanks. I'll log this as a separate issue (#1761) so we don't hold this PR up as I won't have time to look at this until next week.

finestructure (approved):

PR #1766, opened 2022-05-27
Add "images" proxy route

### LINKED ISSUE #1764: Images proxy links broken (Runstone docs)
CleanShot 2022-05-27 at 08 35 23@2x

finestructure:
Fixes #1764

CleanShot 2022-05-27 at 08 41 15@2x

### DIFF
--- a/Sources/App/Controllers/PackageController+routes.swift
+++ b/Sources/App/Controllers/PackageController+routes.swift
@@ -59,6 +59,7 @@ struct PackageController {
         case css
         case data
         case documentation
+        case images
         case js
         case themeSettings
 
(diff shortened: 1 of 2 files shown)


Sources/App/Controllers/PackageController+routes.swift
```diff
@@ -59,14 +59,15 @@ struct PackageController {
...
         case js
         case themeSettings
 
         var contentType: String {
             switch self {
                 case .css:
                     return "text/css"
-                case .data, .themeSettings:
+                case .images, .data, .themeSettings:
                     return "application/octet-stream"
```
finestructure, line 71: I'm not sure if there's a downside to return such a broad content-type for images but it seems to work ok. It would certainly be awkward to have to fiddle with explicit png, jpg etc content types based on the path.

daveverwer: Let's do this until it breaks!

daveverwer (approved):

PR #1767, opened 2022-05-27
Eliminate parsing of dependency code snippet from release url

### LINKED ISSUE #1565: Add dependency code snippet data to PackageShow.Model
Currently, we're constructing the dependency code snippets by parsing the `String` reference and the release url. That's not ideal and has already caused one issue (#1563). We should instead construct these properly when creating the `PackageShow.Model` from the source data without parsing.

finestructure:
Fixes #1565 

Instead of prying apart a release URL to come up with the correct code snippet we're constructing it up front from the available data without parsing.

This should also make it easier to support multiple products at some point.

### DIFF
--- a/Tests/AppTests/__Snapshots__/WebpageSnapshotTests/test_PackageShowView.1.html
+++ b/Tests/AppTests/__Snapshots__/WebpageSnapshotTests/test_PackageShowView.1.html
@@ -118,7 +118,7 @@
                 <span class="branch">main</span>
               </p>
               <form class="copyable_input">
-                <input type="text" data-button-name="Copy Code Snippet" data-event-name="Copy SPM Manifest Code Snippet Button" value=".package(url: &quot;https://github.com/Alamofire/Alamofire&quot;, branch: &quot;main&quot;)" readonly/>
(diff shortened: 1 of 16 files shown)


Tests/AppTests/__Snapshots__/WebpageSnapshotTests/test_PackageShowView.1.html
```diff
@@ -118,7 +118,7 @@ <h4>When working with a Swift Package Manager manifest:</h4>
                 <span class="branch">main</span>
               </p>
               <form class="copyable_input">
-                <input type="text" data-button-name="Copy Code Snippet" data-event-name="Copy SPM Manifest Code Snippet Button" value=".package(url: &quot;https://github.com/Alamofire/Alamofire&quot;, branch: &quot;main&quot;)" readonly/>
+                <input type="text" data-button-name="Copy Code Snippet" data-event-name="Copy SPM Manifest Code Snippet Button" value=".package(url: &quot;https://github.com/Alamofire/Alamofire.git&quot;, branch: &quot;main&quot;)" readonly/>
```
finestructure, line 121: There's just an additional `.git` in all the snippets now, which we were previously not correctly mocking, probably because the release URL we used didn't have it. Either is fine, this is just more realistic, because package urls would have a `.git` suffix.

This is the same for all the updated HTML snapshots.

daveverwer (approved):
This is great 👍

PR #1784, opened 2022-06-03
Issue 1770 s3 content check

### LINKED ISSUE #1770: Update versions based on S3 doc archive content
```
500 - Internal Server Error - http://spi-prod-docs.s3-website.us-east-2.amazonaws.com/parse-community/Parse-Swift/main/documentation/parseswift (ios) is an invalid URL
```

CleanShot 2022-05-31 at 09 43 16@2x

finestructure:
Fixes #1770 

I've gated both the migration and the processing _not_ to happen on prod envs. I was thinking we'd roll this out over the weekend and see how it does on staging. The reason for the gating on non-prod is so that we can still push through any other fixes to prod without having to do an awkward rollback dance on staging.

That is if we push this through to staging before the weekend at all.


Sources/App/Commands/Ingest.swift
```diff
@@ -116,6 +117,26 @@ func ingest(client: Client,
...
+                               packages: packages)
+    
+#warning("ingestFromS3 temporarily excluded from production")
+    if Environment.current != .production {
+        try await ingestFromS3(client: client,
+                               database: database,
+                               logger: logger,
+                               packages: packages)
+    }
+}
```
finestructure, line 133: Only running this on non-production envs

daveverwer (approved):

finestructure:
I'll go ahead and merge this now and then observe how it goes.

PR #1794, opened 2022-06-08
Add debug logging

finestructure:
We're not finding doc archives. `ingest_s3_fetch_count` isn't firing, so it looks like `fetchS3DocArchives` is never being called.


finestructure:
I'll deploy from the branch to debug and might litter more logging depending on how it goes. Will leave this in draft until I'm done.

PR #1795, opened 2022-06-08
Fix fetch doc archives

finestructure:
We were tripping over accessing the owner/repo fields on Repository:

```
[ INFO ] Ingesting (id: 38D534BF-4BE1-498E-B973-E7E885FC3A4E) ... [component: ingest] (App/Commands/Ingest.swift:91)
[ DEBUG ] Ingesting [38D534BF-4BE1-498E-B973-E7E885FC3A4E] [component: ingest] (App/Commands/Ingest.swift:119)
[ DEBUG ] Ingesting from Github ... [component: ingest] (App/Commands/Ingest.swift:122)
[ DEBUG ] updateStatus ops: 1 [component: ingest] (App/Commands/Common.swift:57)
[ DEBUG ] Ingesting from S3 ... [component: ingest] (App/Commands/Ingest.swift:130)
[ DEBUG ] ingestFromS3 version candidates: 4 [component: ingest] (App/Commands/Ingest.swift:296)
[ DEBUG ] ingestFromS3 versions with doc targets: 3 [component: ingest] (App/Commands/Ingest.swift:301)
FluentKit/Field.swift:23: Fatal error: Cannot access field before it is initialized or fetched: owner
```

The problem is that the Github ingestion pass saving the model apparently invalidates the relation to Repository. The fix is to refetch the two fields we need.

I've changed the entire `ingestFromS3` function back to just using `[Package.ID]` as its input to avoid any future temptation to access fields on `Joined<Package, Repository>`.

### DIFF
--- a/Sources/App/Commands/Ingest.swift
+++ b/Sources/App/Commands/Ingest.swift
@@ -301,8 +301,13 @@ func ingestFromS3(database: Database,
         logger.debug("ingestFromS3 versions with doc targets: \(versions.count)")
         guard !versions.isEmpty else { continue }
 
-        guard let owner = pkg.repository?.owner,
-              let repository = pkg.repository?.name else { continue }
+        let repo = try await Repository.query(on: database)
+            .filter(\.$package.$id == pkg.model.id!)
+            .field(\.$owner)
+            .field(\.$name)
+            .first()
+        guard let owner = repo?.owner,
+              let repository = repo?.name else { continue }
 
         let prefix = "\(owner)/\(repository)".lowercased()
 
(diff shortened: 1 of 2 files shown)


finestructure:
I've deployed from this branch and we're now correctly discovering doc archives in S3:

CleanShot 2022-06-08 at 07 25 15@2x

finestructure:
Impact on ingest duration is minimal:

CleanShot 2022-06-08 at 07 28 22@2x

(which is to be expected)

daveverwer (approved):

Issue #1801, opened 2022-06-08
Move or repeat ingestFromS3 to analysis


finestructure:
There's a big conceptual problem with our approach to reading the S3 bucket content only as part of ingestion: if analysis discovers a new version on the default branch, it will create a new version, which initially doesn't have any doc archives. If we leave it until ingestion comes around to populate it, we'll potentially switch the docs link off and back on.

Need to think this through further before deploying the final step.

finestructure:
This has been "fixed" by being the wrong approach entirely 😅

PR #1802, opened 2022-06-10
Rollback ingest from s3

finestructure:
In preparation for #1801, this reverts all `ingestFromS3` changes entirely, i.e. all four PRs:

- #1784  
- #1791 
- #1795 
- #1800 

I don't want to risk leaving bits behind when moving this to analysis and it's easier and cleaner to go through and cherry-pick the parts we need anew.

The only difference to a straight revert is that I've left in migration 50, which added `versions.doc_archives`, and reverted it via a new migration 51, which removes it. This reflects the fact that the field is persisted and we can't just revert the PR to revert it.

Leaving this in draft for now. I'll want to go over this again in detail.


finestructure:
Effectively, this rolls us back to 2.66.3 + #1797, which simply modified the `trigger-builds` command to allow triggering by `version-id` + `platform` + `swift-version`.

PR #1808, opened 2022-06-14
Async await analyze

finestructure:
Integrating the S3 check into analysis finally pushed me over the edge making the a/a transition. The way we need to fetch versions and weave the S3 results in with the rest is just a nightmare within the ELF pipeline.

This is a big change, nonetheless. It went well and the tests highlighted a couple of areas that needed care (continuation on error).

One thing I still need to do is check timing behaviour. ELFs are eager and I need to make sure we don't run things serially now where they weren't before. I _think_ that's taken care of with the task group but I need to confirm timing.

- [x] confirm batch timing

We'll want to let this run on staging for a day or two before pushing this to prod.


finestructure:
Interesting, this has even improved performance. Here's a release build ingestion of 25 packages run locally, 3 runs each:

Before:
- 5.46s
- 6.44s
- 6.32s

After:
- 4.41s
- 3.61s
- 3.77s

This is probably not due to the async/await change per se - we should have the same level of concurrency. If I had to guess it's rather due to less overhead passing around structs because of the way I've changed the processing. It's become much more linear (and readable, maintainable).

In ELF land, we pulled in 25 packages and then packages every processing step into an ever larger array of `Result`s until we saved everything in the end. This was in order to be able to continue processing in case of errors.

In a/a I made an initial conversion using the same structure (but didn't time that, might be interesting to do) but then converted to something different when I realised I could ditch the whole shlepping around of the wide array.

Instead, we're now looping over each package and dealing with everything for a single package.

I.e. before

- get all packages → pass result to next stage
- update repository for all packages → pass result to next stage
- update versions for all packages → pass result to next stage
- ...

Now:

- for each package
- update repository for that package
- update version for that package
- ...

daveverwer (approved):

PR #1810, opened 2022-06-14
Product type search, plugin information on package page

icanzilb:
This PR addresses the discussion and the todo items in https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/discussions/1661

More specifically:

 - adds product type to the search view

 - adds a new filter token for product type "type:plugin", "type:executable", etc (1 at a time, no inversion)

image

 - adds the type search filter to the search, including visualization of the predicate

image

 - adds plugin information to the package page

image

----

To Do:

 - [x] icon for plugins. I'd really like this one https://fontawesome.com/v4/icon/plug but I don't know how to generate the image data for the css file. I've added a placeholder
 - [x] this is my first time using vapor + fluentkit + postgres, I just picked up code that was laying around and stitched together this PR. Someone should have a look and see if it makes sense
 - [x] I left a `TODO` in QueryPerformanceTests because I didn't know what number to put in for the new performance test
- [x] I (@finestructure) want to add a new baseline search view refresh query performance test to main before we merge this to get a feel for the impact of the additional join with products


icanzilb:
@daveverwer as we spoke earlier, opening a WIP PR. It should be mostly working but there are still a bunch of tests to update. 

Also there are a few TODOs at the bottom of the description that I'd need help with from someone who's able to have a look

finestructure:
There's a config change on `main` now that should hopefully enable CI to run on forks. If you rebase it should kick in!

finestructure:
Interesting, it let me rebase via the web UI even on your branch 🎉

finestructure (commented):
I know it's still in draft but I thought I'd note a couple of small things while I was at it!

Tests/AppTests/SearchFilterTests.swift
```diff
@@ -412,6 +413,40 @@ class SearchFilterTests: AppTestCase {
...
+                                               bindableValue: .value("executable"),
+                                               displayValue: "Executable"))
+
+        // test view representation
+        XCTAssertEqual(filter.viewModel.description, "type is Executable")
+
+        // test sql representation
+        XCTAssertEqual(renderSQL(filter.leftHandSide), #""type"->>'executable'"#)
+        XCTAssertEqual(renderSQL(filter.sqlOperator), "IS NOT")
+        XCTAssertEqual(binds(filter.rightHandSide), [])
```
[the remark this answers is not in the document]
icanzilb: Great feedback, as I mentioned it's my first time with PG & I was sure we'd find ways to improve during review, thanks

daveverwer:
> * Icon for plugins. I'd really like this one https://fontawesome.com/v4/icon/plug but I don't know how to generate the image data for the css file. I've added a placeholder

We draw all our own icons, but I have drawn one along this theme for you. I didn't go with an angle as the horizontal weighting of that one might look strange compared to the others and I tried to match our style a bit more. I pushed the changes directly up to your branch.

Here's how it looks:

Screenshot 2022-06-15 at 11 12 15@2x

 Are you happy with that?

daveverwer:
I also hope everyone appreciates the restraint it took to not make it look like this:




😂

daveverwer:
I also may play with a three-column layout for that section now there are three of these, but that would be a separate PR.

icanzilb:
> I also hope everyone appreciates the restraint it took to not make it look like this:

Thank you very much Dave. Are you sure two-pin plug wasn't just 33% less work? 😂

Tests/AppTests/SearchFilterTests.swift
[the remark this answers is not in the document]
icanzilb: Oh! I just found out there was a "special" json operator which checks for a top-level JSON key `?`: https://www.postgresql.org/docs/9.5/functions-json.html

icanzilb:
I am somewhat hesitant what's the best wording for displaying the filter. A package can export products of multiple types but you're usually looking for matches to one of them and so I'm unsure whether is best for the filter to be rendered as "Type is plugin" or "Type contains plugin" ... (or maybe even "matches"?) I'd appreciate your input @daveverwer 

image

Sources/App/Core/Search.swift
```diff
@@ -36,13 +36,15 @@ enum Search {
...
     static let summary = SQLIdentifier("summary")
+    static let productType = SQLIdentifier("type")
 
     static let ilike = SQLRaw("ILIKE")
     static let null = SQLRaw("NULL")
     static let nullInt = SQLRaw("NULL::INT")
     static let nullUUID = SQLRaw("NULL::UUID")
     static let nullTimestamp = SQLRaw("NULL::TIMESTAMP")
     static let nullTextArray = SQLRaw("NULL::TEXT[]")
+    static let nullJSONB = SQLRaw("'{}'::jsonb")
```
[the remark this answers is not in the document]
icanzilb: Seems to be working ok, I'll keep it

icanzilb:
> Are you happy with that?

Yup, looks great — thank you.

finestructure:
I wonder if "type" is the term we should go with. It's really the product type, so maybe "product" would be better in search? That's what the manifest declares. Caveat: do people think of a package in terms of having products of different types? But even if not, I think it's better than just `type`.

Just trying a few things:

1. `foo type:executable` → `type is Executable`
1. `foo product:executable` → `product is Executable`
1. `foo product:executable` → `products contain Executable`
1. `foo product:executable` → `exports Executable`
1. `foo product:executable` → `has Executable product`

I think the last one reads best but I'm not sure that can be made to fit how we generate that string, at least not easily.

If we need to keep the shape for sake of simplicity, I think I'd prefer 2 or 3.

icanzilb:
> If we need to keep the shape for sake of simplicity, I think I'd prefer 2 or 3.

My first iteration for the filter was `product_type:plugin` but it feels a bit verbose in the search field. Though it does transmit a bit better the idea of searching for a "product type" instead of a "package type".

Seeing your suggestions here I think maybe it's better to underline the difference between package and product. I'll add  one more to the list:

6. `foo product_type:plugin` -> 'package contains Plugin product(s)'

daveverwer:
Yes, I think number 6 is getting closest here. Unlike some of the other filter types, this filter isn't about the package itself it's about something that the package containing.

The only amendments I'd make is to use Sven's suggestion of `products` over `product_type` and also to remove the (s) by slightly rephrasing it:

7. `foo product:plugin` -> "Package `products` contains a `plugin`"

Looking something like this:

Screenshot 2022-06-16 at 09 32 34@2x

finestructure:
Sounds good! Note though that it should be “contain” if it’s “products”: “products contain an executable“. And we may want to deal with that “a” / “an” depending on the value.

icanzilb:
Closing the external PR

finestructure:
I'll reopen this in draft so we can finish the review discussion and then we can do the CI bits on a new branch/PR in the SPI repo.

finestructure:
Measuring the search view refresh isn't trivial. We can't measure the refresh directly because `Utility statements have no plan structure`.

We could measure the underlying `CREATE MATERIALIZED VIEW` statement. However, that comes with problems:

- The perf test would need to track any migration changes, i.e. any time search is changed we'd need to remember to update the perf test.
- We can't run `CREATE MATERIALIZED VIEW search` in the perf test, because `search` already exists. We could just recreate it as `_search` and drop it after the test but that's a bit too hacky.

finestructure:
Ah wait - but we can just measure the `SELECT` part, ignoring the `CREATE MATERIALIZED VIEW`. We'd still need to track it with the migration but at least we'd have a record of performance even if it drifts.

icanzilb:
I'm afraid that rewording the product search criteria the way we agreed above requires some structural change in how the filter view model is built. I think probably changing the view model for the search is out of scope here? This is how far I got this without doing more changes to it:

image

finestructure:
Yeah, I was wondering if that'd work without restructuring. I don't think that's worth it at this point, tbh. This looks good to me!

Sources/App/Migrations/052/UpdateSearchAddProductType.swift
```diff
@@ -0,0 +1,81 @@
...
+              r.keywords,
+              r.last_commit_date,
+              r.license,
+              r.name AS repo_name,
+              r.owner AS repo_owner,
+              r.stars,
+              r.last_activity_at,
+              r.summary,
+              v.package_name,
+              pr.type
```
finestructure, line 44: I think we should make this

```sql
pr.type AS product_type
```
```

File: Sources/App/Core/SearchFilter/SearchFilter+Key.swift [hunk 4]
```diff
@@ -61,6 +64,8 @@ extension SearchFilter {
                     return SQLIdentifier("platform_compatibility")
                 case .stars:
                     return SQLIdentifier("stars")
+                case .productType:
+                    return SQLIdentifier("type")
```
finestructure, line 68: And then `product_type` here...

Sources/App/Migrations/052/UpdateSearchAddProductType.swift
```diff
@@ -0,0 +1,81 @@
...
+            FROM packages p
+              JOIN repositories r ON r.package_id = p.id
+              JOIN versions v ON v.package_id = p.id
+              LEFT JOIN products pr ON pr.version_id = v.id
```
finestructure, line 48: One thing we need to consider here is that this will add duplicates to the `search` view.

Prior to this change, the search view has exactly as many rows as `packages`. For instance, with the recent prod dump `spi_prod-2022-06-14.dump` we've got:

```
select count(*) from packages;  -- 4672

-- v5
select count(*) from search;    -- 4672
-- v6
select count(*) from search;    -- 6835
```

The duplicates as packages that have multiple products. This is probably ok but we should double check that we're running a `DISTINCT` query for the final results.

finestructure: Not quite 😅

CleanShot 2022-06-17 at 08 21 10@2x

Sources/App/Models/Package.swift
```diff
@@ -131,6 +131,14 @@ extension Package {
 }
 
 
+extension Package {
+    enum ProductType: String, Codable, CaseIterable {
+        case executable
+        case library
+        case plugin
+    }
+}
```
finestructure, line 140: I think we should move this to `ProductTypeSearchFilter`. It's only used there and we've got then closely related `Product.ProductType` at the `Model` level which this partly "overlays".

Tests/AppTests/SearchTests.swift
```diff
@@ -116,44 +116,58 @@ class SearchTests: AppTestCase {
...
+    func test_packageMatchQuery_ProductTypeSearchFilter() throws {
+        for type in Package.ProductType.allCases {
+            let b = Search.packageMatchQueryBuilder(
+                on: app.db, terms: ["a"],
+                filters: [
+                    try ProductTypeSearchFilter(expression: .init(operator: .is, value: type.rawValue))
+                ]
+            )
+            XCTAssertEqual(renderSQL(b), """
+            SELECT 'package' AS "match_type", NULL AS "keyword", "package_id", "package_name", "repo_name", "repo_owner", "score", "summary", "stars", "license", "last_commit_date", "last_activity_at", "keywords", NULL AS "levenshtein_dist", "type" FROM "search" WHERE CONCAT_WS(' ', "package_name", COALESCE("summary", ''), "repo_name", "repo_owner", ARRAY_TO_STRING("keywords", ' ')) ~* $1 AND "repo_owner" IS NOT NULL AND "repo_name" IS NOT NULL AND ("type" ? $2) ORDER BY LOWER("package_name") = $3 DESC, "score" DESC, "package_name" ASC
```
finestructure, line 133: Sweet, I didn't know about the `?` operator!

finestructure:
BTW, I'm happy to look into the points above myself but I don't want to trample over your PR! 🙂

Let me know if you have the time and we can coordinate wrapping up the last few bits here. This has come out great and in record time!

icanzilb:
> BTW, I'm happy to look into the points above myself but I don't want to trample over your PR! 🙂
> 
> Let me know if you have the time and we can coordinate wrapping up the last few bits here. This has come out great and in record time!

@finestructure if you have few minutes to do those, please go ahead — I don't feel strongly about owning the PR. If you don't find the time to, I'll be back to this likely tomorrow during the day

Tests/AppTests/SearchTests.swift
icanzilb: Yeah, the syntax seems somewhat arbitrary ...

Sources/App/Migrations/052/UpdateSearchAddProductType.swift
icanzilb: So I did rename this as you suggested. All tests pass but search doesn't work anymore 😂 I'm not quite sure how to debug this — when I search locally for "csv" the search fails with the following error below and there doesn't seem to be any debug info about it:

image

Would you know off top of your mind what this could be?

Sources/App/Core/SearchFilter/SearchFilter+Key.swift
icanzilb: Renamed

Sources/App/Migrations/052/UpdateSearchAddProductType.swift
icanzilb: Will address it once I get search working again

Sources/App/Migrations/052/UpdateSearchAddProductType.swift
icanzilb: Nevermind, it seems I merged the wrong branch and now everything is a mess. Sorry

finestructure: No worries! I've just rebased on main, fixing that merge conflict I effectively caused by bringing in the migration ahead of your branch. I'll push this up to the SPI repo to kick off CI. Everything still passing locally.

Shall I force push it into this PR as well?

finestructure: Or maybe this is the point where we transition over to a PR off of that branch and carefully pull out the remaining TODOs from this one here. It's getting quite long =)

icanzilb: @finestructure yeah I think the issue is that we're having alternative implementations - my branch on my fork, my branch on your repo, you're working in your repo at the same time, it got complex to keep up everything in check

icanzilb: In hindsight, might've been a better approach to close this PR and move the work to the branch on your repo

finestructure: I'm pretty sure I cherry picked the missing commit but I'll double check the commit list. I had to rush out earlier and pushed a couple of things partway through. I'll take a look again later today. Hope I didn't make a mess for you!

finestructure:
I've had a look again and everything's fine. Here's what happened, I believe.

I had previously pushed your branch into SPI. You subsequently rebased your local branch onto main and that disconnected your branch from this older branch of yours. You then accidentally merged your old branch back into your local branch (merge commit `df0c55b`). You can see how it repeats all the commits except the one new one you had locally, "rename fields" `98d570c`.

CleanShot 2022-06-18 at 11 48 48@2x

This morning I rebased everything on main again and cherry-picked that one recent commit, "rename fields", as well. This is the state of that branch with three new commits I made this morning:

CleanShot 2022-06-18 at 11 50 28@2x

Everything is up-to-date on that branch as far as I can see. a0b3269 is the most up-to-date branch and it contains all our commits.

All we need to do now is fix the two tests I added this morning that expose the two remaining issues :)

finestructure:
Closing again in favour of #1826

icanzilb:
> You can see how it repeats all the commits except the one new one you had locally, "rename fields" [98d579c](https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/commit/98d579c20db346fa1ed7ae984fab6bc5950557cd).

Yeah 😅 I noticed things went astray and was about to retrace my steps back when my free-time slot for the day ended and the little one woke up. Thanks for looking at this & fixing things

PR #1811, opened 2022-06-15
Add pull_request_target

finestructure:
This should allow us to approve and run CI on forks https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks


daveverwer (approved):

daveverwer:
I think this may be causing our CI tests to now run twice:

Screenshot 2022-06-15 at 17 15 19@2x

PR #1814, opened 2022-06-15
Home page redesign step one

daveverwer:
Screenshot 2022-06-15 at 17 06 47@2x


finestructure (approved):

finestructure:
Mmm, I should have run this locally before merging but I wanted to get it onto dev before I merge async analysis. This looks a bit weird for me on staging now:

CleanShot 2022-06-16 at 04 58 04@2x

I think it's the long package name in the "Recently Added" column that messes things up.

finestructure:
Actually, running locally wouldn't have caught that anyway if that's the problem :)

PR #1817, opened 2022-06-16
Run push trigger only on main

finestructure:
This should fix the double runs:

- `pull_request_target` triggers when there are commits on a PR branch
- `push` triggers when there are commis on `main`

The downside is that this doesn't trigger on branches by themselves anymore, only when they're a PR. I do use that quite a bit as a pre-flight check before opening up PRs. I can just open a draft PR but that's noisy.

I wish GH was smart enough not to double trigger...

Also, whatever setting we decide, we'll need to replicate it in the `query-performance.yml` and `snapshots.yml` workflows.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,6 +17,8 @@ name: Continuous Integration
 on:
   pull_request_target:
   push:
+    branches:
+      - main
   workflow_dispatch:
 
 jobs:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,11 +15,8 @@
 name: Continuous Integration
 
 on:
-  pull_request_target:
-  push:
-    branches:
-      - main
   workflow_dispatch:
+  push:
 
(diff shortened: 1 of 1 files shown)


daveverwer:
Heh I was just typing up a response, but your last comment negates them!

So, I think with this we should bring a few people in as collaborators - @icanzilb, @jeroenleenarts, and @DaemonLoki. We have done this in the past with other contributors who we trust, and the `main` branch is always protected.

I'll send out invites

daveverwer (approved):

finestructure:
Sounds good to me! Also, if someone knows what a safe and convenient setting is here, please let us know!

daveverwer:
> > I think with this we should bring a few people in as collaborators - @icanzilb
> 
> Thank you! I think this week was rather an exception when I stumbled upon some free time, but who knows — I might be able to help also in the future

There's no expectation of any additional help! But now you have signed the CLA and are all set up, you might as well be there 👍

PR #1819, opened 2022-06-16
Stop columns crashing into each other and give a little space

### LINKED ISSUE #1818: Columns crashing into each other on the home page
Columns can still crash into each other, just in a different way now!

Screenshot 2022-06-16 at 12 10 52@2x

daveverwer:
Fixes #1818

Screenshot 2022-06-16 at 12 16 42@2x


daveverwer:
Deploying as 2.69.1

PR #1822, opened 2022-06-17
Add test_Search_refresh

finestructure:


Tests/AppTests/QueryPerformanceTests.swift
```diff
@@ -105,6 +105,39 @@ class QueryPerformanceTests: XCTestCase {
         try await assertQueryPerformance(query, expectedCost: 3940, variation: 150)
     }
 
+    func test_Search_refresh() async throws {
+        // We can't "explain analyze" the refresh itself so we need to measure the underlying
```
daveverwer, line 109: I wonder if it might be worth a comment in `UpdateSearch4` to remind whoever changes it next to also update this?

I know that when I update that materialised view, I copy `UpdateSearchX` and make a `UpdateSearchX+1` and go from there, so I'd see the comment.

daveverwer (approved):

PR #1826, opened 2022-06-18
Product type search, plugin information on package page

finestructure:
Supersedes #1810

To do:

- [x] deploy from branch and re-run perf tests to update costs
- [ ] re-analyze versions on staging


Sources/App/Core/Search.swift
```diff
@@ -82,6 +84,7 @@ enum Search {
         var lastActivityAt: Date?
         var summary: String?
         var keywords: [String]?
+        var productType: ProductType?
```
finestructure, line 87: That was `String?` but that only actually didn't throw on decoding, because previously there was a column name mismatch and the field was always read as `nil` - which isn't actually a problem, because we don't need it's value in the search results display.

The mismatch was that prior to f81902e8e8498cf7eb5e85fe45e792d0808b89ba, the search field we were decoding was `product_type` (see `CodingKeys` below) whereas the `search` view's field at that point was `type`. In f81902e8e8498cf7eb5e85fe45e792d0808b89ba we brought the search view's column name in line with what `DBRecord` was decoding - but then the actual issue arose that the JSONB column can't be decoded into `String?`.

It took me a moment to figure out why this worked originally but then cause a decoding issue! =)

Tests/AppTests/QueryPerformanceTests.swift
```diff
@@ -97,6 +97,14 @@ class QueryPerformanceTests: XCTestCase {
         try await assertQueryPerformance(query, expectedCost: 3900, variation: 150)
     }
 
+    func test_Search_query_productTypeFilter() async throws {
+        let filter = try ProductTypeSearchFilter(expression: .init(operator: .is, value: "plugin"))
+        let query = try Search.query(app.db, ["a"], filters: [filter],
+                                     page: 1, pageSize: Constants.resultsPageSize)
+            .unwrap()
+        try await assertQueryPerformance(query, expectedCost: 0, variation: 150)
```
finestructure, line 105: We can't actually run this before deploying to staging, because staging doesn't have the required migration yet and so the perf test fails (as do all the other search perf tests).

Once I'm done with the docc related re-analysis of versions we can deploy from the branch and update the cost before merging.

finestructure:
This should all be done and wrapped up now (except to update the perf figures). 

Thanks again @icanzilb for tackling this! This will be a great addition to search 👌

Tests/AppTests/QueryPerformanceTests.swift
```diff
@@ -129,13 +137,15 @@ class QueryPerformanceTests: XCTestCase {
...
+              v.package_name,
+              pr.type AS product_type
             FROM packages p
               JOIN repositories r ON r.package_id = p.id
               JOIN versions v ON v.package_id = p.id
+              LEFT JOIN products pr ON pr.version_id = v.id
             WHERE v.reference ->> 'branch' = r.default_branch
             """)
-        try await assertQueryPerformance(query, expectedCost: 9_800, variation: 200)
+        try await assertQueryPerformance(query, expectedCost: 13_700, variation: 200)
```
finestructure, line 148: That's a significant increase but total time is still ~100ms so nothing to worry about given that we only run it once every ~20s. I'm just glad that we've got visibility into this.

Tests/AppTests/QueryPerformanceTests.swift
finestructure: Details for reference:

```
ℹ️ TEST:        test_Search_refresh()
ℹ️ COST:        13682.03
ℹ️ EXPECTED:    9800.0 ± 200.0
ℹ️ ACTUAL TIME: 98.161ms
/__w/SwiftPackageIndex-Server/SwiftPackageIndex-Server/Tests/AppTests/QueryPerformanceTests.swift:148: error: QueryPerformanceTests.test_Search_refresh : failed - Total cost of 13682.03 above threshold of 10000.0 (incl variation)

Query plan:

Nested Loop  (cost=9640.49..13682.03 rows=478 width=319) (actual time=53.248..98.161 rows=6864 loops=1)
  Join Filter: (r.package_id = p.id)
  ->  Hash Right Join  (cost=9640.21..13530.45 rows=478 width=271) (actual time=53.103..84.361 rows=6864 loops=1)
        Hash Cond: (pr.version_id = v.id)
        ->  Seq Scan on products pr  (cost=0.00..3526.97 rows=95597 width=58) (actual time=0.012..12.838 rows=95692 loops=1)
        ->  Hash  (cost=9635.52..9635.52 rows=375 width=245) (actual time=53.048..53.051 rows=4691 loops=1)
              Buckets: 8192 (originally 1024)  Batches: 1 (originally 1)  Memory Usage: 1361kB
              ->  Hash Join  (cost=875.27..9635.52 rows=375 width=245) (actual time=6.537..49.707 rows=4691 loops=1)
                    Hash Cond: ((v.package_id = r.package_id) AND ((v.reference ->> 'branch'::text) = r.default_branch))
                    ->  Seq Scan on versions v  (cost=0.00..8356.73 rows=75073 width=195) (actual time=0.014..32.561 rows=75074 loops=1)
                    ->  Hash  (cost=804.91..804.91 rows=4691 width=208) (actual time=6.452..6.453 rows=4691 loops=1)
                          Buckets: 8192  Batches: 1  Memory Usage: 1187kB
                          ->  Seq Scan on repositories r  (cost=0.00..804.91 rows=4691 width=208) (actual time=0.010..4.016 rows=4691 loops=1)
  ->  Index Scan using packages_pkey on packages p  (cost=0.28..0.30 rows=1 width=80) (actual time=0.002..0.002 rows=1 loops=6864)
        Index Cond: (id = v.package_id)
Planning Time: 3.769 ms
Execution Time: 98.576 ms
```

Tests/AppTests/QueryPerformanceTests.swift
finestructure: Details for reference:

```
ℹ️ TEST:        test_Search_refresh()
ℹ️ COST:        13682.03
ℹ️ EXPECTED:    9800.0 ± 200.0
ℹ️ ACTUAL TIME: 98.161ms
/__w/SwiftPackageIndex-Server/SwiftPackageIndex-Server/Tests/AppTests/QueryPerformanceTests.swift:148: error: QueryPerformanceTests.test_Search_refresh : failed - Total cost of 13682.03 above threshold of 10000.0 (incl variation)

Query plan:

Nested Loop  (cost=9640.49..13682.03 rows=478 width=319) (actual time=53.248..98.161 rows=6864 loops=1)
  Join Filter: (r.package_id = p.id)
  ->  Hash Right Join  (cost=9640.21..13530.45 rows=478 width=271) (actual time=53.103..84.361 rows=6864 loops=1)
        Hash Cond: (pr.version_id = v.id)
        ->  Seq Scan on products pr  (cost=0.00..3526.97 rows=95597 width=58) (actual time=0.012..12.838 rows=95692 loops=1)
        ->  Hash  (cost=9635.52..9635.52 rows=375 width=245) (actual time=53.048..53.051 rows=4691 loops=1)
              Buckets: 8192 (originally 1024)  Batches: 1 (originally 1)  Memory Usage: 1361kB
              ->  Hash Join  (cost=875.27..9635.52 rows=375 width=245) (actual time=6.537..49.707 rows=4691 loops=1)
                    Hash Cond: ((v.package_id = r.package_id) AND ((v.reference ->> 'branch'::text) = r.default_branch))
                    ->  Seq Scan on versions v  (cost=0.00..8356.73 rows=75073 width=195) (actual time=0.014..32.561 rows=75074 loops=1)
                    ->  Hash  (cost=804.91..804.91 rows=4691 width=208) (actual time=6.452..6.453 rows=4691 loops=1)
                          Buckets: 8192  Batches: 1  Memory Usage: 1187kB
                          ->  Seq Scan on repositories r  (cost=0.00..804.91 rows=4691 width=208) (actual time=0.010..4.016 rows=4691 loops=1)
  ->  Index Scan using packages_pkey on packages p  (cost=0.28..0.30 rows=1 width=80) (actual time=0.002..0.002 rows=1 loops=6864)
        Index Cond: (id = v.package_id)
Planning Time: 3.769 ms
Execution Time: 98.576 ms
```

icanzilb:
> Thanks again @icanzilb for tackling this! This will be a great addition to search 👌

Great, thank you Sven ... it was great for me too to get a little bit of experience with a PG + Vapor project!

finestructure:
We're seeing quite significant across-the-board increases in search query performance. I don't think any of them are problematic. However, looking at what's causing them I've just realised that I think we're probably making a more significant a change to the search view than we have to.

We used to have one line per package in search, which allowed us to query it without `DISTINCT`. With product types, we're changing the this to

Package  Product type
foo      library
foo      plugin

via the `LEFT JOIN products`. But there's actually no reason to expand it this way. We can rewrite the search view as follows instead:

Package  Product types
foo.     {library, plugin}

This means all the existing package searches are entirely unchanged by the addition of product filters and the change becomes purely additive.

I've prototyped a different search view which will create the following result:

CleanShot 2022-06-20 at 16 36 54@2x

vs

CleanShot 2022-06-20 at 16 37 04@2x

I don't think it should take that much effort to adjust the product filter to query new array field instead of the JSONB. I'll take a look into what it'd take.

icanzilb:
> I don't think it should take that much effort to adjust the product filter to query new array field instead of the JSONB. I'll take a look into what it'd take.

So this did turn out to be more work than expected ... sorry that you're stuck with wrapping this up

finestructure:
> So this did turn out to be more work than expected

Yea weird, this never happened before 🤣

But no worries at all - this should make things nicer!

daveverwer (approved):

finestructure:
Deploying this now to staging and then I'll re-analyze versions so we can have a final test before deploying. Re-analysis will take a few hours.

PR #1829, opened 2022-06-20
Twitter Spaces CTA

daveverwer:
Remember last week when the home page was really sparse? 😬

Screenshot 2022-06-20 at 14 50 14@2x

Screenshot 2022-06-20 at 14 50 32@2x


daveverwer:
I don't think these changes can possibly cause query performance regressions

daveverwer:
Deploying as 2.70.2

finestructure:
Those failures were due to the schema change I'd tentatively deployed from the product filter branch. I've rolled it back a moment ago, so on main everything should be back to what it was.

PR #1832, opened 2022-06-21
Fixed Twitter Spaces layout on mobile devices

daveverwer:
Screenshot 2022-06-21 at 18 18 09@2x


finestructure:
I can’t test this change  right now but I’ve just seen that the current design is also cut off on the iPad, at least on the mini. Just FYI, in case this might not address this.

daveverwer:
It fixes it on the iPad, too.

Issue #1834, opened 2022-06-21
Include a version switcher in the DocC header breadcrumb


daveverwer:
We're slightly changing the rules on the breadcrumb here and making it active navigation, but with the changes alerting people to being on a non-stable docs version, I think this will work OK.

The dropdown may need custom styling to not look 🤮  but I want to avoid much (if not all) JavaScript here so we don't run the risk of interfering with Vue that powers the DocC section of the page.

![Image](https://user-images.githubusercontent.com/5180/174869502-c020476b-0ef0-419f-bdb9-7f73eace3aa9.png)

PR #1843, opened 2022-06-23
Rename `targets` to `doc_archives` in the `DocumentationPageProcessor`

daveverwer:
In preparation for the changes that will be needed when we switch from `targets` to `doc_archives` across the whole system, I wanted to get these fields renamed now as they are very likely to conflict with other changes I'm about to make for #1834.

There should be no visual difference, and the content is still currently target-based. This is just to reduce the chance of difficult merges later.

### DIFF
--- a/Sources/App/Controllers/PackageController+routes.swift
+++ b/Sources/App/Controllers/PackageController+routes.swift
@@ -116,7 +116,7 @@ struct PackageController {
                                                                  repositoryName: repository,
                                                                  packageName: queryResult.version.packageName ?? repository,
                                                                  reference: reference,
-                                                                 targets: queryResult.version.spiManifest?.allDocumentationTargets() ?? [],
+                                                                 doc_archives: queryResult.version.spiManifest?.allDocumentationTargets() ?? [],
                                                                  rawHtml: body.asString())
                 else {
                     return try await awsResponse.encodeResponse(
--- a/Sources/App/Views/DocumentationPageProcessor.swift
+++ b/Sources/App/Views/DocumentationPageProcessor.swift
@@ -105,9 +105,9 @@ struct DocumentationPageProcessor {
                     )
                 ),
                 .if(doc_archives.count > 1, .div(
-                    .class("targets_wrap"),
+                    .class("doc_archives_wrap"),
                     .div(
-                        .class("inner targets"),
+                        .class("inner doc_archives"),
                         .nav(
                             .ul(
                                 .li(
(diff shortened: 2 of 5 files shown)


Sources/App/Controllers/PackageController+routes.swift
```diff
@@ -116,7 +116,7 @@ struct PackageController {
                                                                  repositoryName: repository,
                                                                  packageName: queryResult.version.packageName ?? repository,
                                                                  reference: reference,
-                                                                 targets: queryResult.version.spiManifest?.allDocumentationTargets() ?? [],
+                                                                 doc_archives: queryResult.version.spiManifest?.allDocumentationTargets() ?? [],
```
finestructure, line 119: `docArchives`?

Sources/App/Views/DocumentationPageProcessor.swift
```diff
@@ -23,21 +23,21 @@ struct DocumentationPageProcessor {
     let repositoryName: String
     let packageName: String
     let reference: String
-    let targets: [String]
+    let doc_archives: [String]
```
finestructure, line 26: dto

Sources/App/Controllers/PackageController+routes.swift
daveverwer: What am I even doing? 🤪

Sources/App/Views/DocumentationPageProcessor.swift
```diff
@@ -189,7 +189,7 @@ struct DocumentationPageProcessor {
     }
 
     // Note: When this gets merged back with the refactored SiteURL, note that it's duplicated in `PackageShow.Model`.
-    func relativeDocumentationURL(target: String) -> String {
-        "/\(repositoryOwner)/\(repositoryName)/\(reference)/documentation/\(target.lowercased())"
+    func relativeDocumentationURL(doc_archive: String) -> String {
+        "/\(repositoryOwner)/\(repositoryName)/\(reference)/documentation/\(doc_archive.lowercased())"
```
finestructure, line 193: couple more here :)

Sources/App/Views/DocumentationPageProcessor.swift
```diff
@@ -104,20 +104,20 @@ struct DocumentationPageProcessor {
...
                                 .li(
                                     .text("Documentation for:")
                                 ),
-                                .forEach(targets, { target in
+                                .forEach(docArchives, { archive in
                                         .li(
                                             .a(
-                                                .href(relativeDocumentationURL(target: target)),
-                                                .text(target)
+                                                .href(relativeDocumentationURL(doc_archive: archive)),
```
finestructure, line 119: and here

finestructure (approved):

PR #1844, opened 2022-06-23
Make a globally available representation of our significant release kinds

daveverwer:
Merge after #1843 
More prep for #1834

### DIFF
--- a/Sources/App/Models/Release.swift
+++ b/Sources/App/Models/Release.swift
@@ -35,3 +35,9 @@ extension Release {
         url = node.url
     }
 }
+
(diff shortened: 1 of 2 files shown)


Sources/App/Models/Release.swift
```diff
@@ -35,3 +35,9 @@ extension Release {
         url = node.url
     }
 }
+
+enum SignificantReleaseKind: String {
```
daveverwer, line 39: This didn't feel like it deserved its own file. Here seems like a reasonable place for it to live.

finestructure (approved):

PR #1851, opened 2022-06-27
Revert "Drive documentation url from S3 doc archives"

finestructure:
This reverts commit ad8898684db20835bd5d36f3c00de4e74df1b333.

I believe this broke almost half of our doc urls. Need to investigate why the urls aren't working as expected.


finestructure:
Re-analysis seems to fix it. I'll re-analyze all 37 failure to see if that clears it up. I'm not sure why two passes of re-analysis didn't sort this out before...

finestructure:
After re-analysis only 9 packages remain that have doc targets on their default branch but no doc archives. And they really do not have doc archives, so discovery is working. I'll re-trigger builds to see what's happening.

PR #1858, opened 2022-06-29
Add keyword counts

### LINKED ISSUE #1257: Enhancements to package keyword display
While it's good to see keywords on package pages, and #1254 is a good step forward, it's not great that many keywords are quite useless, being too specific. I think we could make a set of changes to make this feature much more useful.

* Sort the keywords by the number of times they exist in the index (see mockup below)
* Display the number of packages behind each keyword next to the keyword
* Potentially don't display keywords that only apply to one package

My Sketches 6 - 2021-09-23 12 12 12

I know, @finestructure, the idea is to close issues and *not* replace one closed one with another one 😂

jeroenleenarts:
- Introduce WeighedKeyword model.
- Add migration introducing `weighted_keywords` display.
- Refresh materialized view for keyword_weights.
- Add CSS for keyword counts.
- Fix unit tests.

Fixes #1257

image


Tests/AppTests/WebpageSnapshotTests.swift
```diff
@@ -39,20 +39,41 @@ class WebpageSnapshotTests: SnapshotTestCase {
...
+            }
+        }
         var model = PackageShow.Model.mock
-        model.keywords = ["tag1", "tag2", "tag3", "tag4", "tag5", "tag6", "tag7", "tag8", "tag9"]
+        let keywords = ["tag1", "tag2", "tag3", "tag4", "tag5", "tag6", "tag7", "tag8", "tag9"]
+        model.keywords = keywords
+        let weightedKeywords = keywords.map {
+            return PackageShow.Model.WeightedKeyword(keyword: $0, weight: factorial(keywords.firstIndex(of: $0) ?? 1))
+        }
+        model.weightedKeywords = weightedKeywords
```
finestructure, line 57: I'd say let's just keep it simple and stick some values in an array :)

```swift
[("tag1", 123), ...].map(PackageShow.Model.WeightedKeyword.init)
```

Tests/AppTests/WebpageSnapshotTests.swift
```diff
@@ -39,20 +39,41 @@ class WebpageSnapshotTests: SnapshotTestCase {
...
+        let keywords = ["tag1", "tag2", "tag3", "tag4", "tag5", "tag6", "tag7", "tag8", "tag9", "tag10",
                           "tag11", "tag12", "tag13", "tag14", "tag15", "tag16", "tag17", "tag18", "tag19", "tag20",
                           "tag21", "tag22", "tag23", "tag24", "tag25", "tag26", "tag27", "tag28", "tag29", "tag30",
                           "tag31", "tag32", "tag33", "tag34", "tag35", "tag36", "tag37", "tag38", "tag39", "tag40"]
+        model.keywords = keywords
+        let weightedKeywords: [PackageShow.Model.WeightedKeyword] = keywords.map {
+            let index = keywords.firstIndex(of: $0) ?? 1
+            return PackageShow.Model.WeightedKeyword(keyword: $0, weight: index * index * index)
+        }
+        model.weightedKeywords = weightedKeywords
```
finestructure, line 75: Dto.

finestructure (commented):
Hey Jeroen, thanks a lot for the PR!

I've got a number of smallish comments and the tests are throwing a db related assert for me at the moment.

If you don't mind I'll push those changes up into your branch over the next couple of days.

Do the tests run for you? I'm getting a

```
2022-06-30T17:57:08+0200 debug codes.vapor.application : component=server database-id=psql DROP MATERIALIZED VIEW weighted_keywords []
Vapor/Application.swift:164: Fatal error: Application.shutdown() was not called before Application deinitialized.
```

even though I've reset the test db and there should be no schema problem.

jeroenleenarts:
@finestructure Adopted your work in progress. Learning a lot here. So all good. 👍

Sources/App/Views/Extensions/NumberFormatter+default.swift
```diff
@@ -21,6 +21,23 @@ extension NumberFormatter {
         f.numberStyle = .decimal
         return f
     }()
+    
+    static func kString(from value: Int) -> String? {
+        if value < 1000 {
+            return Self.spiDefault.string(from: value)
+        } else {
+            let thousands: Int = value / 1000
+            let fraction: Int = (value - thousands * 1000) / 100
```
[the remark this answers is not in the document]
jeroenleenarts: About the `DefaultStringInterpolation` extension, wouldn't that cause all Int values to be interpolated like this? Is this desired behaviour?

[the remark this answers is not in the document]
finestructure: No, the additional parameter takes care of that, i.e. `”\(value)”` vs `”\(kString: value)”`.

Sources/App/Models/WeightedKeyword.swift
```diff
@@ -0,0 +1,50 @@
...
+    static func query(on database: Database, keywords: [String]) async throws -> [Self] {
+        guard let db = database as? SQLDatabase else {
+            fatalError("Database must be an SQLDatabase ('as? SQLDatabase' must succeed)")
+        }
+
+        return try await db.select()
+            .column(Field.keyword)
+            .column(Field.count)
+            .from(schema)
+            .where(Field.keyword, .in, keywords)
```
jeroenleenarts, line 40: @finestructure This where causes three unit test fails.

Sources/App/Migrations/054/CreateWeightedKeywords.swift
```diff
@@ -0,0 +1,41 @@
...
+    }
+
+    func revert(on database: Database) -> EventLoopFuture<Void> {
+        guard let db = database as? SQLDatabase else {
+            fatalError("Database must be an SQLDatabase ('as? SQLDatabase' must succeed)")
+        }
+
+        return db.raw("DROP MATERIALIZED VIEW weighted_keywords").run()
+    }
+}
```
finestructure, line 41: I commented on the commit itself a moment ago on my iPad but now I can't see where that actually went. The problem wasn't that the migration was async - it should actually be, for future-proofing. The problem was that the initial implementation had a `db.raw(dropSQL)` statement in the `prepare` step. Since this is the _first_ migration that creates the view that will throw the error. I'm guessing you'd copied this from a search view where it's updating an existing search view - and that's why that was necessary in that case.

jeroenleenarts: I see. I will make sure the migration is like this.

jeroenleenarts: And I have been squashing and force pushing on this branch. So you are not crazy. 😇

jeroenleenarts:
@finestructure All review comments resolved. Only needs a design tweak by Dave now.

finestructure (approved):
Looks great, thanks again Jeroen!

finestructure:
Can't log into Percy for some reason to approve the snapshots 🤷‍♂️

daveverwer (approved):
Thanks so much @jeroenleenarts, this is amazing work. 🚀

As promised, I tweaked the CSS a bit and here’s how it now looks.

Screenshot 2022-07-04 at 12 16 47@2x

Screenshot 2022-07-04 at 12 17 01@2x

If you’re interested @jeroenleenarts, some explanation for the changes:

```css
li.keywords {
  grid-column-start: span 2;
  background-image: var(--image-tag);

  a {
    display: flex;
    align-items: center;
    gap: 5px;

    span {
      padding: 1px 5px;
      border-radius: 20px;
      color: var(--page-text);
      background: var(--keyword-count-background);
      font-size: 10px;
    }
  }
}
```

The key thing I did was separate the display of the tags themselves from the layout and spacing of the tags. The containing `a` tag now deals with the spacing by flexing its context, allowing us to use more semantic CSS like `gap` rather than a `margin-left`. That also makes this CSS easier to refactor for elsewhere in the future in cases where we need different spacing or maybe the tag would be on the left of the content elsewhere.

The other significant change I made was to how you did the colours. You were correct that we’d need both an explicit foreground and background colour on the tags. The foreground needs overriding as it’s inside a link and so would use the link colour by default, and the background is obvious. The biggest change is that we could re-use a generic colour though, the text in those tags should be the same as the page’s text, so let’s use that. I also made the background a semi-opaque tint rather than an explicit colour.

Final change, in preparation for #1870, I have moved the display of keyword count `span` CSS into a more generic location and restored the class. We’ll be able to use that in both places now.

PR #1861, opened 2022-06-29
typo fix: st -> set

heckj:
tiny typo fix - ran across it while reading through and setting up a local replica of the server env


daveverwer (approved):
Thanks Joe!

PR #1866, opened 2022-07-01
Add documentation indicator to search results.

### LINKED ISSUE #1792: Highlight packages with documentation in search results

jeroenleenarts:
- Update DB Scheme to include has_docs field.
- Add has_docs to all queries and display "Has documentation" string in search results.

Only needs a nice icon and some CSS.

image

Fixes #1792

- [x] fix unit tests, half way through
- [x] Get asset for documentation item
- [x] Tweak css


Sources/App/Migrations/055/UpdateSearchAddHasDocs.swift
```diff
@@ -0,0 +1,85 @@
...
+              r.keywords,
+              r.last_commit_date,
+              r.license,
+              r.name AS repo_name,
+              r.owner AS repo_owner,
+              r.stars,
+              r.last_activity_at,
+              r.summary,
+              v.package_name,
+              v.doc_archives IS NOT NULL AS has_docs,
```
finestructure, line 47: We'll also need to check `doc_archives = '{}'`, i.e.

```sql
doc_archives is null or doc_archives = '{}' AS has_docs
```
```

Comment (finestructure) [reply Sources/App/Migrations/055/UpdateSearchAddHasDocs.swift:47]: or

```sql
array_length(doc_archives, 1) >= 1 AS has_docs
```

jeroenleenarts: Good catch, collections can also be empty.

daveverwer:
First try, at commit a02994dc4228af577f65974eed6717f1fc301c02 if we want to come back to it.

Screenshot 2022-07-04 at 18 43 43@2x

It looks good at full size...

Screenshot 2022-07-04 at 18 47 29@2x

... but it's not clear to someone with my eyesight what it is at such a small size.

I have another idea.

daveverwer:
Getting there, but I think it needs to be white lines on a solid black background.

Screenshot 2022-07-04 at 19 05 54@2x

daveverwer:
I think this might be the one.

Screenshot 2022-07-04 at 19 14 02@2x

Any opinions @jeroenleenarts?

jeroenleenarts:
Like this one a lot more compared to the open book.

Documentation is more a sheet of information and not a long form novella.

PR #1872, opened 2022-07-05
Revert "Drive documentation url from S3 doc archives"

finestructure:
This reverts commit ad8898684db20835bd5d36f3c00de4e74df1b333.

It looks like this change is a source of 404s. Let's revert this to see if that fixes it. (Unfortunately our logging doesn't show which urls 404 - that needs fixing separately.)

CleanShot 2022-07-05 at 14 21 17@2x

### DIFF
--- a/Sources/App/Views/PackageController/PackageShow+Model.swift
+++ b/Sources/App/Views/PackageController/PackageShow+Model.swift
@@ -117,7 +117,16 @@ extension PackageShow {
                 let packageId = result.package.id
             else { return nil }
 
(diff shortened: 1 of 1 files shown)


finestructure:
(2.71.7 isn't what switched this on but reverting to the "old" way of generating the doc urls should short circuit any issues caused by 2.71.7.)

daveverwer (approved):

finestructure:
Well, whatever the source of the 404 is, it wasn't related to the doc archives change:

CleanShot 2022-07-06 at 11 56 19@2x

PR #1875, opened 2022-07-06
Remove WeightedKeywordModel in favour of base model

finestructure:
I wanted to revisit this, because it bothered me that we had to make it `WeightedKeywordModel` due to a file name clash with the base model type. I think in this case it's absolutely fine we stick with the base type.

The main reason we work with a lot of view model types is that most of the app model types are `Model` types which are actually classes with a lot of Fluent machinery behind them. While they can be instantiated just fine for basic tests, for instance all their relationship accessors need to be handled with care. They're definitely not the type of thing we'd want to pass to a view as the data container.

`WeightedKeyword` is different in that it's not a Fluent type. It's just a plain struct with some static methods to instantiate an array of them from the materialised view and I think in this case it's fine to just use it like that without copying it to a view model.

If I had to come up with a rule, I'd say view model types (and types used to compose view models) should be value types that contain a narrow set of data for a specific domain. That's to say, even if for instance `Repository` was a struct, it'd probably not be a great view model type because it carries a lot of fields, most of which are likely irrelevant for a particular view. In cases like that it still make sense to derive a specific view model subset.

All that is to say, `WeightedKeyword` is really as basic a value type as it could be and we should just use it like that.

Does that make sense?


Sources/App/Controllers/SearchController.swift
```diff
@@ -30,17 +30,12 @@ struct SearchController {
...
-        if page == 1 {
-            weightedKeywords = try await WeightedKeyword.query(on: req.db,keywords: matchedKeywords)
-                .map { WeightedKeywordModel(keyword: $0.keyword,
-                                             weight: $0.count) }
-        } else {
-            weightedKeywords = []
-        }
-        
+        // We're only displaying the keyword sidebar on the first search page.
+        let weightedKeywords = (page == 1)
```
jeroenleenarts, line 35: I specifically avoided using the tertiary statement in my old. code. But seeing it like this, yes. Agreeable code.

jeroenleenarts (approved):
Did a quick evaluation. I like your suggestion. It simplifies things a bit and removes a meaningless conversion.

Sources/App/Controllers/SearchController.swift
finestructure: Yeah, I only collapsed it because we can now get rid of the additional map and shortened the whole expressions enough to bring it sort of into `?:` territory :)

daveverwer (approved):

PR #1886, opened 2022-07-11
Keyword search tweak

heckj:
experimenting with search tweaks to improve search relevancy, starting with keyword relevancy

### DIFF
--- a/Sources/App/Core/Search.swift
+++ b/Sources/App/Core/Search.swift
@@ -37,6 +37,7 @@ enum Search {
     static let searchView = SQLIdentifier("search")
     static let summary = SQLIdentifier("summary")
 
+    static let inVector = SQLRaw("@@")
     static let ilike = SQLRaw("ILIKE")
     static let null = SQLRaw("NULL")
     static let nullInt = SQLRaw("NULL::INT")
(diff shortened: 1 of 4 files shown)


Sources/App/Core/Search.swift
```diff
@@ -235,7 +268,9 @@ enum Search {
         select = select
             .from(unnest(keywords), as: keyword)
         select = select
-            .where(keyword, ilike, SQLBind(searchPattern))
+//            .where(keyword, ilike, SQLBind(searchPattern))
+            .where(to_tsquery(SQLBind(searchPattern)), inVector, SQLRaw("to_tsvector(keyword)"))
```
finestructure, line 272: I missed this PR originally but if it work to simply replace the `ILIKE` in this way then that's great news for extending the search!

One thing to note: this is only modifying the keyword search. The package search is in another subquery. (I'm guessing you've seen this and just used keywords to get started with this in principle - just FYI in case you haven't.)

heckj: sorry @finestructure  - I didn't really have my "stuff" together to actually make a real PR, I was just prepping and got exuberant with the buttons. What's above is - yeah - just keyword update twiddling, but did work. I think even this make a slight difference on the some of the awkward search points today, but the real win will be applying something of the same technique to the packageName search setup.

I'm also a bit uncertain about if or how I should be using SQLBind, and if I should ultimately be encapsulating the bits like `SQLRaw("to_tsvector(keyword)")` into one of the structs that the code uses as aliases - still definitely learning SQLKit.

Now that I've got my search "grabber and ranker" app mostly working (still need to arrange collections from `localhost` to capture my development result) I'll be circling back to this and bolstering it up (and fixing the tests to match, etc)

finestructure: Yeah, we'd definitely want to avoid `SQLRaw` as much as possible, in particular in places where we handle user input. We can write little function adapters for `to_tsvector` et al (we've got a few already) that are then callable like normal Swift functions in SQLKit contexts.

PR #1892, opened 2022-07-13
Drive default doc url via first doc archive (again)

finestructure:
Tested locally by pulling a prod copy after all the doc re-builds were done and pointing the S3 url at the prod bucket.

All doc links are presents as expected and I've manually confirmed the vast majority of them.

One package loses the doc link, and that's ParseSwift, because it's using `macos-xcodebuild` for its doc building platform, which we don't support yet. That's actually a bug I introduced in their `.spi.yml` when they'd shifted away from iOS docs to macos.

PR to correct here: https://github.com/parse-community/Parse-Swift/pull/383

### DIFF
--- a/Sources/App/Views/PackageController/PackageShow+Model.swift
+++ b/Sources/App/Views/PackageController/PackageShow+Model.swift
@@ -117,16 +117,7 @@ extension PackageShow {
                 let packageId = result.package.id
             else { return nil }
 
-            #warning("temporary hotfix until #1770 is properly addressed")
-            let docTargetOverrides = [
-                "https://github.com/apple/swift-docc.git".lowercased() : "DocC",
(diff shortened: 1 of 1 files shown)


finestructure:
Back to draft - I want to test this from a local instance running a prod copy and pointing at the prod bucket first.

finestructure:
Query Performance Test is consistently recording an out of bounds maximum of 19340 (expected: 15000). There are obviously no changes in the PR that would cause this.

A [run on current main](https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/runs/7334366188?check_suite_focus=true) shows that this is already an issue on `main`.

Previous runs were successful, so this is due to data changes in staging. The failing test is `QueryPerformanceTests.test_Search_refresh` and it's likely related to the data changes in `doc_archives`.

I'll take a look into this separately.

daveverwer (approved):

PR #1905, opened 2022-07-19
revised search - ranking results using ts_vector and ts_rank from Postgres on english-only word vectors

heckj:
implementation (and copious experimental notes, which I WILL remove) based on the improving search discussion thread https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/discussions/1876

The code is _mostly_ functional, and in the state where I'm trying to refine initial efforts where I used `SQLRaw` to directly embed the various Postgres functions: `ts_rank`, `to_tsvector`, and `plainto_tsquery`. I wanted to get this up here so y'all could at least see the explicit work.

This version is retaining the `%ILIKE%` for actually identifying what should be returned, so it still does return irrelevant results for the query `ping`, but doesn't drop out the UIBezierCurve in the case of a search on `bezier` - and it doesn't trash the non-english queries (mandarin, for example) which rely on direct pattern matching.

There's been NO fixing of the tests (yet) - this dramatically changes the rendered SQL to extend the union to include a ranking value (`tsrankvalue`) from each sub-query set, and then orders the end results. Part of the code that I have commented out is attempting to make some weighting judgements to influence the weighting of various terms, with the hope that something like a direct package name match wouldn't need that prefixed sort order on "does the package name equal exactly". But... something I'm doing isn't working, and I haven't sorted out how to debug that part yet.

I've also not checked the performance implications of these searches. It's generating these ts_vectors on the fly - some or all of which may be more efficient as pre-generated within the establishment of the materialized search view. It's not brutally noticeable, but I also just haven't checked the numbers.

### DIFF
--- a/Sources/App/Core/Extensions/SQLKit+functions.swift
+++ b/Sources/App/Core/Extensions/SQLKit+functions.swift
@@ -49,11 +49,65 @@ func lower(_ arg: SQLExpression) -> SQLFunction {
     SQLFunction("LOWER", args: arg)
 }
 
-
 func unnest(_ array: SQLExpression) -> SQLFunction {
     SQLFunction("UNNEST", args: array)
 }
 
+func to_tsvector(_ array: SQLExpression) -> SQLFunction {
+    // The argument is meant to be a string, which this wraps and converts
+    // first into a tsvector internal type, and then applies weighting
+    // for query ranking purposes.
+    // The default weight class ('D') results in a  0.1 multiplier on the rank.
+    SQLFunction("to_tsvector", args: array)
+}
+
+func to_tsvectorA(_ array: SQLExpression) -> SQLFunction {
+    // The argument is meant to be a string, which this wraps and converts
+    // first into a tsvector internal type, and then applies weighting
+    // for query ranking purposes.
+    // Details of the setweight function (and to_tsvector) are available
+    // at https://www.postgresql.org/docs/current/textsearch-controls.html
+    // The weight class 'A' results in a 1.0 multiplier on the rank.
+    SQLFunction("setweight", args: [to_tsvector(array), SQLRaw("'A'")])
+}
+
+func to_tsvectorB(_ array: SQLExpression) -> SQLFunction {
+    // The argument is meant to be a string, which this wraps and converts
+    // first into a tsvector internal type, and then applies weighting
+    // for query ranking purposes.
+    // The weight class 'B' results in a 0.4 multiplier on the rank.
+    SQLFunction("setweight", args: [to_tsvector(array), SQLRaw("'B'")])
+}
+
+func to_tsvectorC(_ array: SQLExpression) -> SQLFunction {
+    // The argument is meant to be a string, which this wraps and converts
+    // first into a tsvector internal type, and then applies weighting
+    // for query ranking purposes.
+    // The weight class 'C' results in a 0.2 multiplier on the rank.
+    SQLFunction("setweight", args: [to_tsvector(array), SQLRaw("'C'")])
+}
+
+func plainto_tsquery(_ array: SQLExpression) -> SQLFunction {
+    // generates a simplistic tsquery concatenating all individual words as
+    // 'AND' tokens for the ts_query binary search structure. For example,
+    // an input of 'bezier curve' becomes 'bezier & curve'.
+    SQLFunction("plainto_tsquery", args: array)
+}
+
+func ts_rank(vector: SQLExpression, query: SQLExpression) -> SQLFunction {
+    // returns a ranking value when applying the query to the relevant
+    // tsvector data type. If the query wouldn't match at all, the ranking
+    // returns as `0`. The documentation for the return values is non-specific
+    // about the range, but hints that values can easily exceed 1.0. In
+    // my experimentation, and based on reading the code at
+    // https://github.com/postgres/postgres/blob/master/src/backend/utils/adt/tsrank.c
+    // the returned rank values range from 0 to 1.0, with a single query term
+    // perfectly matching the response returning a ranking of 0.6. Additional
+    // query terms adjust the ranking - OR queries maintain or reduce the value
+    // depending on matches with the vector, and AND queries increase the value
+    // pressing it slowly up towards 1.0.
+    SQLFunction("ts_rank", args: [vector, query])
+}
 
 // MARK: - SQL Binary Expressions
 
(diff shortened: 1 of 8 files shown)


heckj:
I did some quick checking the search metrics for the variety of searches. Only the "ping" search result was impacted, and this update *doesn't* improve recall or precision metrics, but does improve the positioning of the relevant results:

- Jul 8, 2022 (swiftpackageindex.com) - the deployed server content
   - Precision: 0.020
   - Recall: 0.020
   - Mean Reciprocal Rank: 0.071
   - NDCG: 0.256
'ping.searchrank' query terms: ping
 - Jul 18, 2022 (127.0.0.1) - the PR content 
   - Precision: 0.020
   - Recall: 0.020
   - Mean Reciprocal Rank: 1.000
   - NDCG: 1.000

(and yes, I'll happily collapse this all down into a single commit. Apologies if the PR seems abusive with a bunch of commits in the middle, I work in a messy sort of fashion)

Sources/App/Core/Extensions/SQLKit+functions.swift
```diff
@@ -49,20 +49,55 @@ func lower(_ arg: SQLExpression) -> SQLFunction {
...
 }
 
-
 func unnest(_ array: SQLExpression) -> SQLFunction {
     SQLFunction("UNNEST", args: array)
 }
 
+func to_tsvector(_ array: SQLExpression) -> SQLFunction {
+    // vectors without setweight around them default to weight class 'D'
+    // which is a 0.125 multiplier on the rank
```
finestructure, line 58: If I'm reading the page you linked below correctly, that'd be 0.1, wouldn't it? From https://www.postgresql.org/docs/current/textsearch-controls.html

> The weight arrays specify how heavily to weigh each category of word, in the order:
> 
> {D-weight, C-weight, B-weight, A-weight}
>
> If no weights are provided, then these defaults are used:
>
> {0.1, 0.2, 0.4, 1.0}

Not that the detail matters here, just to make sure I'm following along!

Sources/App/Core/Extensions/SQLKit+functions.swift
```diff
@@ -49,20 +49,55 @@ func lower(_ arg: SQLExpression) -> SQLFunction {
...
+    SQLFunction("setweight", args: [to_tsvector(array), SQLRaw("'B'")])
+}
+
+func to_tsvectorC(_ array: SQLExpression) -> SQLFunction {
+    // weight class 'C' results in a 0.25 multiplier on the rank.
+    SQLFunction("setweight", args: [to_tsvector(array), SQLRaw("'C'")])
+}
+
+func plainto_tsquery(_ array: SQLExpression) -> SQLFunction {
+    SQLFunction("to_tsquery", args: array)
```
finestructure, line 81: Shouldn't this be `plainto_tsquery`?

Sources/App/Core/Search.swift
```diff
@@ -175,7 +194,34 @@ enum Search {
...
+//            setweight(to_tsvector(coalesce(abstract,'')), 'C') ||
+//            setweight(to_tsvector(coalesce(body,'')), 'D');
+        
+//            .from(vecMerge(
+//                to_tsvectorB(coalesce(keywords, emptyString)), // keyword vector as weight B
+//                    vecMerge(
+//                        to_tsvectorA(coalesce(packageName, emptyString)), // package name vector  as weight A
+//                        to_tsvector(coalesce(summary, emptyString)) // summary vector as weight D
+//                    )
+//                ), as: tsvector)
```
finestructure, line 224: We'd definitely want to either find an SQLKit way to express it with the existing bind variable mechanisms or at least make sure the `SQLRaw` way isn't open to SQL injection. I suspect it wouldn't be but I'm not sure.

I'm going to be a bit pressed for time in the next week or two but hopefully I'll find some time to take a look!

Sources/App/Core/Search.swift
```diff
@@ -186,6 +232,127 @@ enum Search {
...
+     -- between search values - the distances between them in tsvector.
+     
+     -- On the plus side, we can get a rank for a non-matching term that reliably comes back as 0, so we can
+     -- work with that to provide additional tweaking. Example:
+     
+     SELECT query, a_vector, ts_rank(a_vector, query) AS rank
+     FROM plainto_tsquery('fiddlesticks') query,
+          to_tsvector('Heres my example package of swift package manager') a_vector
+--      WHERE query @@ a_vector
+     ORDER BY rank DESC;
```
finestructure, line 351: Should we schedule a call to go through this at some point? There's a lot of interesting stuff here, and it might be best to go through this together in a quick back and forth rather than an extended async PR discussion?

finestructure:
> I've also not checked the performance implications of these searches. It's generating these ts_vectors on the fly - some or all of which may be more efficient as pre-generated within the establishment of the materialized search view. It's not brutally noticeable, but I also just haven't checked the numbers.

We have some search related performance tests, which should cover this but it's definitely something we'll want to extend if there are gaps.

We'll probably also want to think through if we can somehow capture search behaviour in an integration test or something mocked if that's possible, to assert baseline behaviour more than we do currently (which is very basic).

Sources/App/Core/Extensions/SQLKit+functions.swift
heckj: Yep, you've got it right. The comments in the code for ts_rank.c indicated it was the full values that I had in my comments originally, but at the top they have defines that set it down to those specific numbers.

Sources/App/Core/Search.swift
heckj: I suspect my error is in the SQL, I need to iterate and make sure I was doing something legit, but I'm confident that we can make it work with SQLKit once I nail down the correct underlying SQL.

I wasn't 100% sure about how to properly form a SQLFunction with just two, three, or four parameters - so in the case of a SQLKit version of `tsquery`, I made it explicitly named parameters. The tsvector concatenation gets a little more awkward seen in swift code though.

Sources/App/Core/Search.swift
heckj: Sounds excellent to me! When you're not slammed, let me know and I can step through things. I'd love the feedback on "proper" SQLKit structure in case I've chosen incorrectly, and in the mean time I've a few pieces to iterate out (the SQLKit for concatenating the tsvectors with weights, getting tests operational, verifying the performance characteristics)

heckj:
cleaned up all the mess in there, moving comment blocks on experimentation over to [the related discussion](https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/discussions/1876), and rebased down to a single commit in order to make reviewing this a bit easier.

heckj:
NOTE: When I was updating the search related tests to match the implementation updates leveraging ts_vector, I found that the time tests elsewhere in the project were "just right" to be sensitive to time zone changes - so where some tests were expecting Jan 1, I was getting Dec 31 from the previous year - which is causing some unit tests to fail for me locally. However, I suspect that if I fix them for me "me" locally, then we'll have failures in CI.

heckj:
closing in favor of #1933

PR #1908, opened 2022-07-19
Guide visitors towards latest stable documentation where available

### LINKED ISSUE #1835: Guide people towards the latest stable version DocC documentation
If the project has a stable release and the user is looking any other documentation version, include a note in the header that highlights that they're on a beta/branch version and gives them a link to the stable version.

daveverwer:
Fixes #1835

Screenshot 2022-07-19 at 19 11 00@2x


Sources/App/Controllers/PackageController+routes.swift
```diff
@@ -152,11 +152,18 @@ struct PackageController {
                     documentationVersions.filter { $0.latest == .preRelease }.first
```
finestructure, line 152: Unrelated but I just noticed it:

```suggestion
                    documentationVersions.first { $0.latest == .preRelease }
```

(also the line above but I can select that as part of the diff.)

Sources/App/Views/DocumentationPageProcessor.swift
```diff
@@ -235,4 +258,19 @@ struct DocumentationPageProcessor {
...
+            case .preRelease: return "This documentation is from a pre-release of \(packageName) and may not reflect the latest stable release."
+            case .defaultBranch: return "This documentation is from the \(packageName) \(reference) branch and may not reflect the latest stable release."
+            default: return ""
+        }
+    }
+}
+
+extension Array where Element == DocumentationPageProcessor.AvailableDocumentationVersion {
+    var latestStableVersion: DocumentationPageProcessor.AvailableDocumentationVersion? {
+        self.filter(\.isLatestStable).first
```
finestructure, line 274:
```suggestion
        first(where: \.isLatestStable)
```

but tbh I wouldn't push a simple filter into an extension like this. I think this is clearer at the call site as `first(where: \.isLatestStable)` because it tells you how it's finding that version.

Sources/App/Views/DocumentationPageProcessor.swift
```diff
@@ -147,6 +145,31 @@ struct DocumentationPageProcessor {
                         )
                     )
                 ),
+                .if(Environment.current != .production,
+                    .if(referenceKind != .release,
+                        // Only try and show a link to the latest stable if there *is* a latest stable.
+                        .unwrap(allAvailableDocumentationVersions.latestStableVersion) { latestStable in
```
finestructure, line 151:
```suggestion
                        .unwrap(allAvailableDocumentationVersions.first(where: \.isLatestStable)) { latestStable in
```

(see below)

finestructure (approved):

Sources/App/Views/DocumentationPageProcessor.swift
daveverwer: Yea, when I first made the extension, it did more than this but eventually came down to a filter on a boolean. Good catch 👍

PR #1920, opened 2022-07-25
Add A "Has Binary Targets" Line To Metadata

samdeane:
See #1899 for previous discussions.

### DIFF
--- a/Sources/App/Models/Version.swift
+++ b/Sources/App/Models/Version.swift
@@ -91,7 +91,7 @@ final class Version: Model, Content {
     var url: String?
 
     @Field(key: "has_binary_targets")
-    var hasBinaryTargets: Bool?
+    var hasBinaryTargets: Bool
     
     // relationships
 
--- a/Sources/App/Models/Version.swift
+++ b/Sources/App/Models/Version.swift
@@ -90,6 +90,9 @@ final class Version: Model, Content {
     @Field(key: "url")
     var url: String?
 
+    @Field(key: "has_binary_targets")
+    var hasBinaryTargets: Bool?
+    
     // relationships
 
     @Children(for: \.$version)
--- a/Sources/App/Models/Version.swift
+++ b/Sources/App/Models/Version.swift
@@ -122,7 +122,8 @@ final class Version: Model, Content {
          supportedPlatforms: [Platform] = [],
          swiftVersions: [SwiftVersion] = [],
          toolsVersion: String? = nil,
-         url: String? = nil) throws {
+         url: String? = nil,
+         hasBinaryTargets: Bool = false) throws {
         self.id = id
         self.$package.id = try package.requireID()
         self.commit = commit
@@ -140,6 +141,7 @@ final class Version: Model, Content {
         self.swiftVersions = swiftVersions
         self.toolsVersion = toolsVersion
         self.url = url
+        self.hasBinaryTargets = hasBinaryTargets
     }
 
     enum Kind: String, Codable {
--- a/FrontEnd/styles/package.scss
+++ b/FrontEnd/styles/package.scss
@@ -100,7 +100,7 @@
         }
       }
 
-      li.binary {
+      li.has_binary_targets {
         grid-column-start: span 2;
         background-image: var(--image-warning);
       }
--- a/FrontEnd/styles/package.scss
+++ b/FrontEnd/styles/package.scss
@@ -100,6 +100,11 @@
         }
       }
 
+      li.binary {
+        grid-column-start: span 2;
+        background-image: var(--image-warning);
+      }
+
       li.stars {
         background-image: var(--image-star);
       }
--- a/Sources/App/Commands/Analyze.swift
+++ b/Sources/App/Commands/Analyze.swift
@@ -582,6 +582,7 @@ extension Analyze {
         version.supportedPlatforms = manifest.platforms?.compactMap(Platform.init(from:)) ?? []
         version.toolsVersion = manifest.toolsVersion?.version
         version.spiManifest = packageInfo.spiManifest
+        version.hasBinaryTargets = packageInfo.packageManifest.targets.contains { $0.type == .binaryTarget }
 
         return version.save(on: database)
     }
--- a/Sources/App/Commands/Analyze.swift
+++ b/Sources/App/Commands/Analyze.swift
@@ -582,7 +582,7 @@ extension Analyze {
         version.supportedPlatforms = manifest.platforms?.compactMap(Platform.init(from:)) ?? []
         version.toolsVersion = manifest.toolsVersion?.version
         version.spiManifest = packageInfo.spiManifest
-        version.hasBinaryTargets = packageInfo.packageManifest.targets.contains { $0.type == .binaryTarget }
+        version.hasBinaryTargets = packageInfo.packageManifest.targets.contains { $0.type == .binary }
 
         return version.save(on: database)
     }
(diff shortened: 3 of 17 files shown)


daveverwer:
Thanks so much, Sam! I've not had a full look at it yet but I will get in the visual side of this today.

Just one note for now, I could reproduce the test failures from CI locally. I'm getting 38 test failures locally, all around analyze/re-analyze.

daveverwer:
I tried running this locally and ran into an issue with `re-analyze-versions`. I migrated successfully and then wanted to test it with a package that I knew had binary targets, in this case "GoogleAppMeasurement" (id: 2CE8EA4D-2426-4B8D-8F6C-13C0723BAC8B).

`re-analyze-versions` ran successfully:

```
$ swift run Run re-analyze-versions --id 2CE8EA4D-2426-4B8D-8F6C-13C0723BAC8B 
Building for debugging...
Build complete! (0.28s)
[ INFO ] Re-analyzing versions (id: 2CE8EA4D-2426-4B8D-8F6C-13C0723BAC8B) ... [component: re-analyze-versions]
[ INFO ] Re-analyzing package https://github.com/google/GoogleAppMeasurement.git ... [component: re-analyze-versions]
[ INFO ] cloning https://github.com/google/GoogleAppMeasurement.git to /Users/dave/Developer/SwiftPackageIndex/Checkouts/github.com-google-googleappmeasurement [component: re-analyze-versions]
[ INFO ] Updating 25 versions (id: 2CE8EA4D-2426-4B8D-8F6C-13C0723BAC8B) ... [component: re-analyze-versions]
[ WARNING ] Environment variable not set: METRICS_PUSHGATEWAY_URL [component: re-analyze-versions]
[ INFO ] Done. [component: re-analyze-versions]
```

But after running, the `has_binary_targets`field` in versions still showed `FALSE` for all rows.

I'll manually update those rows in the database so i can check the front-end side of things, but I wanted to report the issue.

daveverwer:
I just pushed some experimentation with the wording. I want to make sure that we use the word "may" rather than being absolute. Packages may have ways to make the source code available through licensing deals or paid purchases, so we should avoid being absolute about it.

Original:

Screenshot 2022-07-25 at 12 51 03@2x

I expected to want to de-emphasise the orange warning, but once I looked at it, I thought it needed a little more. Maybe some bold. Like:

Screenshot 2022-07-25 at 12 49 11@2x

Screenshot 2022-07-25 at 12 50 14@2x

For comparison, this is how an [archived package](https://swiftpackageindex.com/NSHipster/swift-log-github-actions) looks.

Screenshot 2022-07-25 at 13 12 00@2x

My preference is:

Screenshot 2022-07-25 at 12 49 11@2x

But all opinions welcome!

daveverwer:
I've not updated snapshots yet until we're happy with the wording.

samdeane:
Is it possible that the analyse phase could be failing to decode the Package description?

I just ran into something with another project where I happen to also be decoding the target type of a swift package, where I hadn’t made the TargetType enum String-backed, and it didn’t decode property from the JSON package definition. I didn’t think that was necessary for a simple enum without associated values, but in my other project it seemed to be (I need to look into that)

samdeane:
> But all opinions welcome!

I agree with your preference - I think having the first part of the sentence bold is enough.

finestructure:
Hey Sam, thanks a lot for the PR! I’m travelling at the moment and it’s tricky to give this thorough review right away. I can try and help with debugging the test failures in the coming days.

> Is it possible that the analyse phase could be failing to decode the Package description?

That’s actually one thing I noticed when scanning the changes: it would be good to add/extend a decoding test of a manifest with the new properties to make sure the format is correct. We should have a json snippet in the fixtures directory in Tests that shows how to set that up.

samdeane:
@daveverwer If I arrange for Xcode to supply the arguments `re-analyze-versions --id 2CE8EA4D-2426-4B8D-8F6C-13C0723BAC8B`, can I debug the analysis process that way?

I tried to run the analysis with `make analyze` but it didn’t seem to do anything. I suspect I was doing something else wrong!

daveverwer:
> @daveverwer If I arrange for Xcode to supply the arguments `re-analyze-versions --id 2CE8EA4D-2426-4B8D-8F6C-13C0723BAC8B`, can I debug the analysis process that way?

Yes

samdeane:
@finestructure A test sounds like a good plan indeed.

Re the travelling: I’ve tagged @daveverwer as well, but either way, I understandm, and there’s no hurry.

samdeane:
> it would be good to add/extend a decoding test of a manifest with the new properties to make sure the format is correct.

It turns out that the new decoding broke a substantial minority of tests, precisely because the target type was not specified in the test JSON. I would say that fixing these has provided sufficient coverage for the new code!

samdeane:
> I believe that this is working now.

(obviously I realise that there may still be review comments, so I simply mean that I'm not aware of any outstanding tasks)

finestructure (approved):

daveverwer (approved):
Sven's review reminded me that while I had reviewed this, I had not left a tick!

This is great work, @samdeane. Thank you!

daveverwer:
There will be a non-trivial rebase based on the CSS changes I have made this weekend. I'll rebase this now and get it merged.

Sources/App/Models/Version.swift
```diff
@@ -90,6 +90,9 @@ final class Version: Model, Content {
     @Field(key: "url")
     var url: String?
 
+    @Field(key: "has_binary_targets")
+    var hasBinaryTargets: Bool
+
```
finestructure, line 95: Nit: we're not consistent with it everywhere but we try to keep properties within their "groups" (data fields, relationships) sorted alphabetically. I believe `Version` is one of the types where it's sorted.

Sources/App/Models/Version.swift
```diff
@@ -119,7 +122,8 @@ final class Version: Model, Content {
          supportedPlatforms: [Platform] = [],
          swiftVersions: [SwiftVersion] = [],
          toolsVersion: String? = nil,
-         url: String? = nil) throws {
+         url: String? = nil,
+         hasBinaryTargets: Bool = false) throws {
```
finestructure, line 126: dto re: sorting

Sources/App/Models/Version.swift
```diff
@@ -137,6 +141,7 @@ final class Version: Model, Content {
         self.swiftVersions = swiftVersions
         self.toolsVersion = toolsVersion
         self.url = url
+        self.hasBinaryTargets = hasBinaryTargets
```
finestructure, line 144: and here

daveverwer:
OK, I just merged `main` back into this (turns out I was right about the rebase being too hard 😂) and this should be good to go apart from Sven's nits.

Once merged, we'll need to co-ordinate the rollout of the onto production and run a re-analyze versions.

samdeane:
Ok, thanks for the review guys - I’ll try to sort out the nits shortly.

samdeane:
Have moved the new `hasBinaryTargets` property so that it’s in the right place, alphabetically. Hopefully should be good to merge I think.

daveverwer:
Just a note @samdeane to keep you up to date. We're planning to merge this tomorrow so we can run the re-analysis at the start of the day!

samdeane:
Thanks for keeping me in the loop! Don't worry though, I realise that it's not a trivial one to roll out.

I'm going to be travelling for the next little while, but I'm up for another issue if you think of something appropriate.

PR #1922, opened 2022-07-25
Cope better with badly encoded URLs in README files

### LINKED ISSUE #1917: Invalid URL 500s
There have been a number of invalid URL 500s starting Friday, July 22 morning:

- [Rollbar #75](https://rollbar.com/finestructure/SPI-server/items/75)

Abort.500: http://spi-prod-docs.s3-website.us-east-2.amazonaws.com/nordicsemiconductor/ios-dfu-library/raw/main/App/fastlane/screenshots/en-US/iPhone 8 Plus-01Devices.png is an invalid URL

- [Rollbar #76](https://rollbar.com/finestructure/SPI-server/items/76)

Abort.500: http://spi-prod-docs.s3-website.us-east-2.amazonaws.com/nordicsemiconductor/ios-dfu-library/raw/main/App/fastlane/screenshots/en-US/iPhone 8 Plus-02Settings.png is an invalid URL

- [Rollbar #77](https://rollbar.com/finestructure/SPI-server/items/77)

Abort.500: http://spi-prod-docs.s3-website.us-east-2.amazonaws.com/airbnb/lottie-ios/raw/master/_Gifs/Community 2_3.gif is an invalid URL

- [Rollbar #78](https://rollbar.com/finestructure/SPI-server/items/78)

Abort.500: http://spi-prod-docs.s3-website.us-east-2.amazonaws.com/tuist/tuist/raw/main/assets/companies/Depop Logo.svg is an invalid URL

- [Rollbar #79](https://rollbar.com/finestructure/SPI-server/items/79)

Abort.500: http://spi-prod-docs.s3-website.us-east-2.amazonaws.com/tuist/tuist/raw/main/assets/companies/hh mono.svg is an invalid URL

- [Rollbar #80](https://rollbar.com/finestructure/SPI-server/items/80)

Abort.500: http://spi-prod-docs.s3-website.us-east-2.amazonaws.com/tuist/tuist/raw/main/assets/companies/Compass Black Logo.png is an invalid URL

These all seem to be related to lack of quoting of the requested url.

daveverwer:
Fixes #1917

Alright. The fix to #1917 is tricky and I could use a second opinion on the best way to fix it. I originally assumed *all* relative URLs were broken but this is not the case. It’s an issue with unencoded characters in URLs.

For example:

```
https://example.com/encoded%20url.png
/encoded%20url.png
```

Both render fine as the URLs are correctly encoded. Whereas:

```
https://example.com/unencoded url.png
/encoded url.png
```
Both break as these are not valid URLs as spaces are not allowed. This would also break for other characters that are not allowed, like `<`, `{`, etc… 

So, that’s the problem defined.

---

My thought on how to fix this was to make a new initialiser on `URL` that coped with this situation. `URL(string:)` returns nil for the unencoded URLs (because they are not URLs) and I thought that could be used to work around the issue. This draft PR contains an implementation based on that.

However, while it works well for relative URLs where *all* invalid characters are either encoded or unencoded. It does not work at all for URLs that have partially encoded URLs, or *any* absolute URL.

Take the example of initialising with an absolute URL:

```
https://full.host/unencoded spaces
```

That fails initialisation with `URL(string:)` because it’s not a valid URL, so the code attempts to encode the input, which now includes a valid `:` between the protocol and the host, meaning it outputs:

```
https%3A//full.host/unencoded%20spaces
```

I’m not sure how to fix this without writing a URL parser, which we absolutely should not do.

### DIFF
--- a/Sources/App/Views/PackageController/PackageReadme+Model.swift
+++ b/Sources/App/Views/PackageController/PackageReadme+Model.swift
@@ -61,7 +61,7 @@ extension PackageReadme {
             do {
                 let imageElements = try element.select("img")
                 for imageElement in imageElements {
-                    guard let imageUrl = URL(string: try imageElement.attr("src"))
+                    guard let imageUrl = URL(withPotentiallyPercentEncodedPath: try imageElement.attr("src"))
                     else { continue }
 
                     // Assume all images are relative to GitHub as that's the only current source for README data.
--- a/Sources/App/Views/PackageController/PackageReadme+Model.swift
+++ b/Sources/App/Views/PackageController/PackageReadme+Model.swift
@@ -61,7 +61,7 @@ extension PackageReadme {
             do {
                 let imageElements = try element.select("img")
                 for imageElement in imageElements {
-                    guard let imageUrl = URL(withPotentiallyPercentEncodedPath: try imageElement.attr("src"))
+                    guard let imageUrl = URL(withPotentiallyUnencodedPath: try imageElement.attr("src"))
                     else { continue }
 
                     // Assume all images are relative to GitHub as that's the only current source for README data.
@@ -82,7 +82,7 @@ extension PackageReadme {
             do {
                 let linkElements = try element.select("a")
                 for linkElement in linkElements {
-                    guard let linkUrl = URL(withPotentiallyPercentEncodedPath: try linkElement.attr("href"))
+                    guard let linkUrl = URL(withPotentiallyUnencodedPath: try linkElement.attr("href"))
                     else { continue }
 
                     // Assume all links are relative to GitHub as that's the only current source for README data.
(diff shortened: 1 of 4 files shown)


finestructure:
IIRC there’s an encoding that only touches characters that aren’t part of the domain, i.e. it’d leave : and / alone. If that’s the case we could run it on any urls that don’t contain a % (as an indicator if they’re already encoded).

daveverwer:
> `addingPercentEncoding(withAllowedCharacters: .urlQueryAllowed)` or something

That's interesting, and works for all cases apart from the partially encoded one:

```
/partially%20{encoded}
```

I'm happy to allow URLs that are *that* badly malformed give us an occasional error. It should be extremely rare.

finestructure (approved):

daveverwer:
Deploying as 2.73.8.

PR #1927, opened 2022-07-28
Documentation pages header shrink/redesign

### LINKED ISSUE #1865: Packages with many documentation targets/archives break the DocC header
[GatheredKit](https://swiftpackageindex.com/JosephDuffy/GatheredKit/main/documentation/gatheredkit)

Screenshot 2022-07-01 at 17 21 14@2x

[Epoxy](https://swiftpackageindex.com/airbnb/epoxy-ios/master/documentation/epoxybars)

Screenshot 2022-07-01 at 17 21 45@2x

### LINKED ISSUE #1907: Reduce the size of the SPI documentation header
With potentially five 😱 levels of "header" on our hosted docs, it's grown out of control quite fast. There are header rows for:

* SPI Nav
* Breadcrumb
* Latest stable warning
* Different documentation archives
* The DocC header (which we can't/shouldn't touch)

Screenshot 2022-07-19 at 17 17 54@2x

We could reduce this quite significantly with a couple of changes:

- [ ] Merge the SPI Nav with the breadcrumb. A smaller logo to the left of the "Home" Breadcrumb would potentially get the message across and remove the biggest band completely. People can go home to find package search friends.
- [ ] Merge the doc archive switcher to be another segment in the breadcrumb. Now we have dropdown breadcrumbs, it could be: Home > Org > Package > Reference (drop-down) > Archive (drop-down)

### LINKED ISSUE #1914: Version switcher should not appear if there is only one version!
Screenshot 2022-07-20 at 18 41 36@2x

daveverwer:
Fixes #1865
Fixes #1907
Fixes #1914

The new header is *much* smaller.

I compacted the main top header with the logo and menu into the breadcrumb, which fixes #1907. I also integrated the multiple archive links into the breadcrumb, too, which fixes #1865. Finally, I introduced selected states on the breadcrumb dropdown menus, fixing #1914.

Lots of screenshots are incoming! Discussion on the design is welcome.

### The simplest possible case. One version and one set of documentation archives:

The dropdown on the version still shows, but it also shows that the default branch is selected. Also, note that no warning about being on the default branch is visible as there is no stable branch with documentation. I decided not to even show the `doc_archive` breadcrumb if there is only one. It looked weird with an "extra" `LeftPad` as a final breadcrumb.

Screenshot 2022-07-28 at 21 42 12@2x

Screenshot 2022-07-28 at 21 43 04@2x

### Multiple versions of documentation with the stable branch selected

Screenshot 2022-07-28 at 21 49 08@2x

### Multiple versions of documentation with the default branch selected

Here's where the warning comes in. There is stable documentation, so people should know about that.

Screenshot 2022-07-28 at 21 51 06@2x

### Multiple versions, multiple archives

Screenshot 2022-07-28 at 21 52 13@2x

Screenshot 2022-07-28 at 21 52 17@2x

### LOTS of archives

Screenshot 2022-07-28 at 21 53 30@2x


daveverwer:
I approved the Percy build, but someone else should take a quick look, too. There should only be two snapshots that have changed (including one renamed snapshot, which is why it says there's a missing snapshot). However it has marked *lots* of snapshots as changed, but with no visual diffs that I can see.

Weird

finestructure:
I'm not sure how to interpret this but it's not really flagging any diffs. I see two "new" snapshots (first one and one towards the end), lots of "reintroduced" snapshots like this:

Screenshot 2022-07-29 at 17 32 43

and one actual diff - which I guess is the change?

Screenshot 2022-07-29 at 17 36 34

daveverwer:
> I'm not sure how to interpret this but it's not really flagging any diffs. I see two "new" snapshots (first one and one towards the end), lots of "reintroduced" snapshots

Yea, I'm not sure what's going on. It has never done this before.

> and one actual diff - which I guess is the change?

There are two *real* diffs, but I also renamed one of the tests (it was incorrect) and so maybe my rename messed this up.

finestructure:
As an aside, should we drop the "Switch to" from the drop down? It's pretty clear how drop downs work, isn't it?

finestructure:
Is there a way to force percy to re-compare? Maybe that shakes it lose.

daveverwer:
> Is there a way to force percy to re-compare? Maybe that shakes it lose.

I just had a quick look and I don't think so. I think it'll be fine. Once this is merged, it will adjust the base screenshots to match these.

Sources/App/routes.swift
```diff
@@ -61,7 +61,10 @@ func routes(_ app: Application) throws {
             app.get(":owner", ":repository", ":reference", "documentation") {
                 try await packageController.documentation(req: $0, fragment: .documentation)
             }
-            app.get(":owner", ":repository", ":reference", "documentation", "**") {
```
finestructure, line 64: I wonder if we're missing any routes if we drop this, although I may have added this for sake of completeness more than anything.

finestructure (approved):

Sources/App/routes.swift
daveverwer: No. This one is important if people land deeper into the documentation.

Eg. `/owner/repo/ref/documentation/symbol/method/something_else`

finestructure: Well, should we be removing it then? :) This is getting deleted in this PR!

daveverwer: Deleted?

finestructure: Yes, this route as is is being removed isn't it?

Screenshot 2022-07-29 at 17 55 21

finestructure: (the others might be 1:1 replacements but I'm not 100% sure if they are, i.e. how the `**` actually captures and deals with precedence etc)

daveverwer: Ah, I see. Not really. I needed to extract the first part of the fragment out into a named parameter `:archive`, so this route is now basically the one on L67.

daveverwer:
> As an aside, should we drop the "Switch to" from the drop down? It's pretty clear how drop downs work, isn't it?

Removed

Screenshot 2022-07-31 at 10 16 23@2x

PR #1933, opened 2022-07-31
Keyword search tweak

finestructure:
Rebased version of #1905, to kick off CI and have an Xcode PR view open.

To do:

- [x] Review/merge review changes #1955
- [ ] Re-run/review query perf tests after merging #1955


Sources/App/Core/Search.swift
```diff
@@ -286,8 +319,14 @@ enum Search {
         select = select
             .column(SQLFunction("LEVENSHTEIN", args: repoOwner, SQLBind(mergedTerms)),
                     as: levenshteinDist)
+        select = select
+            .column(ts_rank(vector: tsvector, query: tsquery), as: tsrankvalue)
```
finestructure, line 323: I believe this only compiles because `tsvector` and `tsquery` are defined as a constants at `Search` level. We'll need to fill these in.

Sources/App/Core/Search.swift
```diff
@@ -38,6 +38,12 @@ enum Search {
     static let summary = SQLIdentifier("summary")
     static let hasDocs = SQLIdentifier("has_docs")
 
+    static let tsquery = SQLIdentifier("tsquery")
+    static let tsvector = SQLIdentifier("tsvector")
+    static let tsrankvalue = SQLIdentifier("tsrankvalue")
+    
+    static let emptyString = SQLRaw("''")
```
finestructure, line 45: I believe `SQLRaw("")` should suffice.

Sources/App/Core/Search.swift
finestructure: Never mind, I missed the `from` clauses just below.

However, we'll need to keep in mind that multiple `from` clauses will affect the cardinality

finestructure: This is already the case in `keywordMatchQueryBuilder` where we currently unnest the keywords (live version). 

That's potentially an existing bug. Not in the sense that it'll give wrong results I believe but I suspect it selects N(search view rows) x (N keywords) rows without a join and filters on the product when we really want to filter just on keywords.

heckj:
The current setup fails the performance test, but only but a small amount. 

Based on what I found while digging around in pgAdmin and the explanation of the query, the FROM terms that build up the ts_vectors get inlined into the flow, and don't add to the cardinality of the query - but I think we could pre-load that computation into the materialized search view (by adding a migration that builds the vectors for each row), and I was fully expecting we might want to do that prior to merging this.

As a general pattern, I tried to "keep and extend" all the existing queries, enhancing them with the ranking rather than replacing any of the functionality. Figured that would be the best proof of concept path, and we could talk through the existing query structure to refine anything we want to clean up, including the potential bug you highlighted above with the keyword queries.

For what it's worth, all of the keywords (today/currently) are in English - which makes them particularly amenable to the tsvector normalization and ranking process, and while I haven't measured the same metrics for the keyword results (precision, recall, etc) - they're noticeable better with a mechanism using the `tsquery @@ tsvector` WHERE comparison. There wasn't really an impact to adding ranking to that set though (the keyword query) - it doesn't flow to impacting the end result search query all that much.

Sources/App/Core/Search.swift
```diff
@@ -175,7 +183,18 @@ enum Search {
             .column(keywords)
             .column(hasDocs)
             .column(nullInt, as: levenshteinDist)
+            .column(ts_rank(vector: tsvector, query: tsquery), as: tsrankvalue)
             .from(searchView)
+            .from(plainto_tsquery(mergedTerms), as: tsquery)
```
finestructure, line 188: I've checked and neither additional "FROM" clause poses a problem WRT cardinatily, they're both (1).

They could actually just be select clauses. Is there reason you placed them in FROM so they're easier to use because the alternative would be to write out a bigger `ts_rank` clause that has them both inline?

If we pull "tsvector" into the search view it might make sense to do all this inline.

Sources/App/Core/Search.swift
```diff
@@ -175,7 +183,18 @@ enum Search {
...
+            .from(
+                TextSearch.toVector(
+                    concat(with: " ",
+                        coalesce(packageName, emptyString),
+                        coalesce(summary, emptyString),
+                        arrayToString(keywords, delimiter: " ")
+                    ),
+                    weight: .a
```
finestructure, line 196: Actually, I don't think we need to use `setweight` at all. We're only applying weights A and B in distinct subqueries which don't actually mix. I.e. we weight the vector in the package (1) and the author (2) subqueries with A and the keywords (3) subquery with B.

However, these results aren't actually displayed in the same list and so their respective weightings have no effect:

CleanShot 2022-08-07 at 16 56 01@2x

It does not harm but we should remove that to simplify the SQL a bit to keep it a little more readable.

Sources/App/Core/Search.swift
```diff
@@ -230,14 +249,29 @@ enum Search {
...
+            // the ts_vector (normalized english lexemes) representations, we can do so like:
+            //
+            //.where(to_tsquery(SQLBind(mergedTerms)), inVector, to_tsvector(keyword))
+            //
+            // Negative side effect is that terms prefixed with UI and NS (such as NSBezierCurve)
+            // won't be identified for the search term `bezier`, and non-english search terms
+            // (such as mandarin characters) completely fail to find anything.
+        select = select
+            //.orderBy(levenshteinDist)
+            .orderBy(tsrankvalue)
```
finestructure, line 274: I believe our ordering here is actually redundant, because we're sorting (client side) by the keyword frequency (descending).

We can remove both the levenshteinDist *and* the tsvector work here.

finestructure: Just another note: If you're searching for `ftp` you probably want that match on top:

CleanShot 2022-08-07 at 17 23 32@2x

However, we can't `ORDER BY "tsrank", "frequency"`, because we don't have `frequency` available in that query. We do the sorting in Swift after fetching (what I called "client side" above, in the sense of "client of the db").

I think it's out of scope to try and address this and it's certainly not a regression introduced by this PR.

finestructure:
I suspect there's not much to do WRT performance tuning after we merge #1955. Some of the deltas should go away and the explain analyze I ran on the package subquery should only very, very minor changes. It's probably not even worth touching the materialised view for that - but let's see once we've merged #1955.

Sources/App/Core/Search.swift
heckj: We don't need the weights. Originally I thought I'd be trying to mix different weights, but wasn't able to work out how to do so outside of the SELECT inline setup. End result - didn't end up using it. If we care about the specific rank values (for example, to fuse in Score as a modifier), then it might be worthwhile, but for this round I think we can absolutely strip it out.

finestructure: I'm pretty sure we can fix the joining issue you encountered. Are there some cases we know would benefit from using weights? Might be something to schedule for a follow-up. It's definitely valuable to have had it around so we can pull it from git history should we need it!

heckj: Since I didn't get it working, I don't have any hard numbers for what benefit it could provide. My gut was telling me to rank keywords and package titles above anything in the summary content based on past experiences, but I don't know that we'd see much benefit from even doing that right now.

finestructure:
I'm puzzled by why percy is flagging visual changes for this PR. There are no html changes. The diff in percy shows something in the footer which looks like aliasing issues. Any idea what this is about, @daveverwer ?

finestructure:
Ok, rebase fixed that. I guess there was interference due to changes on `main` which we didn't have in this branch yet.

PR #1945, opened 2022-08-03
Redirect to the first available documentation archive if no archive is supplied on the URL

### LINKED ISSUE #1924: Documentation path without an archive should redirect to the first archive or 404 if no archives exist
There's a test in PackageController+routesTests.swift:248 which is passing right now, but the experience of that URL is really bad as it shows this when serving `/owner/package/ref/documentation`:

Screenshot 2022-07-28 at 21 04 28@2x

We know that either:

1. The package has no documentation archives. It should 404 with a regular 404 (not the documentation one)
2. The package has at least one documentation archive, we should redirect to the first one.

daveverwer:
Fixes #1924

Now, adding a reference and /documentation on any package URL will redirect to the first docArchive if one exists.

### DIFF
--- a/Sources/App/Controllers/PackageController+routes.swift
+++ b/Sources/App/Controllers/PackageController+routes.swift
@@ -105,7 +105,9 @@ struct PackageController {
 
         let url = try Self.awsDocumentationURL(owner: owner, repository: repository, reference: reference, fragment: fragment, path: path)
         let awsResponse = try await Current.fetchDocumentation(req.client, url)
-        guard (200..<399).contains(awsResponse.status.code) else {
+
+        // Never let a request continue if the AWS request fails, except to a potential `/owner/repo/ref/documentation/` redirect.
+        guard (200..<399).contains(awsResponse.status.code) || fragment == .documentationRedirect else {
             // Convert anything that isn't a 2xx or 3xx from AWS into a 404 from us.
             throw Abort(.notFound)
         }
--- a/Sources/App/Controllers/PackageController+routes.swift
+++ b/Sources/App/Controllers/PackageController+routes.swift
@@ -142,22 +142,6 @@ struct PackageController {
                                                                                              reference: reference,
                                                                                              docArchive: docArchive.name))
 
-// Possible solution if the `first`-based one doesn't work
-//            case .documentationRedirect:
-//                let queryResult = try await Joined3<Version, Package, Repository>
-//                    .query(on: req.db,
-//                           join: \Version.$package.$id == \Package.$id, method: .inner,
-//                           join: \Package.$id == \Repository.$package.$id, method: .inner)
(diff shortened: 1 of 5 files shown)


Sources/App/Controllers/PackageController+routes.swift
```diff
@@ -104,18 +105,59 @@ struct PackageController {
...
+                    }
+                }()
+
+                guard let queryResult = try await Joined3<Version, Package, Repository>
+                    .query(on: req.db,
+                           join: \Version.$package.$id == \Package.$id, method: .inner,
+                           join: \Package.$id == \Repository.$package.$id, method: .inner)
+                    .filter(Repository.self, \.$owner, .custom("ilike"), owner)
+                    .filter(Repository.self, \.$name, .custom("ilike"), repository)
+                    .filter(\Version.$reference == referenceToMatch)
```
daveverwer, line 131: Is there a way I'm missing to do a JSON comparison here, @finestructure? The SQL that this outputs:

```
SELECT "versions"."doc_archives" AS "versions_doc_archives" FROM "versions" INNER JOIN "packages" ON "versions"."package_id" = "packages"."id" INNER JOIN "repositories" ON "packages"."id" = "repositories"."package_id" WHERE "repositories"."owner" ilike $1 AND "repositories"."name" ilike $2 AND "versions"."reference" = $3 AND "versions"."doc_archives" IS NOT NULL LIMIT 1 [Optional("owner"), Optional("package"), main]
```

So the `reference` where clause is simply comparing against `'main'`, but it's a JSON field.

finestructure: I'm pretty sure that's just rendering as `main` in the dump because of `Reference`'s `CustomStringConvertible` conformance. You're really comparing a `JSONB` field to a structured type `Reference`, which should work as expected. I believe we're doing this elsewhere as well.

finestructure (approved):

Sources/App/Controllers/PackageController+routes.swift
```diff
@@ -347,7 +379,7 @@ extension PackageController {
         let baseURL = "http://\(baseURLHost)/\(baseURLPath)"
 
         switch fragment {
-            case .css, .data, .documentation, .images, .index, .js:
+            case .css, .data, .documentationRedirect, .documentation, .images, .index, .js:
                 return URI(string: "\(baseURL)/\(fragment)/\(path)")
```
finestructure, line 383: I'm not sure this is actually working as intended. This is generating the following urls in AWS which would definitely not be working:

```
(lldb) po URI(string: "\(baseURL)/\(fragment)/\(path)")
▿ http://awsDocsBucket.s3-website.us-east-2.amazonaws.com/owner/package/1.2.3/documentationRedirect/
  - string : "http://awsDocsBucket.s3-website.us-east-2.amazonaws.com/owner/package/1.2.3/documentationRedirect/"
```

PR #1946, opened 2022-08-03
Gate the has binary targets line to dev/staging only for now.

daveverwer:
More info coming but this isn't quite working as expected now it has run on staging.


daveverwer:
Actually, don't merge this yet. I may be wrong about there being a problem!

daveverwer:
Yep. There is an issue. Ready for review @finestructure

finestructure (approved):

PR #1953, opened 2022-08-05
Fix Footer Link Padding

Sherlouk:
Noticed that the thank you sponsors link in the footer had extra padding which makes it sit weirdly in the sentence (looks like it has extra spaces on either side).

This fixes it, in theory, by only adding padding to the list item links (the navigation bar items).

Edited via GitHub, untested. Can verify once I'm on my laptop but this should only impact the "all our sponsors for their generosity" link and fix the extra padding before and after.

### DIFF
--- a/FrontEnd/styles/header_footer.scss
+++ b/FrontEnd/styles/header_footer.scss
@@ -43,7 +43,7 @@ footer {
             li {
                 margin: 0 5px;
                 list-style: none;
-                
+
                 a {
                     padding: 3px;
                 }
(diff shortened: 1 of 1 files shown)


daveverwer:
Thanks James! I'll test, lint, and merge this later today.

daveverwer (approved):
Thanks, James!

daveverwer:
From looking at the snapshots, this actually fixed a very small misalignment with the site title/logo, too!

Sherlouk:
Ahh that was totally my intention 👀😂

Thanks for picking that up Dave, and sorry for the half ass job finishing/testing. 

It just randomly caught my eye this morning and wanted to make a reminder for myself aha

PR #1954, opened 2022-08-05
Tweak setweight parameter type

finestructure:
Make the weight a parameter on `to_tsvector`

### DIFF
--- a/Sources/App/Core/Extensions/SQLKit+functions.swift
+++ b/Sources/App/Core/Extensions/SQLKit+functions.swift
@@ -54,40 +54,12 @@ func unnest(_ array: SQLExpression) -> SQLFunction {
     SQLFunction("UNNEST", args: array)
 }
 
-
 func to_tsvector(_ array: SQLExpression) -> SQLFunction {
     // The argument is meant to be a string, which this wraps and converts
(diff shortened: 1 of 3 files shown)


Sources/App/Core/Extensions/SQLKit+functions.swift
```diff
@@ -49,44 +49,46 @@ func lower(_ arg: SQLExpression) -> SQLFunction {
     SQLFunction("LOWER", args: arg)
 }
 
+
 func unnest(_ array: SQLExpression) -> SQLFunction {
     SQLFunction("UNNEST", args: array)
 }
 
+
 func to_tsvector(_ array: SQLExpression) -> SQLFunction {
```
finestructure, line 58: Actually, we should probably fully "swiftify" this and make it `toTSVector` or `toTextSearchVector` or something.

finestructure: Although there's an argument to be made to keep it aligned with the underlying SQL function. OTOH `to_tsvector` with a weight which wraps a `to_tsvector` in `setweight` kind of breaks that convention already.

heckj (approved):
works for me!

PR #1955, opened 2022-08-07
Review changes

finestructure:
This addresses three of my review comments

- [Move tsvector into select clause](https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/pull/1933/files#r939672817)
- [Remove redundant setweight](https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/pull/1933/files#r939680900)
- [Remove redundant keyword order by](https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/pull/1933/files#r939683333)


finestructure:
This should also address some of the query performance failures.

heckj (approved):
Looks good, reducing the complexity and pushing as much inline as possible makes a lot of sense.

Sources/App/Core/Search.swift
```diff
@@ -183,18 +179,21 @@ enum Search {
...
+                            arrayToString(keywords, delimiter: " ")
+                        )
                     ),
-                    weight: .a
-                ), as: tsvector)
+                    query: plainto_tsquery(mergedTerms)
+                ),
+                as: tsrank
+            )
+            .from(searchView)
```
finestructure, line 196: I need to review the query plans in detail to confirm but at first glance this change is a huge regression in query performance. Joe's initial query was

```
ℹ️ TEST:        test_Search_packageMatchQuery()
ℹ️ COST:        854.52
ℹ️ EXPECTED:    670.0 ± 50.0
ℹ️ ACTUAL TIME: 259.785ms
```

My change is

```
ℹ️ TEST:        test_Search_packageMatchQuery()
ℹ️ COST:        3372.92
ℹ️ EXPECTED:    670.0 ± 50.0
ℹ️ ACTUAL TIME: 421.16ms
```

Looking at the actual time I'm also concerned. When I checked this yesterday I thought it looked ok but I may have run a slightly different query or been mistaken about the numbers. Needs re-evaluating.

On current main we have:

```
ℹ️ TEST:        test_Search_packageMatchQuery()
ℹ️ COST:        681.65
ℹ️ EXPECTED:    670.0 ± 50.0
ℹ️ ACTUAL TIME: 46.49ms
```

So in terms of actual time that's 5-10x faster.

finestructure: I definitely need to re-run this against staging. Locally, I'm not seeing this regression when manually running against the package match subquery itself.

The three queries have the costs

- [308](https://explain.dalibo.com/plan/o17)
- [309](https://explain.dalibo.com/plan/ybo)
- [327](https://explain.dalibo.com/plan/VSo)

and execution times in the range 15ms-30ms.

finestructure: Ok, this is interesting. Our query performance tests is actually exposing an interesting aspect of `plainto_tsquery` in that it searches simply for `a`. When I run the query with the values substituted in:

```
SELECT 'package' AS "match_type", NULL AS "keyword", "package_id", "package_name", "repo_name", "repo_owner", "score", "summary", "stars", "license", "last_commit_date", "last_activity_at", "keywords", "has_docs", NULL::INT AS "levenshtein_dist", ts_rank(to_tsvector(CONCAT_WS(' ', COALESCE("package_name", ''), COALESCE("summary", ''), ARRAY_TO_STRING("keywords", ' '))), plainto_tsquery('a')) AS "tsrank" 
FROM "search" 
WHERE CONCAT_WS(' ', "package_name", COALESCE("summary", ''), "repo_name", "repo_owner", ARRAY_TO_STRING("keywords", ' ')) ~* 'a' AND "repo_owner" IS NOT NULL AND "repo_name" IS NOT NULL ORDER BY LOWER("package_name") = 'a' DESC, "tsrank" DESC, "score" DESC, "package_name" ASC
```

I get loads of warnings:

```
NOTICE:  text-search query contains only stop words or doesn't contain lexemes, ignored
```

CleanShot 2022-08-08 at 17 40 53@2x

These seems to be repeated once per line. When I run the same query with Joe's variant where `plainto_tsquery` appears in the `FROM` clause, I only get this message once.

I suspect that by moving it from `FROM` to `SELECT` this notice is triggered once per row instead of just once due to the way the results are processed.

When I replace `a` with `ping` as the search term, the perf test gives us the same results I'm getting when running it manually.

It's hard to tell if this is an artefact that's only impacting interactive sessions and/or `explain analyze` but in the absence of knowing exactly, I don't think it makes sense to pull `plainto_tsquery` into the select like that.

heckj: At first I was surprised, but then it made sense that searches for `a`, `the`, `to`, and other terms that would otherwise be considered stop words to a normalized full-text search have such a negative effect. I'd hoped the ranking wouldn't be so brutal, but yeah - that's a definite search-time regression. I suspect if you'd happened to use `b` as the search term, it wouldn't have been so brutal. ("B" isn't a common stop word)

heckj: My sense is that if we want to hoist something into the materialized view, pulling up the calculated ts_vector (and any weighting) would make a lot of sense.

finestructure: I should clarify: in your original version this is not really a problem, because the nature of having the `plainto_tsquery` in the `FROM` clause triggers this notice only once (even in this absolutely pathological case). My variant makes this much worse, because PG seems to then trigger the notice for each row.

I'm rolling back my change of inlining the `FROM` clauses into `SELECT`.

I don't think the materialised view would help here, because the notice is raised by the dynamic part in `plainto_tsquery`, which we can't push into the view. We can only push the static `to_tsvector` bit. But before considering that I want to measure things against with the latest changes.

finestructure:
Closing in favour of #1960

PR #1960, opened 2022-08-09
Reworked review changes

finestructure:
Supersedes #1955 

This reverts back to Joe's original queries (additional `FROM` clauses), which a few minor tweaks:

- remove additional `FROM` clauses and ordering from `keyword` search (because we're not actually using the sort order and can therefore remove ts_vector-based clauses and even the existing levenshtein-based ordering)
- remove ts_vector-based clauses from `author` search because we were actually still using levenshtein for ordering (which we could of course change)
- remove `setweight` (picked from #1955)
- a few other small bits picked from #1955 

The main difference to #1955 is that the queries are back to what Joe had minus `setweight` and _only_ affecting the package search.

We now have only one perf regression to deal with, `test_Search_packageMatchQuery`:

```
test_Search_packageMatchQuery(): failed - Total cost of 854.51 above threshold of 720.0 (incl variation)

Query plan:

Sort  (cost=841.26..854.51 rows=5303 width=296) (actual time=278.094..284.010 rows=4749 loops=1)
  Sort Key: ((lower(search.package_name) = 'a'::text)) DESC, (ts_rank(tsvector.tsvector, tsquery.tsquery)) DESC, search.score DESC, search.package_name
  Sort Method: quicksort  Memory: 2109kB
  ->  Nested Loop  (cost=0.51..513.20 rows=5303 width=296) (actual time=1.147..222.525 rows=4749 loops=1)
        ->  Function Scan on plainto_tsquery tsquery  (cost=0.25..0.26 rows=1 width=32) (actual time=1.024..1.028 rows=1 loops=1)
        ->  Nested Loop  (cost=0.26..420.13 rows=5303 width=255) (actual time=0.113..207.337 rows=4749 loops=1)
              ->  Seq Scan on search  (cost=0.00..313.82 rows=5303 width=223) (actual time=0.047..18.250 rows=4749 loops=1)
                    Filter: ((repo_owner IS NOT NULL) AND (repo_name IS NOT NULL) AND (concat_ws(' '::text, package_name, COALESCE(summary, ''::text), repo_name, repo_owner, array_to_string(keywords, ' '::text)) ~* 'a'::text))
                    Rows Removed by Filter: 121
              ->  Function Scan on to_tsvector tsvector  (cost=0.26..0.27 rows=1 width=32) (actual time=0.029..0.030 rows=1 loops=4749)
Planning Time: 0.771 ms
Execution Time: 290.612 ms
```

That increase in execution time is significant vs `main`:

```
ℹ️ TEST:        test_Search_packageMatchQuery()
ℹ️ COST:        681.65
ℹ️ EXPECTED:    670.0 ± 50.0
ℹ️ ACTUAL TIME: 48.504ms
```

i.e. x6. Note that this is likely due to the pathological search term `a`. That's something to look at more closely and fix either by adjusting what we search for or in some other way, but it's something we can do on the new search branch rather than in this PR here.


heckj (approved):
Looks good!

PR #1961, opened 2022-08-09
Fix documentation header for outdated versions of documentation

### LINKED ISSUE #1950: Handle links to out of date versions of documentation
When a documentation version is out of date (no longer latest stable) there are some issues with it, presumably as the query fails in some way. It looks like this:

Screenshot 2022-08-03 at 21 16 25@2x

Problems:

- [ ] No icon next to the version number (this is caused by the lack of a CSS class driven by the DB query)
- [ ] No sentence before "View latest stable docs"

daveverwer:
Fixes #1950 

Screenshot 2022-08-09 at 22 12 22@2x


Sources/App/Controllers/PackageController+routes.swift
```diff
@@ -87,6 +87,7 @@ struct PackageController {
         var packageName: String
         var docArchives: [String]
         var latest: Version.Kind?
+        var kind: Version.Kind
```
finestructure, line 90: It doesn't seem necessary to make this a property - we already have `reference` on this type and can get `reference.versionKind` from it (the same that's assigned into the value). Is there a reason we need this I'm missing?

I've removed the property and simply set `DocumentationPageProcessor(... referenceKind: documentation.reference.versionKind, ...)` and that seems to work fine?

Sources/App/Models/Reference.swift
```diff
@@ -56,6 +56,14 @@ enum Reference: Equatable, Hashable {
                 return tagName
         }
     }
+
+    var versionKind: Version.Kind {
+        if let semVer = semVer {
+            return semVer.isPreRelease ? .preRelease : .release
+        } else {
+            return .defaultBranch
+        }
```
finestructure, line 65:
```suggestion
        switch self {
            case .branch:
                return .defaultBranch
            case .tag(let semVer, _):
                return semVer.isPreRelease ? .preRelease : .release
        }
```

I think that's a little clearer to the reader that this unwraps the cases, rather than going through the `semVer` accessor indirectly.

Sources/App/Views/DocumentationPageProcessor.swift
```diff
@@ -23,7 +23,8 @@ struct DocumentationPageProcessor {
     let repositoryName: String
     let packageName: String
     let reference: String
-    let referenceKind: Version.Kind?
+    let referenceLatest: Version.Kind?
+    let referenceKind: Version.Kind
```
finestructure, line 27: It might make sense to make `reference` of type `Reference` here and drop `referenceKind` because it can be accessed through `reference.versionKind` like above.

Sources/App/Views/DocumentationPageProcessor.swift
```diff
@@ -151,7 +154,7 @@ struct DocumentationPageProcessor {
                         )
                     )
```
finestructure, line 155: There's an `unwrap` of `referenceKind` in this file (because `referenceKind` used to be optional) that should be redundant now (line 117, I can't reference it directly because it's not new in the diff).

Sources/App/Views/DocumentationPageProcessor.swift
finestructure: Actually, that's probably not practical since `reference` here is from the input parameter `req.parameters.get("reference")`. While we _could_ construct a `Reference` from that it's probably not a good idea to do that here.

finestructure (approved):

PR #1968, opened 2022-08-12
Route test

finestructure:
This adds a `route-tests.restfile` testing all GET routes (generated via `swift run Run show-routes`).

The tests all pass when running against a local server pointing to S3 `spi-dev-docs`. There are some server-specific bits of data in the test routes that are hard to generalise. I.e. `route-tests.restfile` is always going to be env/server-specific to a degree. For instance, we won't have a build-UUID that we can test against that exist on all envs.

The purpose of this test file is to have an easy way to ensure all urls are working as expected when transitioning to pointfreeco's URL parser/printer library.

This test also surfaced a number of issues in the doccy-proxy routes:
- the `images` route was misspelled, it needed to be `img`
- there are a number of directory routes that seem superfluous (i.e. they can't actually be tested)

### DIFF
--- a/Sources/App/routes.swift
+++ b/Sources/App/routes.swift
@@ -82,11 +82,11 @@ func routes(_ app: Application) throws {
             app.get(":owner", ":repository", ":reference", "data", "**") {
                 try await packageController.documentation(req: $0, fragment: .data)
             }
-            app.get(":owner", ":repository", ":reference", "images") {
-                try await packageController.documentation(req: $0, fragment: .images)
+            app.get(":owner", ":repository", ":reference", "img") {
+                try await packageController.documentation(req: $0, fragment: .img)
             }
-            app.get(":owner", ":repository", ":reference", "images", "**") {
-                try await packageController.documentation(req: $0, fragment: .images)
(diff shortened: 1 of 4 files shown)


daveverwer (approved):

Sources/App/routes.swift
```diff
@@ -70,33 +70,18 @@ func routes(_ app: Application) throws {
             app.get(":owner", ":repository", ":reference", "**") {
                 try await packageController.documentation(req: $0, fragment: .root)
             }
-            app.get(":owner", ":repository", ":reference", "css") {
-                try await packageController.documentation(req: $0, fragment: .css)
-            }
```
finestructure, line 75: I believe these routes that have a matching `**` cousin aren't really used. IIUC, these are "browsing" routes that would resolve to `css/index.html` - which doesn't actually exist. So they should be safe to drop.

This could use some more testing/review but I'd like to trim the set of required routes down to a minimum before starting the transition to Pointfree's routing.

daveverwer: Let's maybe make #1917 a part of this PR, too. I believe what's needed there is to focus the wildcard routes that we added for the `favicon` and I think there was another one, too, down to specific routes.

PR #1969, opened 2022-08-13
Make all controllers enums with static funcs

finestructure:
Controllers don't carry any state, so top level or static functions are a better model for them than instantiating a bare struct (without any properties) just to call a function on it. This will also reduce the change footprint for the proposed parser/printer site routing changes.

### DIFF
--- a/Sources/App/routes.swift
+++ b/Sources/App/routes.swift
@@ -55,80 +55,74 @@ func routes(_ app: Application) throws {
     }
 
     do {  // package pages
-        let packageController = PackageController()
-
         do {  // temporary, hacky docc-proxy
             app.get(":owner", ":repository", ":reference", "documentation") {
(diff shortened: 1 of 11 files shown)


Sources/App/routes.swift
```diff
@@ -55,80 +55,74 @@ func routes(_ app: Application) throws {
...
     do {  // package collection page
         app.get(SiteURL.packageCollection(.key).pathComponents,
-                use: PackageCollectionController().generate)
+                use: PackageCollectionController.generate)
     }
 
     do {  // author page
-        let authorController = AuthorController()
-        app.get(SiteURL.author(.key).pathComponents, use: authorController.show)
+        app.get(SiteURL.author(.key).pathComponents, use: AuthorController.show)
```
finestructure, line 109: This little diff demonstrates the purpose of this change best.

daveverwer (approved):

PR #1975, opened 2022-08-15
Rework documentation fragments

finestructure:


Sources/App/Controllers/PackageController+routes.swift
```diff
@@ -380,9 +381,9 @@ extension PackageController {
         let baseURL = "http://\(baseURLHost)/\(baseURLPath)"
 
         switch fragment {
-            case .css, .data, .documentationRedirect, .documentation, .img, .index, .js:
+            case .css, .data, .documentation, .img, .index, .js:
```
finestructure, line 384: I'm not sure how this actually worked but the URL generated by `documentationRedirect` are definitely incorrect:

```
http://awsDocsBucket.s3-website.us-east-2.amazonaws.com/owner/package/1.2.3/documentationRedirect/
```

Instead, I'm using

```
let isDocumentationRedirect = (fragment == .documentation && archive == nil)
```

to detect the redirect. I _think_ that's what the want but TBH that whole function `documentation(req:fragment)` has grown way too big.

See https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/pull/1945/files#r945803903

daveverwer: Agree completely that this funcrtion is out of control! 😬 I considered breaking it up several times during the versioned documentation changes, but there's no clear best approach to that.

The documentationRedirect is one of those "one-way" routes we were talking about this morning. Only ever use incoming, we never use it to generate a URL.

daveverwer (approved):
Changes look good. We should talk about a way to refactor this, too.

Sources/App/routes.swift
```diff
@@ -57,7 +57,7 @@ func routes(_ app: Application) throws {
     do {  // package pages
         do {  // temporary, hacky docc-proxy
             app.get(":owner", ":repository", ":reference", "documentation") {
-                try await PackageController.documentation(req: $0, fragment: .documentationRedirect)
+                try await PackageController.documentation(req: $0)
```
finestructure, line 60: In fact, the better way to deal with this is to simply route to a different handler that is much simpler that what we have in our current `documentation(req:fragment:)` handler - it doesn't even need to call out to AWS at all - and that's also why the bad url doesn't even have any effect I believe. We were making an unnecessary AWS call when hitting the root documentation URL without an `:archive` path, ignoring entirely what it did, if I'm reading this right.

finestructure: This also trims down the long `documentation(req:fragment:)` function a bit.

daveverwer: You're correct, we don't need to hit AWS for this. Only look whether there are `docArchives` and redirect.

Sources/App/Controllers/PackageController+routes.swift
finestructure: I think at least everything inside the `case .documentation` should be a function, something we can also test in isolation. There might be further ways to break this down and it might make sense to do this before fiddling with `SiteURL`.

Tests/AppTests/PackageController+routesTests.swift
```diff
@@ -365,9 +365,12 @@ class PackageController_routesTests: AppTestCase {
 
         // MUT
         try app.test(.GET, "/owner/package/1.2.3/documentation") {
+            // triggers a redirect which then finds no archive (doesn't actually
```
daveverwer, line 368: This is potentially the wrong place for this comment, or at least it should be in another place, too. This test tests the `404` state when there are no `docArchives`. 

The successful test where this really comes into play is `test_documentationRoot_redirect`

Sources/App/Controllers/PackageController+routes.swift
finestructure: I won't do this as part of this PR. That's logic that's happening after routing so it doesn't really need to be tested for the purpose of switching out the routers.

Tests/AppTests/PackageController+routesTests.swift
finestructure: This tests a different scenario though. The redirect tests cover the handling of a 404 from AWS, this one here is a 500, which we actually surface back.

I've updated the inputs so the root access shows the redirect and doesn't fail with a 404, because that's already covered in the redirect tests. It's now asserting that the 500 from AWS doesn't affect the root doc route and remains a redirect.

PR #1976, opened 2022-08-15
Fix order of documentation archives being alphabetical

### LINKED ISSUE #1974: Use SPI manifest `documentation_targets` sort order for documentation archives
### Discussed in https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/discussions/1970

<div type='discussions-op-text'>

<sup>Originally posted by **Mordil** August 14, 2022</sup>
I just added SPI support for hosting DocC documentation, and the output is not following [my specified ordering in the YAML file](https://github.com/Mordil/RediStack/blob/master/.spi.yml):

```yaml
builder:
  configs:
    - documentation_targets: ['RediStack', 'RedisTypes', 'RediStackTestUtils']
```

Instead, it sorts alphabetically, which leaves the module [**RedisTypes** as the entry point for documentation and first in the list of module documentation](https://swiftpackageindex.com/Mordil/RediStack/master/documentation/redistypes).

This causes me a problem, because **RediStack** is the main purpose of the package.</div>

daveverwer:
Fixes #1974 

So as I mentioned in our chat on Discord, what I’d love to do is make the logic in `docArchivesInManifestOrder` be the default getter for `docArchives`. There should be no reason to ever get these out in alphabetical order.

I tried making the `docArchives` property that’s annotated with the Fluent `@Field` attribute, but that removes the `$docArchives` definition needed for selecting the fields. Looks like that’s a `FieldProperty`, so it’s possible we could re-create that, but it feels like maybe going down the wrong road. I also tried asking in the Vapor Discord, and the consensus was that replacing the getter would be tricky and probably not worth the work.

So, I have a new plan of attack for tomorrow. A fix to how the data gets inserted via the API and a migration to correct all the data we have inserted so far.

Putting this up in draft in case you have any better ideas @finestructure.

### DIFF
--- a/Package.resolved
+++ b/Package.resolved
@@ -141,8 +141,8 @@
         "repositoryURL": "https://github.com/SwiftPackageIndex/SPIManifest",
         "state": {
           "branch": null,
-          "revision": "ad6d1fb021022788af310e9048dbecfaa14e0aa1",
-          "version": "0.7.1"
+          "revision": "36eac3222fd744fe3b5621c351751d83ed518d64",
+          "version": "0.7.2"
         }
       },
       {
@@ -240,8 +240,8 @@
         "repositoryURL": "https://github.com/apple/swift-log.git",
         "state": {
           "branch": null,
-          "revision": "8fc79cae9d03a5e623c52cc2dbd2df165a7832b0",
-          "version": "1.4.3"
+          "revision": "6fe203dc33195667ce1759bf0182975e4653ba1c",
+          "version": "1.4.4"
         }
       },
       {
--- a/Package.swift
+++ b/Package.swift
@@ -36,7 +36,7 @@ let package = Package(
         .package(name: "SnapshotTesting",
                  url: "https://github.com/pointfreeco/swift-snapshot-testing.git", from: "1.7.2"),
         .package(url: "https://github.com/SwiftPackageIndex/SemanticVersion", from: "0.3.0"),
-        .package(url: "https://github.com/SwiftPackageIndex/SPIManifest", from: "0.7.1"),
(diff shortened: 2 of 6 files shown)


finestructure:
I think this change is reasonable as is. IIUC, from now on we're storing in the correct order. We could simply review this in 6 months or so and remove `docArchivesInManifestOrder` on account of enough recent data having moved to the new order without having to deal with a full-blown migration 🤷‍♂️

finestructure (approved):

daveverwer:
I don't think we are now storing it in the right order. Yesterday's PR on SPIManifest made sure the *targets* didn't come in in the wrong order. My plan for this is to now make sure the archives match that order on insert, and I think we should also do the migration as we need to fix at least @Mordil's package. We should fix them all.

Mordil:
Just wanna say, thank you both for working on this so quickly and being responsive!

daveverwer:
Our pleasure! I jumped on it quickly as it's actually a really undesirable situation that you end up with and we want this to be the *best* way to host docs, not just *a* way 🚀

I think we fixed it on the back-end side with the builder, and I don't think this PR will be needed. It's deploying now and I'll test it later this evening with your package.

daveverwer:
I have good news, and I have bad news...

The good news is that the builder is deployed, and I just re-built your package with the new builder version. [Logs here](https://swiftpackageindex.com/builds/273C8456-FA55-4879-BF3D-91AE6583CC0F) and 4.1.4 was the version that contained this fix.

The bad news is that it did not correctly sort your documentation archives as it should have.

Will investigate more tomorrow.

daveverwer:
I didn't want to call this fixed until all docsets had finished generating for the affected packages. That has now been completed, and all looks good, so I'm going to close this and #1974.

PR #1980, opened 2022-08-18
Match the new package guidelines.

daveverwer:


finestructure:
We should probably only merge this once we've updated validation, right?

finestructure (approved):

Issue #1947, opened 2022-08-03
Re-analyze version skipping reanalysis of some versions, maybe?


daveverwer:
We ran a re-analysis last night, and the feature isn't working as expected. At least it appears not to be!

The good news is that there are now 1,763 rows in `versions` covering 88 packages with `has_binary_targets` set. So, the analysis is definitely setting data.

The other good news is that where the flag has been set on the default branch version, the flag is correctly displayed on the page. See [SemverBridge](https://staging.swiftpackageindex.com/Jomy10/SemverBridge) or [Firebase](https://staging.swiftpackageindex.com/firebase/firebase-ios-sdk) for examples.

There was a bit of a bad smell, though. There are 88 distinct packages with the flag set on any of their versions, but only 43 where it is set on the default branch version. This may not be a problem, but it struck me as suspicious that over 50% of packages that had a binary target at some point no longer have it. 

I investigated some examples:

* [Pulse](https://staging.swiftpackageindex.com/kean/Pulse) has no flag on the package page. I looked back in the version history in our database and it was v0.15.0 that had the flag set, and [sure enough, there’s a binary target in that Package.swift](https://github.com/kean/Pulse/blob/0.15.0/Package.swift) - ✅ No problem found.
* [Facebook](https://staging.swiftpackageindex.com/facebook/facebook-ios-sdk) is a problem, but one which is potentially caused by a complicated `Package.swift` file. There are 16 versions in the database flagged as having binary targets, but the default branch version does not. The `Package.swift` is so complex I can’t manually compile it in my head, though 😂 - ⚠️ Needs investigation.
* [OnfidoAuth](https://staging.swiftpackageindex.com/onfido/onfido-auth-sdk-ios) looks like a legitimate problem. All tagged versions apart from 0.2.4, and the default branch version have the flag set, but the `Package.swift` file on the default branch [definitely has a binary target](https://github.com/onfido/onfido-auth-sdk-ios/blob/main/Package.swift) - 🛑 Confirmed error.
* [Appfigurate](https://staging.swiftpackageindex.com/electricbolt/appfiguratesdk) is the same deal. Four versions with the flag set but not the default branch. However, look at the `Package.swift` file [on the default branch](https://github.com/electricbolt/appfiguratesdk/blob/master/Package.swift) - 🛑 Confirmed error.

I stopped looking for more examples at this point.

[...]

PR #1999, opened 2022-08-31
Enable “has binary targets” in production

daveverwer:
Thanks so much, @samdeane, and apologies that this took so long to go live!

There is still the outstanding issue in #1947, but it may be a while until that's fixed so let's get this change live. More [context here](https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/issues/1947#issuecomment-1232705370).


daveverwer:
I think we should hold off on promoting it, though, as people who have packages with binary targets will check their page and it'll increase support load as we explain the subtleties of #1947.

finestructure (approved):

daveverwer:
Deploying as 2.75.6

PR #2027, opened 2022-09-15
Fix 4.7 downscaling

finestructure:


daveverwer (approved):

finestructure:
I'll deploy and monitor this tonight. It's easy to roll back in case there are problems.

PR #2028, opened 2022-09-15
SPI-Indexing 11

finestructure:


daveverwer (approved):

Sherlouk:
I'm sure there's a reason we're not, but could we use the Twitter API for this? 👀

daveverwer:
> I'm sure there's a reason we're not

We could do this PR for 500 episodes before it'd be more efficient to automate. 😂

Sherlouk:
> We could do this PR for 500 episodes before it'd be more efficient to automate. 😂

B-but what if we had a page where you can see all of the past episodes!?

and what if it had all these amazing magical podcast features!? 😂

Worth opening an issue if someone feels like it? Feels like a relatively easy piece 😊

daveverwer:
> B-but what if we had a page where you can see all of the past episodes!?

https://blog.swiftpackageindex.com

> and what if it had all these amazing magical podcast features!? 😂

I too have a list of improvements for Twitter Spaces 😂

PR #2032, opened 2022-09-16
Tweak error logging

finestructure:
For some reason the log only shows the last line, perhaps multi-line errors are getting truncated:

```
2022-09-16 12:46:38 | error: keyNotFound(CodingKeys(stringValue: "web_url", intValue: nil), Swift.DecodingError.Context(codingPath: [], debugDescription: "No value associated with key CodingKeys(stringValue: \"web_url\", intValue: nil) (\"web_url\").", underlyingError: nil)) [component: trigger-builds]
```

It worked before but that was when throwing an error - not that that should make a difference 🤷‍♂️

Either way, I just need the status code info so I can handle the rate limiting error and show only actual errors beyond that.

### DIFF
--- a/Sources/App/Core/Gitlab.swift
+++ b/Sources/App/Core/Gitlab.swift
@@ -20,7 +20,6 @@ enum Gitlab {
     static let baseURL = "https://gitlab.com/api/v4"
 
     enum Error: LocalizedError {
-        case decodingFailed(String)
         case missingConfiguration(String)
         case missingToken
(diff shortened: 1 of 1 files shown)


daveverwer (approved):

finestructure:
Brilliant. Now that it's deployed, we've drained the < 5.7 backlog and the error condition isn't triggering anymore 🤪

daveverwer:
> we've drained the < 5.7 backlog

They'll be back 😂

PR #2041, opened 2022-09-23
Update dockerfile

finestructure:
Adopts static linking from Vapor template Dockerfile but skips adding vapor user.

- [x] deploy from branch to dev
- [x] check that new versions are properly analysed http://spi-dev.swiftpackageindex.com:3000/d/K2re2w2Gk/spi-stats?from=now-6h&to=now&viewPanel=26

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,10 +12,15 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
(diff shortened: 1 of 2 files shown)


daveverwer (approved):

finestructure:
New versions coming through as they should:

CleanShot 2022-09-23 at 12 30 46@2x

PR #2048, opened 2022-09-28
Bump to latest revision on all monitoring components

finestructure:
Will deploy from branch to dev to test this.

Deployment notes:

- will need to stop and remove loki config in order to update it:
```
failed to update config mon_loki_cfg: Error response from daemon: rpc error: code = InvalidArgument desc = only updates to Labels are allowed
```

Process:

- stop loki
- remove config

```
docker service rm mon_loki
docker config rm mon_loki_cfg
```

- [x] figure out the procedure of this on a multi-node swarm
- [x] find and replace prometheus-loki data source related queries/alerts with grafana-based one(s)

### DIFF
--- a/infrastructure/loki.yml
+++ b/infrastructure/loki.yml
@@ -30,6 +30,8 @@ ingester:
   chunk_target_size: 1048576  # Loki will attempt to build chunks up to 1.5MB, flushing first if chunk_idle_period or max_chunk_age is reached first
   chunk_retain_period: 30s    # Must be greater than index read cache TTL if using an index cache (Default index read cache TTL is 5m)
   max_transfer_retries: 0     # Chunk transfers disabled
+  wal:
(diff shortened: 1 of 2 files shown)


daveverwer (approved):

finestructure:
At least one error remaining:

CleanShot 2022-09-28 at 11 18 37@2x

finestructure:
I believe this Prometheus-Loki data source (which IIRC was to use prometheus alerting on logs, which Grafana didn't support) is now deprecated in favour of using Grafana's alerting: https://github.com/grafana/grafana/pull/34650

finestructure:
Ok, here's the graph/alert that's affected:

CleanShot 2022-09-28 at 13 13 51@2x

Should be feasible to port that query over to whatever the new mechanism actually is.

PR #2054, opened 2022-10-04
Add a default documentation redirect

### LINKED ISSUE #2053: Move logic to redirect to default documentation into a route handler
- [ ] Make a generic link that encodes the logic we use for the "Documentation" link from package pages. So, visiting https://swiftpackageindex.com/vapor/postgres-nio/documentation would redirect to https://swiftpackageindex.com/vapor/postgres-nio/1.11.1/documentation/postgresnio or https://swiftpackageindex.com/vapor/postgres-nio/main/documentation/postgresnio if there was no stable release.
- [ ] Remove that logic from the package page so that right-clicking the "Documentation" page lets people copy a stable link that will always work.

Thanks to @fabianfett for the suggestion!

daveverwer:
Fixes #2053

Does not address https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/issues/2053#issuecomment-1265584000 as that change needs doing in two places and this PR is large enough.

### DIFF
--- a/Sources/App/Controllers/PackageController+routes.swift
+++ b/Sources/App/Controllers/PackageController+routes.swift
@@ -92,6 +92,22 @@ enum PackageController {
         var updatedAt: Date
     }
 
+    static func defaultDocumentation(req: Request) async throws -> Response {
+        guard
+            let owner = req.parameters.get("owner"),
+            let repository = req.parameters.get("repository")
+        else {
+            throw Abort(.notFound)
+        }
+
+        let result = try await PackageResult.query(on: req.db, owner: owner, repository: repository)
+        if let documentationUrl = result.defaultDocumentationUrl {
+            throw Abort.redirect(to: documentationUrl)
+        } else {
+            throw Abort(.notFound)
+        }
+    }
+
     static func documentation(req: Request) async throws -> Response {
         guard
             let owner = req.parameters.get("owner"),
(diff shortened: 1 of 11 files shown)


Sources/App/Controllers/PackageController+routes.swift
```diff
@@ -92,6 +92,22 @@ enum PackageController {
...
 
+    static func defaultDocumentation(req: Request) async throws -> Response {
+        guard
+            let owner = req.parameters.get("owner"),
+            let repository = req.parameters.get("repository")
+        else {
+            throw Abort(.notFound)
+        }
+
+        let result = try await PackageResult.query(on: req.db, owner: owner, repository: repository)
```
finestructure, line 103: This works great.

There's one inefficiency here and I've spent a little time exploring if/how to avoid it. When we hit the default URL, we do the following:

* call `defaultDocumentation(req:)` via /simonbs/Runestone/documentation
* run `PackageResult.query(...)`
* redirect to /simonbs/Runestone/0.2.5/documentation/runestone
* call `documentation(req:fragment:)` via that /simonbs/Runestone/0.2.5/documentation/runestone redirect
* run `Joined3.query(...)`
* proxy to http://spi-prod-docs.s3-website.us-east-2.amazonaws.com/simonbs/runestone/0.2.5/documentation/runestone

I.e. we run a query that gets us the doc parameters and then we redirect to a url - which runs essentially the same query again to then proxy to AWS.

I think that's fixable. (I've got a change locally but it's currently failing some tests.)

This isn't a new problem actually - we're already doing the same for the redirect when we get an underspecified doc request like

http://localhost:8080/simonbs/Runestone/0.2.5/documentation

This goes to `documentation(req: Request)`, which runs a `Joined3.query(...)` to find the archive and then redirects to 

http://localhost:8080/simonbs/Runestone/0.2.5/documentation/runestone

and this will run another `Joined3.query(...)`.

We should have all the information to avoid that redirect and second query.

I'll go ahead and merge this in the meantime.

finestructure (approved):

PR #2055, opened 2022-10-04
Catch all on documentation routes without a reference

daveverwer:


Sources/App/Controllers/PackageController+routes.swift
```diff
@@ -102,7 +102,10 @@ enum PackageController {
 
         let result = try await PackageResult.query(on: req.db, owner: owner, repository: repository)
         if let documentationUrl = result.defaultDocumentationUrl {
-            throw Abort.redirect(to: documentationUrl)
+            var pathComponents = req.parameters.getCatchall()
+            pathComponents.insert(documentationUrl, at: 0)
+            let fullDocumentationUrl = pathComponents.joined(separator: "/")
+            throw Abort.redirect(to: fullDocumentationUrl)
```
daveverwer, line 108: So this is an interesting problem. The documentation URL that `defaultDocumentationUrl` passes back includes the archive in the URL. So, for example with Lindenmayer, it returns:

```
/heckj/lindenmayer/documentation/lindenmayer
```

Since the archive isn’t a parameter in this route, if we grab the catch all it includes the archive in the catch all and the redirect ends up as:

```
/heckj/lindenmayer/documentation/lindenmayer/lindenmayer
```

I think we’ll need a new route where we also have the archive as a parameter. That’s more complicated than I’d have liked.

finestructure: I think this might be fixed automatically by the change I'm making to address the double query (#2056)

PR #2058, opened 2022-10-05
WIP

finestructure:
All tests pass locally but default docs route shows an error:

CleanShot 2022-10-05 at 10 51 22@2x

while the direct link works:

CleanShot 2022-10-05 at 10 51 26@2x

Both route through the (WIP) function

```swift
    static func documentation(req: Request,
                              documentationVersions: [DocumentationVersion],
                              reference: String,
                              archive: String?,
                              awsResponse: ClientResponse,
                              owner: String,
                              repository: String,
                              fragment: Fragment) async throws -> Response {
```

and I cannot figure out why one would fail but not the other. All input parameters except `req` are identical. I can't see what in `req` would cause the different behaviour.


finestructure:
Superseded by #2063

Issue #175, opened 2020-05-27
Wire up author information in the Package page


daveverwer:
Depends on #173

finestructure:
I believe this is fully done now, isn't it? Closing for now.

Issue #174, opened 2020-05-27
Allow overriding of author information from the additional metadata file


daveverwer:
The name and link for every displayed author should be able to be overridden by our extra metadata file.

The *list of contributors* should be custom at this point, and we should replace any automated data with *only* the names and links in the metadata file. The link may point to *any* web site, not just their GitHub profile. (Yes, this opens up the possibility of spam, but let's deal with that as it comes up).

PR #2060, opened 2022-10-07
Authors

### LINKED ISSUE #173: Back-end support for gathering author information
- [ ] Extend schema
- [ ] Collect data

Here's a basic plan:

* Any contributor with more than 10 commits is eligible to be listed here. This rule will save projects where someone fixes a typo in a README in a newly released project with just an "Initial commit."
* For all eligible contributors, they are listed if they have made more than 5% of the total commits.
* The name should be fetched from the GitHub API for the username
* The link should go to the contributors GitHub profile.

jcubit:
This change fixes #173 and #175 . 
Screenshot 2022-10-07 at 17 15 39
Screenshot 2022-10-07 at 17 07 07


daveverwer:
I'm so happy to see this @jcubit! Thank you for all your work on it.

 I'll give it a first review on Monday 😂

Tests/AppTests/AnalyzerTests.swift
```diff
@@ -137,6 +143,16 @@ class AnalyzerTests: AppTestCase {
...
+            if cmd == .gitShortlog { return
+                """
+                10 Person 1 <person1@example.com>
+                 2 Person 2 <person2@example.com>
+                """
+            }
+
+            if cmd == .gitShortlog {
+                return "10 Person 1 <person1@example.com>"
+            }
```
Sherlouk, line 155: Second block can't be called, right? 👀

(you know it's good code when this is the only thing I spot 😂)

Sources/App/Views/PackageController/PackageShow+Model.swift
```diff
@@ -296,11 +296,22 @@ extension PackageShow.Model {
...
+        var nodes = authors.authors.map { author -> Node<HTML.BodyContext> in
+            if let authorURL = author.url {
+                return .a(.href(authorURL), .text(author.name))
+            } else {
+                return .text(author.name)
+            }
+        }
+
+        if (authors.numberOfContributors > 0) {
+            nodes.append(.text("\(authors.numberOfContributors) other " + "contributor".pluralized(for: authors.numberOfContributors)))
```
Sherlouk, line 309: Should the number here be numberOfContributors minus the number of author nodes shown on the page?

"Written by Author 1, Author 2 and 3 other contributors" where the numberOfContributors is 5?

May be mistaken on what the values are.

edit:// scrap that I've just seen the earlier code where the minus is done. No problem here 😛

Sources/App/Core/PackageContributorService.swift
```diff
@@ -0,0 +1,133 @@
...
+}
+
+
+
+final class PackageContributorService {
+    
+    
+    /// Extracts the possible authors of the package according to the number of commits.
+    /// A contributor is considered an author when the number of commits is at least a 60 percent
+    /// of the maximum commits done by a contributor
```
Sherlouk, line 116: I'm not sure I understand the comment, maths or functionality here.

We return and display an array of authors, but only one person can have ever created more than 60% of the total commits.

Or is the behaviour here that we take the top contributor (who is always an author), and then anybody who has more than 60% of that count is also an author? So essentially a top 40% performer?

(Not saying it's wrong but just some clarification on expected behaviour for my own brain aha)

Sources/App/Core/PackageContributorService.swift
```diff
@@ -0,0 +1,133 @@
...
+    }
+    
+    /// Parses the result of queryVCHistory into a collection of contributors
+    private static func parseVCHistory(log: String) throws -> [Contributor] {
+        var committers = [Contributor]()
+        
+        for line in log.components(separatedBy: .newlines) {
+            let log = line.split(whereSeparator: { $0 == " " || $0 == "\t"})
+            
+            if (log.count != 2) {
```
Sherlouk, line 65: What's special about two here? 

My understanding is there is at least 3 pieces of information we expect, so perhaps a >= would be better?

(Especially as we have some force unwraps in here which I suppose could crash if log.count was 0 or 1)

Sherlouk (commented):
My comments hold zero weight but was interested so thought I'd take a quick read through over lunch 😛

Tests/AppTests/AnalyzerTests.swift
jcubit: upps! this slip in when we were analysing this. We did not want to give person 2 a chance 😂

Sources/App/Core/PackageContributorService.swift
jcubit: We take the top contributor (who is always an author), and then anybody who has *at least* 60% of that count is also an author. Say for instance the top contributor has 100 commits, then anybody who has at least 60 commits (a bit more than the half of that top count) is also an author. The 60% is actually arbitrary and rather irrelevant for our bigger purpose; the key point here is that we want to have a *default* way to extract authors information. There is no perfect way to get this right for all the cases. An algorithm that only takes into account the number of commits as a metric for authorship is certainly questionable. However, we do not need to dwell on this as the authors themselves will have the last word on authorship and what author information is shown 🤓. This last thing will be addressed in #174  :)

Sherlouk: Makes sense - thanks for explaining 😀

As you say hard to perfect as there are so many distributions of contributors (some repos are very flat with lots of contributions but all at the same level, but also some with basically one author with 90% of the total commits). 

The issue attached will probably resolve the "Core Maintainers" concept which some orgs institute.

Sherlouk:
Separate issue entirely but this data also means we could, if desired, add contributions to author pages.

Like "Sherlouk" owns these 5 repos, and is an author of these 20 repos. Could be good to see active community members.

Suppose there's also the idea of top open source contributors leaderboard across all tracked packages 👀

@daveverwer just spitting ideas - any of them sound interesting to you?

Sources/App/Core/PackageContributorService.swift
jcubit: Oh yes, `>2` would be better. Thank you! Under the assumption that the logs are of the form `numberOfCommits identifier email` the logic here is
1. take the first token as the number of commits
2. take the last token as the email
3. the identifier is the join of the tokens in between (some people just write one name, others full name)

Sources/App/Core/PackageContributorService.swift
```diff
@@ -0,0 +1,133 @@
...
+    
+    
+    /// Extracts the possible authors of the package according to the number of commits.
+    /// A contributor is considered an author when the number of commits is at least a 60 percent
+    /// of the maximum commits done by a contributor
+    /// - Parameters:
+    ///   - cacheDirPath: path to the cache directory where the clone of the package is stored
+    ///   - packageID: the UUID of the package
+    /// - Returns: PackageAuthors
+    static func authorExtractor(cacheDirPath: String, packageID: UUID?) throws -> PackageAuthors {
```
daveverwer, line 121: Nit: I wonder if `cacheDirPath` is the best name here, maybe `gitRepositoryPath`?

Sources/App/Core/PackageContributorService.swift
```diff
@@ -0,0 +1,133 @@
...
+        }
+        return committers
+    }
+    
+}
+
+
+
+/// Strategy for selecting contributors based entirely on the number of commits 
+struct CommitSelector {
```
daveverwer, line 86: We could potentially remove this type, moving the `filter` function into the `PackageContributorService`, maybe as a private function with a bit more context. Something like `primaryContributors()` maybe?

Sources/App/Core/PackageContributorService.swift
```diff
@@ -0,0 +1,133 @@
...
+        })!.numberOfCommits
+        
+        return candidates.filter { canditate in
+            return Float(canditate.numberOfCommits) > threshold * Float(maxNumberOfCommits)
+        }
+    }
+}
+
+
+struct PackageAuthors : Encodable, Decodable, Equatable {
```
daveverwer, line 104: Nit: We could potentially structure the order of these types to make more sense for others coming to look at this file later. I'd suggest we start with `PackageAuthors` being the first type in the file, followed by `PackageContributorService` and finally the `GitHistoryLoader`, which could even be nested inside `PackageContributorService`.

Sources/App/Commands/Analyze.swift
```diff
@@ -357,11 +357,14 @@ extension Analyze {
         guard let gitDirectory = Current.fileManager.cacheDirectoryPath(for: package.model) else {
             throw AppError.invalidPackageCachePath(package.model.id, package.model.url)
         }
-
+        
         repo.commitCount = (try? Current.git.commitCount(gitDirectory)) ?? 0
         repo.firstCommitDate = try? Current.git.firstCommitDate(gitDirectory)
         repo.lastCommitDate = try? Current.git.lastCommitDate(gitDirectory)
 
+        let pkgAuthors = try PackageContributorService.authorExtractor(cacheDirPath: gitDirectory, packageID: package.model.id)
```
daveverwer, line 365: Nit: This could be one line.

daveverwer (commented):
This is looking so good Javier! I just did a first pass over it. Most of it is naming/small tweaks, so don't get disheartened!

Sources/App/Views/PackageController/PackageResult+ModelSupport.swift
```diff
@@ -1,4 +1,4 @@
-// Copyright 2020-2021 Dave Verwer, Sven A. Schmidt, and other contributors.
+// Copyright 2020-2022 Dave Verwer, Sven A. Schmidt, and other contributors.
```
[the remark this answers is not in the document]
daveverwer: Don't worry. When I do it project-wide, I'll use a regular expression, so it catches every file!

Sources/App/Models/Package.swift
```diff
@@ -111,6 +111,7 @@ extension Package {
         case analysisFailed = "analysis_failed"
         case ingestionFailed = "ingestion_failed"
         case invalidCachePath = "invalid_cache_path"
+        case unexistentCacheDir = "unexistent_cache_dir"
```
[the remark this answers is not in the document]
jcubit: I should have tried with the closest word to Spanish (`inexistente` 😉) instead of inventing words 😂. I like the longer, but clearer name `cacheDirectoryDoesNotExist `

Sources/App/Core/PackageContributorService.swift
```diff
@@ -0,0 +1,133 @@
...
+                                              "queryVCHistory failed: \(error.localizedDescription)")
+        }
+    }
+    
+    /// Parses the result of queryVCHistory into a collection of contributors
+    private static func parseVCHistory(log: String) throws -> [Contributor] {
+        var committers = [Contributor]()
+        
+        for line in log.components(separatedBy: .newlines) {
+            let log = line.split(whereSeparator: { $0 == " " || $0 == "\t"})
```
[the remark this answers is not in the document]
daveverwer: In that case, I think we can simplify this code by splitting only on tabs and then parsing the email address (if there is one) out of the name section. Let's chat about this on our call tonight. 👍

daveverwer:
Great work on the amendments, @jcubit! I think this is getting really close. 👍

Sources/App/Commands/Analyze.swift
```diff
@@ -357,10 +357,11 @@ extension Analyze {
         guard let gitDirectory = Current.fileManager.cacheDirectoryPath(for: package.model) else {
             throw AppError.invalidPackageCachePath(package.model.id, package.model.url)
         }
-
+        
         repo.commitCount = (try? Current.git.commitCount(gitDirectory)) ?? 0
         repo.firstCommitDate = try? Current.git.firstCommitDate(gitDirectory)
         repo.lastCommitDate = try? Current.git.lastCommitDate(gitDirectory)
+        repo.authors = try PackageContributors.extract(gitCacheDirectoryPath: gitDirectory, packageID: package.model.id)
```
Sherlouk, line 364: @daveverwer: A number of the previous data extractions are using `try?` in order to ignore thrown errors. Are we happy that author extraction will completely throw the repo update if it fails?

Sources/App/Commands/Analyze.swift
```diff
@@ -724,7 +725,10 @@ extension Analyze {
             logger.warning("Twitter.postToFirehose failed: \(error.localizedDescription)")
         }
     }
+    
 
+    
+
```
Sherlouk, line 731: nit: formatting (excessive whitespace)

@daveverwer: We don't have a format/linter specified, right? Are we against adding one to make it easier for folks to contribute while maintaining a common style? A small bonus is that [SwiftFormat](https://github.com/nicklockwood/SwiftFormat) will handle all the file headers automatically which I know you said you were going to update.

Sources/App/Core/AppError.swift
```diff
@@ -37,6 +38,8 @@ enum AppError: LocalizedError {
                 return "Invalid packge URL: \(value) (id: \(id))"
             case let .invalidPackageCachePath(id, value):
                 return "Invalid packge cache path: \(value) (id: \(id))"
+            case let .cacheDirectoryDoesNotExist(id, value):
+            return "Package cache directory, \(value), does not exist: (id: \(id)"
```
Sherlouk, line 42: nit: formatting (line should be indented one level, as per other returns above)

Sources/App/Core/PackageContributors.swift
```diff
@@ -0,0 +1,132 @@
...
+
+import Foundation
+import ShellOut
+import Vapor
+
+
+struct PackageAuthors : Encodable, Decodable, Equatable {
+    var authors : [Author]
+    var numberOfContributors : Int
+}
```
Sherlouk, line 23: nit: space before colon shouldn't exist

Very, **very**, picky but it's valuable to have consistency in the way we write our code to make it as easy as possible for others to read and pick up on common patterns.

Additionally, you can use `Codable` which is simply a typealias for a type which is both `Encodable` and `Decodable`.

```suggestion
struct PackageAuthors: Codable, Equatable {
    var authors: [Author]
    var numberOfContributors: Int
}
```

Sources/App/Core/PackageContributors.swift
```diff
@@ -0,0 +1,132 @@
...
+                throw AppError.cacheDirectoryDoesNotExist(packageID, gitCacheDirectoryPath)
+            }
+
+            // attempt to shortlog
+            do {
+                return try Current.git.shortlog(gitCacheDirectoryPath)
+            } catch {
+                throw AppError.shellCommandFailed("gitShortlog",
+                                                  gitCacheDirectoryPath,
+                                                  "queryGitHistory failed: \(error.localizedDescription)")
```
Sherlouk, line 74: Since we map all `Current.git.shortlog` errors to `invalidString` the localised description here won't provide much value. Perhaps it would be worth throwing the original error in the shortlog function so we have a better understanding of why it failed?

Sources/App/Core/PackageContributors.swift
```diff
@@ -0,0 +1,132 @@
...
+        }
+
+        /// Parses the result of queryGitHistory into a collection of contributors
+        private static func parseGitHistory(log: String) throws -> [Contributor] {
+            var committers = [Contributor]()
+
+            for line in log.components(separatedBy: .newlines) {
+                let log = line.split(whereSeparator: { $0 == " " || $0 == "\t"})
+
+                if (log.count > 2) {
```
Sherlouk, line 85: nit: this would probably read nicer as a `guard`.

```
guard log.count > 2 else {
  continue
}

// continue your logic for handling identifier/number extraction.
```

Sources/App/Core/PackageContributors.swift
```diff
@@ -0,0 +1,132 @@
...
+                    var identifier = [String]()
+                    for i in 1..<(log.count - 1) {
+                        identifier.append(String(log[i]))
+                    }
+                    var numberOfCommits = 0
+                    if (log.first != nil) {
+                        numberOfCommits = Int(log.first!) ?? 0
+                    }
```
Sherlouk, line 93: I know it's been mentioned above that we need to improve the safety of the line here (the less exclamation marks the better!)

Given with the guard/if statement we know there is at least 3 items in the array. We could maybe do something like:

```swift
let numberOfCommits = Int(log.removeFirst()) ?? 0
let identifier = log.dropLast().joined(separator: " ")
let committer = Contributor ...
```

Docs: https://developer.apple.com/documentation/swift/array/removefirst(_:)
Docs: https://developer.apple.com/documentation/swift/array/droplast(_:)

This would remove the need for `for` loops and array builders!

Sources/App/Core/PackageContributors.swift
```diff
@@ -0,0 +1,132 @@
...
+
+}
+
+
+
+
+
+
+
+
```
Sherlouk, line 132: nit: formatting (excessive line breaks)

Sources/App/Models/Repository.swift
```diff
@@ -126,7 +126,7 @@ final class Repository: Model, Content {
     
     init(id: Id? = nil,
          package: Package,
-         authors: [Author] = [],
+         authors: PackageAuthors? = nil, // [Author] = [],
```
Sherlouk, line 129: nit: cleanup the comment here
```suggestion
         authors: PackageAuthors? = nil,
```

Sources/App/Views/PackageController/PackageShow+Model.swift
```diff
@@ -296,11 +296,23 @@ extension PackageShow.Model {
...
+        guard let authors else { return .empty }
+        var nodes = authors.authors.map { author -> Node<HTML.BodyContext> in
+            if let authorURL = author.url {
+                return .a(.href(authorURL), .text(author.name))
+            } else {
+                return .text(author.name)
+            }
+        }
+
+        if (authors.numberOfContributors > 0) {
```
Sherlouk, line 309: nit: no need for brackets around the if statement

```suggestion
        if authors.numberOfContributors > 0 {
```

Sherlouk (commented):
Mostly nit/nitpicks (minor trivial issues that just ensure we have a bit of a common style to how we write code). Feels like we could introduce a formatter (in a separate pull request) to make this easier!

Sources/App/Core/PackageContributors.swift
jcubit: @Sherlouk could you elaborate a bit more here what you mean :)

Sources/App/Core/PackageContributors.swift
jcubit: That's interesting! Thank you for bringing this array methods up. This will indeed help removing the foor loops :) I have some questions though:

1. Why after guarding the array to have at least 3 elements is not safe to use `array.first!`, but is safe to call `array.removeFirst()` 🤔  is it because is making an entire copy of the array except the first one and then replacing the entire array with the copy (I noticed it has complexity $O(n)$)? In addition, It is returning the first element (as a non-optional) assuming it was safe from the very beginning, isn't?
2. Is it possible somewhere to see the details of the implementation of this `removeFirst()` or is this hidden?  🥷

Sherlouk: This comment is less about the safety, and more about the readability and reducing the need for many loops and creating array builders (which requires more memory).

As you say, we already have the guard to make sure there are enough elements so this will be safe whether it's with force unwraps (!) or 'unsafe' functions such as removeFirst.

The implementation is not public 🙂

Sources/App/Core/PackageContributors.swift
Sherlouk: The `Current.git.shortlog` function has a do/catch in it, which means regardless of the actual error which is thrown, the same 'invalidString' error is caught in the function this review comment is attached to.

You're including the `error.localizedDescription` in the `shellCommandFailed` but this will always be 'invalidString', rather than the actual error from the shell command.

My proposal is that we get rid of the do/catch in the shortlog function itself (not where this comment is). This means the shellCommandFailed error message will contain the actual reason for the error.

Sources/App/Core/AuthorMetadata.swift
```diff
@@ -0,0 +1,26 @@
+// Copyright 2020-2022 Dave Verwer, Sven A. Schmidt, and other contributors.
```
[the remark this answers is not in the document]
daveverwer: I removed this file.

Sherlouk (approved):
Fantastic work @jcubit!

daveverwer:
Congratulations @jcubit. This is a wonderful contribution that you should be proud of! 🎉

Thank you so much. It's deploying to staging now, where we can test it.

finestructure:
Congratulations, @jcubit ! Well done 🎉

PR #2063, opened 2022-10-07
Handle documentation URL paths (catchall)

### LINKED ISSUE #2061: Handle default documentation link catchall
The following url should work:

https://swiftpackageindex.com/vapor/postgres-nio/documentation/postgresnio/postgresconnection

and resolve to

https://swiftpackageindex.com/vapor/postgres-nio/1.11.1/documentation/postgresnio/postgresconnection

(or whatever the latest release is instead of 1.11.1)

finestructure:
Fixes #2061 

- The diff looks a lot messier than it is – it's mostly moving logic around
	- document target finding queries moved into `DocumentationTarget.query`
	- logic previously attached to `PackageResult` moved into `DocumentationTarget` because it's now using `Joined3` (`DocumentationTarget.query`) instead of `Joined5` 
	- existing `DocumentationVersion` into own file (because it gains a `query` method, making it more than just a container struct)
	- factor out some code in `PackageController+routes` to use these new `query` methods and the `DocumentationTarget` type
	- NB: there is no change in functionality - all redirect remain redirects, we run the same number of queries (although using `DocumentationTarget.query` instead of `PackageResult.query` means it's a slightly lighter query)
	- documentation URL generation now in `SiteURL.relativeURL`
- Only new functionality is catchall handling for internal and external doc links + tests

Todo:

- [x] add tests for internal catchall urls
- [x] add test for `**` default docs route
- [x] test `PackageResult.hasDocumentation`
- [x] test `PackageResult.documentationTarget`
- [x] add tests for various catchall urls
- [x] test `#fragment` paths


finestructure:
Ok, this is now good to go. It's gotten a bit bigger, because I ended up cleaning up a bit of my initial docs-proxy mess while I was at it...

Sources/App/Controllers/PackageController+routes.swift
```diff
@@ -83,29 +83,27 @@ enum PackageController {
...
+        let fragment = req.url.fragment.map { "#\($0)"} ?? ""
+        let path = req.parameters.getCatchall().joined(separator: "/") + fragment
 
-        let result = try await PackageResult.query(on: req.db, owner: owner, repository: repository)
-        if let documentationUrl = result.defaultDocumentationUrl {
-            throw Abort.redirect(to: documentationUrl)
-        } else {
+        guard let target = try await DocumentationTarget.query(on: req.db,
+                                                               owner: owner,
+                                                               repository: repository)
```
finestructure, line 98: This is probably the core change - push all the doc target queries into new type's `query` method so we can keep it all together a bit better and add some more focused tests.

Sources/App/Controllers/PackageController+routes.swift
```diff
@@ -117,33 +115,18 @@ enum PackageController {
...
-        else { throw Abort(.notFound) }
+        let referenceToMatch: Reference = SemanticVersion(reference)
+            .map { .tag($0, reference) } ?? .branch(reference)
 
-        // This package has at least one docArchive, so redirect to it.
-        guard let docArchive = queryResult.model.docArchives?.first
+        guard let target = try await DocumentationTarget.query(on: req.db,
+                                                               owner: owner,
+                                                               repository: repository,
+                                                               reference: referenceToMatch)
```
finestructure, line 124: dto `Joined3.query` → `DocumentationTarget.query`

Sources/App/Controllers/PackageController+routes.swift
```diff
@@ -159,96 +142,22 @@ enum PackageController {
...
-                    )
-                }
-
-                return try await processor.processedPage.encodeResponse(
-                    status: .ok,
-                    headers: req.headers.replacingOrAdding(name: .contentType,
-                                                           value: fragment.contentType),
-                    for: req
+                let documentationVersions = try await DocumentationVersion
+                    .query(on: req.db, owner: owner, repository: repository)
```
finestructure, line 150: dto `Joined3.query` → `DocumentationTarget.query`

daveverwer (approved):

PR #2069, opened 2022-10-12
Periphery scan

finestructure:


daveverwer (approved):

Sources/App/Core/Query+Support/JoinedQueryBuilder.swift
```diff
@@ -24,6 +24,7 @@ struct JoinedQueryBuilder<J: ModelInitializable> {
         return self
     }
 
+    // periphery:ignore
```
finestructure, line 27: Leaving these unused ones in here, because they're tricky to write and handy to have around when we write new queries. It's something you don't want to have to do when you're working on a query.

Sources/App/Core/Gitlab.swift
```diff
@@ -139,6 +139,7 @@ extension Gitlab.Builder {
         case success
     }
 
+    // periphery:ignore
```
finestructure, line 142: Periphery seems to have trouble detecting the use of types that are only used in decoding.

Sources/App/Core/RecentRelease.swift
```diff
@@ -65,7 +66,7 @@ extension RecentRelease {
     }
     
     static func fetch(on database: Database,
-                      limit: Int = Constants.recentPackagesLimit,
+                      limit: Int = Constants.recentReleasesLimit,
```
finestructure, line 69: Interesting bug discovered by periphery: we were using the wrong constant (although the value is 7 for both of them).

Sources/App/Core/Ref.swift
```diff
@@ -44,16 +44,19 @@ extension Version: Referenceable {}
 ///                              |
 ///                              '-< Product
 /// ```
+// periphery:ignore
```
finestructure, line 47: As above, currently unused but keeping it around for when we write new queries. These are useful tools to have around.

Sources/App/Core/Rollbar.swift
```diff
@@ -41,9 +41,12 @@ enum Rollbar {
 
 extension Rollbar {
     struct Item: Content {
+        // periphery:ignore
```
finestructure, line 44: Again `Codable` related false positives.

PR #2077, opened 2022-10-17
Fix 1793

### LINKED ISSUE #1793: Bump package score if a package has documentation

daveverwer:
Fixes #1793

It may one day be worth making analysis use the `PackageController.PackageResult` rather than pulling the versions out in `compute`, but as discussed, today is not that day.


Sources/App/Core/Score.swift
```diff
@@ -82,14 +87,21 @@ enum Score {
             let defaultVersion = versions.latest(for: .defaultBranch),
             let repo = package.repository
         else { return 0 }
-
+        
+        let hasDocumentation = [
+            defaultVersion,
+            versions.latest(for: .release),
+            versions.latest(for: .preRelease)
+        ].compactMap { $0 }.documentationTarget() != nil
```
finestructure: I've just had a look and realised that we can push `hasDocumentation()` up from `PackageResult` to `[Version]` and then simply reuse it here. `PackageResult.documentationTarget()` isn't actually being used anywhere so this is even simpler overall.

I'll push this in a PR to clarify.

finestructure:
Does that make sense or shall I redo that as a PR?

finestructure (approved):

daveverwer:
Makes perfect sense, thanks!

PR #2081, opened 2022-10-18
Remove incompatibleWithAppStore branding for LGPL*

marcprux:
LGPL licenses are not incompatible with the App Store. There are countless apps that ship with libraries like ffmpeg ([LGPL](https://ffmpeg.org/legal.html)). And nearly every app links, either directly or indirectly, with WebKit and JavaScriptCore ([LGPL](https://webkit.org/licensing-webkit/)).

There was once a potential issue around static linking (see [this ancient thread](https://trac.ffmpeg.org/ticket/1229)), but those concerns were obviated once iOS apps supported dynamic linking in iOS7 in 2013.

### DIFF
--- a/Sources/App/Models/License.swift
+++ b/Sources/App/Models/License.swift
@@ -138,10 +138,7 @@ enum License: String, Codable, Equatable, CaseIterable {
             case .agpl_3_0,
                  .gpl,
                  .gpl_2_0,
-                 .gpl_3_0,
-                 .lgpl,
-                 .lgpl_2_1,
-                 .lgpl_3_0: return .incompatibleWithAppStore
+                 .gpl_3_0: return .incompatibleWithAppStore
             default: return .compatibleWithAppStore
         }
     }


daveverwer:
Thank you for this, Marc.

What you say sounds very reasonable. I might try and get one other person to verify it if you'll allow us a day or two to get another opinion on it.

Thank you!

marcprux:
> What you say sounds very reasonable. I might try and get one other person to verify it if you'll allow us a day or two to get another opinion on it.

Sure, take your time.

I'm starting to think that maybe the `incompatibleWithAppStore` branding should be removed altogether. The reasons being:

1. GPL-style licenses are popular for server-side Linux projects, where Swift is making inroads. Compatibility with any App Store rules is irrelevant for these projects, so packages that use these libraries should not be disadvantaged.
2. Some commonly-used licenses like CC0, Unlicense and WTFL introduce legal hazards in some jurisdictions such as Germany and other EU countries that don't have the same notions of "public domain" as the US. Should projects using these licenses also be flagged?
3. Nowhere does Apple state that GPL-style licenses are prohibited, and I've never heard of any documented cases of an app being rejected because it used any *GPL code.

I don't know the history behind the `incompatibleWithAppStore` tag and what perceived problems it was meant to address, but I feel that calling out certain licenses as a speculative cause for concern ought to be out of scope for a software index, and instead should be left up to the judgement of the developers.

daveverwer:
> I don't know the history behind the `incompatibleWithAppStore` tag and what perceived problems it was meant to address, but I feel that calling out certain licenses as a speculative cause for concern ought to be out of scope for a software index, and instead should be left up to the judgement of the developers.

I can see this argument and have had thoughts along the same lines, but I come down on the side of it being beneficial right now. I think it’s likely that most people assume that everything on a Swift package index would be compatible with the App Store.

For Swift on the Server, I think most people will immediately recognise that any App Store issues don’t apply to server projects and move on. I think it’s clear this only applies to the App Store.

The copy on the package page under the license “Why might the GPL 3.0 be problematic?” is careful to say both “might” and “problematic” rather than stronger words. We could cover that situation much better in the longer copy, though:

> If the package license shows with an orange background but does not say "No License" or "Unknown License", then the package is using a license which is incompatible with the terms and conditions of the App Store, such as any GPL-style license. If you are considering shipping your app to the App Store, you should be aware that using code licensed under one of these licences presents a legal risk.

We could add a sentence about non App Store projects, like Swift on the Server, and also soften wording like “may be incompatible with” and “may present a legal risk”. I think if we do nothing else, we should do this.

Of course, it’s always the developer's choice whether to use any package. We’re just trying to present what we think is relevant information in an obvious way and make people pay attention to licenses so they are aware of what they are doing. I believe it’s a net positive if it makes even a few people think about licensing more.

daveverwer:
> Nowhere does Apple state that GPL-style licenses are prohibited, and I've never heard of any documented cases of an app being rejected because it used any *GPL code.

I forgot to reply to this point.

I don’t think this is about rejection or guidelines. It’s the licenses themselves. As I understand it, the mechanisms of the App Store make it impossible to comply with the terms of some open-source licenses. That’s not Apple’s responsibility, that’s the developer’s responsibility and it is the developer that is at risk of violating the license.

daveverwer:
Thanks once again for this @marcprux. We really appreciate the attention on this.

Unfortunately, after a bit of reading and some [external opinions](https://twitter.com/migueldeicaza/status/1582712070226644993) we're not going to accept this change. We don't want to go back and forth trying to get a definitive answer, but it's clear to us that there are enough questions about this that we should leave these on the list of licenses that encourage people to be cautious with them.

I have softened the wording in our long-form copy with "may" and "might" etc (#2082) and I think we're in a good place.

Thanks again for your contribution.

marcprux:
I beg reconsideration for this change. My rebuttals on @daveverwer's [thread](https://twitter.com/migueldeicaza/status/1582712070226644993) with @migueldeicaza went unanswered, but the upshot is these two points:

1. Many companies are shipping apps with LGPL libraries like [ffmpeg](https://github.com/FFmpeg/FFmpeg/blob/master/COPYING.LGPLv3) ([TuneIn](https://cms.tunein.com/licensing/ios/), to cite one example; I have more that I can list if needed).
2. Apple themselves distributes LGPL libraries (such as [JavaScriptCore](https://webkit.org/licensing-webkit/)) that we all link to in our apps. Every app that uses a web view is linking with LGPL software.

How is it that those highly-lawyered companies all feel comfortable shipping LGPL libraries with their **closed-source** apps, but we are warned away from using an LGPL library in our own **open-source** Swift packages? That wouldn't smell right to me.

Did these orange warning badges initially get implemented due to some specific concerns that were raised by the community? Or were they just a whim? 

I would like to hear more points of view, so I'd be happy to expand this into a discussion topic on whether an impartial catalog should be advantaging certain licenses over others. But I feel that the case for the **L**GPL is, at least, clear and uncontroversial.

daveverwer:
Discussions are always welcome, and any decision we make now will not be final.

I take all of your points and still come down on the side of leaving the “L” licenses in with the rest of the GPL licenses and showing the current text we have.

> highly-lawyered companies

This is part of the problem. If I were a highly-lawyered company, I might feel more comfortable doing something that could in the future be challenged by the author of some LGPL code. If I am a small developer without the time/budget/expertise, I’m going to be grateful that something made me look at the licensing terms more closely to avoid any chance of potential future problems.

It’s still the decision of every developer what code they use. All we should do is highlight things that might affect their decision.

From what I have read, and from that conversation with @migueldeicaza on Twitter, it appears to me that there is still very much a discussion with two sides about whether LGPL code is compatible with the mechanisms of the App Store. 

The fact that there is even a question or discussion around it makes me want to err on the side of caution and encourage people to look carefully at both the “L” licenses as well as the other GPL licenses. What we _really_ want to do is make people think about licensing in _every_ situation, but this feels to me still like the right thing to do given that there are still views on both sides of this “L” issue.

If you would like to propose even more gentle wording over what I have, I think that would make a good [discussion](https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/discussions/new) that might be able to go through to a PR, but I personally feel comfortable with what we have right now.

I’d also like to thank you again for bringing this up. It has been a valuable contribution even if we didn’t accept the fundamental change.

Sherlouk:
Not joining sides here, it's an interesting discussion and I agree that everyone needs to have the facts available to them.

I wonder, Dave, if there's a blog post that you'd consider publishing which explores open source licensing in a bit more detail than a small badge? We could then link to that to help aide developers in making the right decision for them. It could be an opportunity to raise both sides of this discussion.

Just a thought 😊

finestructure:
> If I am a small developer without the time/budget/expertise, I’m going to be grateful that something made me look at the licensing terms more closely to avoid any chance of potential future problems.

This is exactly the first thought I had. As an indie developer I find it very important to see this information to inform my decision.

Moreover, nothing prevents you from still using the package. I'm not sure what the downside is that removing it would correct?

People who don't care about licenses won't care about the warning. People who _do_ care, would be misled if we removed it. To me it's clear that removing the warning is more harmful.

PR #2089, opened 2022-10-21
Update maintainer info with links to new SPIManifest docs

### LINKED ISSUE #864: Improve spi.yml file documentation
We need to fully document the `.spi.yml` somewhere beyond the blog post. In particular, we need to list the platform and swift version strings that we support.

finestructure:
Fixes #864 

CleanShot 2022-10-21 at 12 31 12@2x

### DIFF
--- a/Sources/App/Core/DocumentationTarget.swift
+++ b/Sources/App/Core/DocumentationTarget.swift
@@ -18,6 +18,7 @@ import Fluent
 enum DocumentationTarget: Equatable {
     case external(url: String)
     case `internal`(reference: String, archive: String)
+    case universal
 
     /// Fetch DocumentationTarget for a given package.
     /// - Parameters:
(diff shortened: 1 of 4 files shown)


finestructure:
There's one issue with this: the `#` anchor links aren't working in my testing. The all land on `spimanifest/commonusecases` and not `spimanifest/commonusecases#Images-for-Linux` etc.

This doesn't seem to happen because we're not passing the fragment on. When I debug into this, there _is no_ fragment on `req.url` when visiting these urls. Which is bizarre, because we have a test `test_defaultDocumentation` that tests exactly that successfully.

I haven't figured out why there's this discrepancy.

finestructure:
Breaking out the anchor link problem into a new issue. These links not landing on the exact section should be a deal-breaker for this.

daveverwer (approved):
This is great 👍👍👍

PR #2109, opened 2022-11-01
Fix author analysis

finestructure:


daveverwer (approved):
Thanks so much for persevering with this and finding the issue.

That test is great, too. It passes locally without `HEAD`, but it's great that the CI catches the failure if it ever gets removed, which is great. 👍 Thank you!

jcubit:
Well played @finestructure ! Thank you for finding the issue!

Issue #1888, opened 2022-07-12
Set a flag on a build record indicating if it ran the doc generation step


daveverwer:
Flag will be used by #1887

The flag should be set regardless of whether the doc build succeeded or failed, so we can easily go and check which build it was even for failures.

PR #2110, opened 2022-11-01
Add documentation generation badges

### LINKED ISSUE #1887: Indicate which builds include documentation generation
We should indicate on the [buld results page](https://swiftpackageindex.com/SwiftPackageIndex/SemanticVersion/builds) which build(s) were used to generate docs so that package authors (and us) can easily get to the logs and results of the doc build step.

daveverwer:
Fixes #1887 

The intention here is that the build should show the badge if it even attempted a documentation build so we can find the correct build to start debugging.

I'm not super happy with this wording:

> If successful, this build generated package documentation.

I did consider:

> This build attempted to build documentation.

Here's how it looks:

Screenshot 2022-11-01 at 16 20 27@2x

Screenshot 2022-11-01 at 16 20 41@2x


finestructure:
I must be missing something – I can't figure out where the logic sits that determines which build is supposed to build the docs.

I was expecting to see somewhere something looking at the `SPIManifest` or to have it reported back from the build system (probably better) what build generated the docs.

finestructure:
Oh, I see. This is the display and #1888 is the data :+1:

finestructure (approved):

PR #2115, opened 2022-11-02
fixes contributors format

### LINKED ISSUE #2114: Use thousands separator for formatting the number of contributors when reporting authors
Who would have thought that we'd have a package with more than 1,000 contributors?`

Screenshot 2022-11-02 at 16 47 12@2x

jcubit:
fixes #2114

### DIFF
--- a/Sources/App/Views/PackageController/PackageShow+Model.swift
+++ b/Sources/App/Views/PackageController/PackageShow+Model.swift
@@ -286,9 +286,13 @@ extension PackageShow.Model {
             }
         }
 
-        if authors.numberOfContributors > 0,
-           let numberOfContributors = NumberFormatter.spiDefault.string(from: authors.numberOfContributors) {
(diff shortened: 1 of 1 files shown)


Sources/App/Views/PackageController/PackageShow+Model.swift
```diff
@@ -287,7 +287,9 @@ extension PackageShow.Model {
         }
 
         if authors.numberOfContributors > 0 {
-            nodes.append(.text("\(authors.numberOfContributors) other " + "contributor".pluralized(for: authors.numberOfContributors)))
+            guard let numberOfContributors = NumberFormatter.spiDefault.string(from: authors.numberOfContributors)
+            else { return .empty }
```
daveverwer, line 291: Just a minor suggestion, we could use an `if let` here instead of a `guard let` and prevent the possibility of this ever killing  the whole section. It's very unlikely to ever fail, but might be worth doing?

```suggestion
"\(authors.numberOfContributors)"
```

daveverwer (approved):

Issue #1470, opened 2021-12-29
Change Score.compute to take defaultBranch version and release count


finestructure:
It currently takes all versions but only uses the release count and some default version info. We can make the call into `Score.compute` in `updatePackages` (`Common.swift`) cheaper by not loading all versions.

daveverwer:
Closing this. As part of #1793 we need to bring the latest stable release into the score calculation.

finestructure:
This fix is showing weird performance behaviour. We've deployed this in two parts on `dev`:

- d6216bfe86cbab893a5a68495b4f37b0a6b42826 was just the preparatory changes, mostly ELF → a/a
- cadc77d9e5e36d7f0aafbbc9ea78b2f86af95244 was the change to skip loading versions in ingestion and passing the already computed versions into `updatePackages` in analysis

Here's what's happening on `dev`:

![CleanShot 2022-10-27 at 10 29 46@2x](https://user-images.githubusercontent.com/65520/198233132-9103628d-0bd1-4ebf-98b1-5da12a3cc1cf.png)

I've also seen weird performance after deploying the second part (cadc77d9e5e36d7f0aafbbc9ea78b2f86af95244) yesterday afternoon, so it doesn't seem to be a fluke.

Both times I've seen elevated CPU of the analysis job, which seems to hover around 20-25%.

![CleanShot 2022-10-27 at 10 23 49@2x](https://user-images.githubusercontent.com/65520/198234882-2ed78fde-7d27-449e-a4fa-04183adbae1c.png)

Also, analysis doesn't seem to progress:

![CleanShot 2022-10-27 at 10 24 06@2x](https://user-images.githubusercontent.com/65520/198234980-6b329a3f-0bdd-4cb4-b89e-8e480fd7552b.png)

The time of the screenshot was 10:24 and the last analysis log was 10:01. Normally, this should update at least every 20s or so.

finestructure:
What's odd is that the second part of the fix should be the less problematic one 🤷‍♂️

I'll back out of this change now so that we can continue deploying generally and then I'll try and break down this change further and deploy it piece by piece.

I'll also delete the 2.78.16 tag so we don't accidentally deploy it or roll back to it. No 2.78.16 images have been produced yet, so there's no risk of a collision when we re-tag:

```
 docker-user@bastion  d1  ~  docker pull registry.gitlab.com/finestructure/swiftpackageindex:2.78.16                           371ms
[...]

PR #2132, opened 2022-11-10
Swift 5.7.1 base image

finestructure:


.github/workflows/ci.yml
```diff
@@ -23,7 +23,7 @@ jobs:
     name: Test
     runs-on: ubuntu-latest
     container:
-      image: registry.gitlab.com/finestructure/spi-base:0.8.0
+      image: registry.gitlab.com/finestructure/spi-base:0.9.2
```
finestructure, line 26: I wonder if this was the cause for the recent TSAN issues when I first tried to fix #1470: we were still using the 0.8.0 base image with the pre-release 5.7 nightly in our CI!

PR #2135, opened 2022-11-14
Rework #1470 part 1

### LINKED ISSUE #1470: Change Score.compute to take defaultBranch version and release count

finestructure:
Another attempt to bring back the changes addressing #1470. This is part one with the following non-functional changes:

- Common.updatePackages -> a/a
- Common.recordError -> a/a
- AppEnvironment.reportError -> a/a
- AppError.report -> a/a
- Rollbar.createItem -> a/a
- Add ActorIsolated to handle assertions with a/a Current functions
- replaced AppTestCase.testQueue with QueueIsolated to handle assertions with sync Current functions
- Converted related tests to a/a, fixed up Actor/QueueIsolated

CI failed twice [with the following error](https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/actions/runs/3470758552/jobs/5799361127) (as we've seen previously with this change, before I rolled everything back):

```
    ✖ test_analyze, XCTAssertEqual failed: ("33") is not equal to ("34") -
    ✖ test_analyze, failed - Snapshot does not match reference.
```

I've since run it another 8 times in CI without errors. It never fails on macOS. I cannot get it to fail on `lb1`.

I feel like the changes are important enough to pull in even if we don't manage to properly fix #1470 and even if there's a chance for a race condition in a single test. If this turns out to become a nuisance, perhaps there are ways to tweak this test to mitigate that.

We'll need to go a/a eventually so this can't be a fundamental blocker for us doing so.

The other issue we were seeing with #1470 changes was analysis/ingestion stalling. I'll deploy from the branch to dev so we can see how it goes over night.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -33,10 +33,8 @@ build:
 run:
 	swift run
 
-test: xcbeautify
-	set -o pipefail \
-	&& swift test --disable-automatic-resolution --sanitize=thread \
-	2>&1 | ./xcbeautify
+test:
+	swift test --disable-automatic-resolution --sanitize=thread
 
 test-query-performance: xcbeautify
 	set -o pipefail \
--- a/Makefile
+++ b/Makefile
@@ -33,8 +33,10 @@ build:
 run:
 	swift run
 
-test:
-	swift test --disable-automatic-resolution --sanitize=thread
+test: xcbeautify
+	set -o pipefail \
+	&& swift test --disable-automatic-resolution --sanitize=thread \
+	2>&1 | ./xcbeautify
 
 test-query-performance: xcbeautify
 	set -o pipefail \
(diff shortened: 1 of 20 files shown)


finestructure:
So far so good:

CleanShot 2022-11-16 at 07 00 33@2x

Let's maybe give this until the afternoon if possible (no other deployments to `dev`) and then we should be ok to merge.

finestructure:
This is good to be merged. Once it's in I'll ad-hoc deploy from the part 2 branch for testing.

daveverwer (approved):

PR #2139, opened 2022-11-15
Add top level documentation pages to the sitemap for documented packages

### LINKED ISSUE #2138: Force indexing of top-level package documentation for documented packages
See context in https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/discussions/2078#discussioncomment-4145500

We should link the top-level documentation page for any package with documentation to make sure Google regularly crawls the documentation pages from that starting point.

We should leverage the work we have done on redirects and keep the processing in the sitemap request to a minimum by simply adding the `/documentation` redirect URL to the sitemap.

```xml
<url>
    <loc>https://swiftpackageindex.com/vapor/postgres-nio</loc>
    <changefreq>daily</changefreq>
</url>
<url>
    <loc>https://swiftpackageindex.com/vapor/postgres-nio/documentation</loc>
    <changefreq>daily</changefreq>
</url>
```

If it's easier during implementation, since I think this will require a different query, these additional entries do *not* need to be next to their original package page URL in the map.

**Note:** We should *not* include the `/documentation` path for packages with external documentation. I'm not sure whether there are strict rules for a sitemap pointing to a completely different domain, but it doesn't feel like that's what a sitemap should be used for.

daveverwer:
Fixes #2138

> **Note:** We should _not_ include the `/documentation` path for packages with external documentation. I'm not sure whether there are strict rules for a sitemap pointing to a completely different domain, but it doesn't feel like that's what a sitemap should be used for.

This PR does *not* do this. We had a `has_docs` field already in search, so it was trivial to implement this using that. I'm going to check in on this in a week or so to see if it has had any effect and if it has, we can consider a better fix to this than what we have here, but as a quick experiment, this is the best way.

### DIFF
--- a/Sources/App/Core/SiteURL+siteMap.swift
+++ b/Sources/App/Core/SiteURL+siteMap.swift
@@ -31,13 +31,27 @@ extension SiteURL {
                     .changefreq($0.changefreq)
                 )
             },
-            .forEach(packages) { p in
-                .url(
-                    .loc(
-                        SiteURL.package(.value(p.owner), .value(p.repository), .none).absoluteURL()),
-                    .changefreq(
-                        SiteURL.package(.value(p.owner), .value(p.repository), .none).changefreq)
-                )
+            .forEach(packages) { package in
+                    .group(
+                        .url(
+                            .loc(SiteURL.package(.value(package.owner),
+                                                 .value(package.repository),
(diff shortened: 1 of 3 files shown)


finestructure (approved):

daveverwer:
Deploying as 2.79.9

PR #2142, opened 2022-11-16
Rework #1470 part 2

### LINKED ISSUE #1470: Change Score.compute to take defaultBranch version and release count

finestructure:
Merge after #2135 

Fixes #1470

- Analyze.updateLatestVersion → a/a, returning the updated versions
- `analyze` is calling a new function `updateScore` using the previously loaded `Version`s
- `Common.updatePackages` is not calling `pkg.$versions.load(on: db)` anymore, fixing #1470
- Test updates according to above changes

### DIFF
--- a/Sources/App/Commands/Analyze.swift
+++ b/Sources/App/Commands/Analyze.swift
@@ -168,8 +168,8 @@ extension Analyze {
         }
 
         let packageResults = await withThrowingTaskGroup(
-            of: Joined<Package, Repository>.self,
-            returning: [Result<Joined<Package, Repository>, Error>].self
+            of: (Joined<Package, Repository>, [Version]).self,
+            returning: [Result<(Joined<Package, Repository>, [Version]), Error>].self
         ) { group in
             for pkg in packages {
                 group.addTask {
@@ -182,9 +182,9 @@ extension Analyze {
                                               package: pkg)
                         }
                     }
-                    try result.get()
+                    let versions = try result.get()
 
-                    return pkg
+                    return (pkg, versions)
                 }
             }
 
@@ -208,7 +208,7 @@ extension Analyze {
     static func analyze(client: Client,
                         transaction: Database,
                         logger: Logger,
-                        package: Joined<Package, Repository>) async throws {
+                        package: Joined<Package, Repository>) async throws -> [Version] {
         try refreshCheckout(logger: logger, package: package)
         try await updateRepository(on: transaction, package: package)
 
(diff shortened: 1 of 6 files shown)


daveverwer (approved):

finestructure:
It's flatlining again on this revision:

CleanShot 2022-11-18 at 12 22 34@2x

finestructure:
Deploying ea46568e7024f02de02b526c3e69596dfd9c0c2f to see if that tweak in `updateLatestVersions` makes a difference.

finestructure:
Stuck again:

CleanShot 2022-11-21 at 11 35 35@2x

The good news is this seems to be failing quickly and only for analysis. So even if this isn't fixable, we'd still have a win by avoiding the version fetch in ingestion.

finestructure:
I have a theory why this might be failing in analysis and I'm preparing to test this. We're processing batches of 25 packages and in the live analysis we only send the packages to the update function which refetches the versions for each version. In this updated version we send 25 packages _and all their versions_ on each invocation, avoiding the re-fetch. I haven't actually checked when the process failed but I could imagine that for certain batches we're hitting a RAM limit with this change.

I'll be testing this by reducing the batch size (and increasing the frequency accordingly) to see if this fixes the problem.

finestructure:
Running at batch size 5, sleep 4 is also stalling. Don't think it's that.

finestructure:
Ok, I think I've finally fixed this by moving the score update to analysis entirely. We were doing it as part of `updatePackages` which was called from both ingestion and analysis and yesterday I realised that there's no point actually trying to do it there if we're then adding an overload that updates the score in one case (when called from analysis) and not in the other (when called from ingestion).

I'm not sure _why_ the process stalled when we passed along the versions but this is the better solution anyway. I deployed 6289dcc1a9692b2ce1b28675c918f7a8ed122e81 from the branch this morning and it's been working fine so far:

CleanShot 2022-11-22 at 08 03 25@2x

There's a dramatic improvement in ingestion processing time to a bit less than half.

finestructure:
Ready for review but let's not merge it straight away and give the deployed version more time.

finestructure:
I want to make sure the mon-update/rollback works ok on dev before tagging that version and deploying it. Then I'll merge this when things have settled down.

finestructure:
We have another "lockup" this morning on dev:

CleanShot 2022-11-25 at 08 19 54@2x

However, dev is running b2f22b592e5511dc1223e36ed291851c471d7a82, which is current `main`, i.e. does not include the changes here. So this issue isn't actually related to these changes.

The only difference between b2f22b592e5511dc1223e36ed291851c471d7a82 and 2.79.10, which is running on prod without any issues, is minimal:

[2.79.10..b2f22b592e5511dc1223e36ed291851c471d7a82](https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/compare/2.79.10..b2f22b592e5511dc1223e36ed291851c471d7a82)

PR #2153, opened 2022-11-23
Spiyml authors override

### LINKED ISSUE #174: Allow overriding of author information from the additional metadata file

jcubit:
fixes #174. It has been tested locally with [LeftPad](https://swiftpackageindex.com/daveverwer/LeftPad)

Screenshot 2022-11-23 at 18 45 34

### DIFF
--- a/Sources/App/Views/PackageController/PackageShow+Model.swift
+++ b/Sources/App/Views/PackageController/PackageShow+Model.swift
@@ -281,43 +281,42 @@ extension PackageShow.Model {
     func authorsListItem() -> Node<HTML.ListContext> {
         guard Environment.current == .development else { return .empty }
 
-        guard let authors
-        else {
-            return .empty
-        }
+        guard let authors else { return .empty }
+
         switch authors {
-        case .fromSPIManifest(let spiymlAuthors) :
-            guard var spiymlAuthors
-            else { return .empty }
-            if spiymlAuthors.count > 200 {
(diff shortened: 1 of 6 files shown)


Sources/App/Views/PackageController/PackageShow+Model.swift
```diff
@@ -275,33 +281,44 @@ extension PackageShow.Model {
...
+        guard let authors
+        else {
+            return .empty
+        }
+        switch authors {
+        case .fromSPIManifest(let spiymlAuthors) :
+            guard var spiymlAuthors
+            else { return .empty }
+            if spiymlAuthors.count > 200 {
+                spiymlAuthors = String(spiymlAuthors.prefix(200)) + "..."
```
daveverwer, line 293:
```suggestion
                spiymlAuthors = String(spiymlAuthors.prefix(200)) + "&hellip;"
```

daveverwer (commented):
I'll do a proper review soon, but I just noticed this little one.

daveverwer (approved):
This is great work. Thank you @jcubit!

🚀

PR #2156, opened 2022-11-25
Dev build allow list

finestructure:
This adds a new `BUILD_TRIGGER_ALLOW_LIST` environment variable which allows us to specify packages to be built regardless of any global downscaling settings. I.e. any package in this list will receive build triggers even on `dev`, where we downscale to 5%.

To do before merge:

- [x] configure `BUILD_TRIGGER_ALLOW_LIST` on dev


Sources/App/Commands/TriggerBuilds.swift
```diff
@@ -203,18 +203,20 @@ func triggerBuilds(on database: Database,
...
-    }
-
     return Current.getStatusCount(client, .pending)
         .and(Current.getStatusCount(client, .running))
         .flatMap { (pendingJobs, runningJobs) in
             AppMetrics.buildPendingJobsCount?.set(pendingJobs)
             AppMetrics.buildRunningJobsCount?.set(runningJobs)
             var newJobs = 0
             return packages.map { pkgId in
+                let allowListed = Current.buildTriggerAllowList().contains(pkgId)
```
finestructure, line 213: I initially intended to make the allow list URL based but it turns out we only have package IDs available at this point. So we'll have to seed the allow list with IDs instead.

Since this list is quite nitty-gritty and should see few changes once updated (and package IDs are stable once packages are added), that should be ok for a feature that's to be used in the `dev` environment only.

One advantage this has is that there's not guessing if it's `http://github.com/foo/bar`, `.../bar.git`, `Bar.git`, `https://...` etc but admittedly the initial lookup is a bit more awkward.

As a reminder: the package id is listed in the HTML source of a package page in a comment, so it's easy to look up even without database access.

CleanShot 2022-11-25 at 16 32 17@2x

finestructure: (We could of course add some sort of lookup from `URL` → `Package.Id` inside build trigger but that's quite a bit of hassle and would pretty much force a conversion over to async/await as a first step, which I want to avoid at this point.)

daveverwer (approved):

PR #2175, opened 2022-12-05
Issue 2174

### LINKED ISSUE #2174: Frequent CI failures in `test_analyze`
The race condition in `test_analyze` is getting annoying. It only happens in CI and not consistently. I'd say the rate is somewhere between 5-10% of CI runs. The error is:

```
✖ test_analyze, XCTAssertEqual failed: ("33") is not equal to ("34") -
✖ test_analyze, failed - Snapshot does not match reference.
```

For example: https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/actions/runs/3618702077/attempts/1

Typically clears on re-run.

My previous attempt to cure this with a `Task.yield()` didn't help.

The difficulty is that it is unreproducible anywhere else.

I've tried:

- macOS
- Linux
- Docker on Linux

I suspect the machine's performance plays a role, so maybe trying to slow down the test artificially might help.

There are no TSAN warnings when running this test, nor can I see anything obviously racey about it.

finestructure:
Fixes #2174

### DIFF
--- a/Tests/AppTests/Helpers/QueueIsolated.swift
+++ b/Tests/AppTests/Helpers/QueueIsolated.swift
@@ -30,9 +30,7 @@ public final class QueueIsolated<Value> {
 
     public var value: Value {
         get {
-            _queue.sync {
-                return self._value
-            }
+            _queue.sync { self._value }
         }
     }
 
--- a/Tests/AppTests/Helpers/QueueIsolated.swift
+++ b/Tests/AppTests/Helpers/QueueIsolated.swift
@@ -22,15 +22,23 @@ import Foundation
 public final class QueueIsolated<Value> {
     private var _queue = DispatchQueue(label: "queue-isolated")
 
-    public var value: Value
+    private var _value: Value
 
     public init(_ value: Value) {
-        self.value = value
+        self._value = value
+    }
+
+    public var value: Value {
+        get {
+            _queue.sync {
+                return self._value
+            }
+        }
     }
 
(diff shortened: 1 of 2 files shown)


daveverwer (approved):

Tests/AppTests/Helpers/QueueIsolated.swift
```diff
@@ -22,31 +22,39 @@ import Foundation
...
+        self._value = value
+    }
+
+    public var value: Value {
+        get {
+            _queue.sync {
+                return self._value
+            }
+        }
     }
```
finestructure, line 37: I _think_ this was a real problem. As opposed to the `IsolatedActor`, which I modelled this after, plain access to the property was _not_ guarded by the queue, while on `IsolatedActor` the plain property _is_ guarded by the actor.

I believe that was the mistake and wrapping the property in the queue should fix this.

I.e. this access which is the failing assert:

```swift
        XCTAssertEqual(commands.value.count, 34)
```

raced with the `setValue` in the closure. If I'd written this as

```swift
    commands.withValue {
        XCTAssertEqual($0.count, 34)
    }
```

it wouldn't have raced, either.

We need the `value` attribute accessor in other places, so I've simply fixed that properly.

Fingers crossed 🤞

I'm going to run this a fair few times in CI to see if it reappears.

finestructure:
I think that was it:

CleanShot 2022-12-05 at 14 51 15@2x

PR #2188, opened 2022-12-13
Update to Loki 2.7.1

finestructure:


finestructure:
Deployed manually from the branch and it doesn't appear to have the issue with the annotations anymore:

CleanShot 2022-12-14 at 08 55 40@2x

PR #2196, opened 2022-12-19
Lowercase path for data doc routes

### LINKED ISSUE #2168: NIO's additional targets not in docs
New targets were added to the SPI manifest: https://github.com/apple/swift-nio/pull/2329

but they don't appear in the docs yet:

https://swiftpackageindex.com/apple/swift-nio/main/documentation/NIOCore is a 404.

finestructure:
Fixes #2168

### DIFF
--- a/Sources/App/Controllers/PackageController+routes.swift
+++ b/Sources/App/Controllers/PackageController+routes.swift
@@ -144,15 +144,15 @@ enum PackageController {
         let catchAll = [archive].compactMap { $0 } + req.parameters.getCatchall()
         let path: String
         switch fragment {
-            case .documentation, .tutorials:
-                // DocC lowercases documentation and tutorial URLs. Since these routes can also
+            case .data, .documentation, .tutorials:
+                // DocC lowercases "target" names in URLs. Since these routes can also
                 // appear in user generated content which might use uppercase spelling, we need
-                // to lowercase the input in these cases.
+                // to lowercase the input in certain cases.
                 // See https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/issues/2168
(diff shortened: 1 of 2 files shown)


Sources/App/Controllers/PackageController+routes.swift
```diff
@@ -144,15 +144,15 @@ enum PackageController {
...
+                // DocC lowercases "target" names in URLs. Since these routes can also
                 // appear in user generated content which might use uppercase spelling, we need
-                // to lowercase the input in these cases.
+                // to lowercase the input in certain cases.
                 // See https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/issues/2168
                 // and https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/issues/2172
                 // for details.
                 path = catchAll.joined(separator: "/").lowercased()
-            case .css, .data, .faviconIco, .faviconSvg, .images, .img, .index, .js, .themeSettings:
+            case .css, .faviconIco, .faviconSvg, .images, .img, .index, .js, .themeSettings:
```
finestructure, line 155: I tried making the `lowercased()` apply for all fragments but we've actually got a test in place in `test_documentation_images` that asserts:

```swift
        try app.test(.GET, "foo/bar/1.2.3/images/BAZ.png") {
            // validation
            XCTAssertEqual($0.status, .ok)
            XCTAssertEqual($0.content.contentType?.description, "application/octet-stream")
            XCTAssertEqual($0.body.asString(), "/foo/bar/1.2.3/images/BAZ.png")
        }
```

and which fails if we do that. So at the very least `.images` we'd need to leave alone.

But it's probably safer to do this bit-by-bit in case other fragments needs this rather than risk breaking more with a big change.

daveverwer (approved):

PR #2207, opened 2023-01-04
Fixes the HTML XSS vulnerability

### LINKED ISSUE #1409: Speech mark splits search input
**Please describe the bug**

If your search query includes a speech mark, then everything after (and including) the speech mark is then seemingly lost in the search box.

1. Navigate to to https://swiftpackageindex.com/search?query=test+%22+two
3. See that the 'results for' section includes the full string, but the search box only has the the string before the speech mark.

The search box should include the full query input string.

**Screenshots**

Screenshot 2021-12-01 at 2 05 23 pm

- Platform: Desktop
- Browser: Chrome

### LINKED ISSUE #2206: XSS vulnerability
https://www.openbugbounty.org/reports/3128202/

daveverwer:
Fixes #2206 
Fixes #1409

This ended up fixing #1409 too, as it was exercising the same code in a much less malicious way 😂

### DIFF
--- a/Sources/App/Views/Search/SearchShow+Model.swift
+++ b/Sources/App/Views/Search/SearchShow+Model.swift
@@ -24,8 +24,8 @@ enum SearchShow {
 
         internal init(page: Int, query: String, response: Search.Response, weightedKeywords: [WeightedKeyword]) {
             self.page = page
-            self.query = query
-            self.term = response.searchTerm
+            self.query = query.filteredForDisplay()
+            self.term = response.searchTerm.filteredForDisplay()
             self.filters = response.searchFilters
             self.response = Model.Response(response: response)
             self.weightedKeywords = weightedKeywords
@@ -67,4 +67,17 @@ enum SearchShow {
         }
     }
 
+}
+
+extension String {
+    func filteredForDisplay() -> String {
+        // This does not strip HTML tags as it's possible people would want to search for them.
+        // Instead, it changes angle brackets and quotes and when combined with our SQL sanitisation does everything we need.
+        self.replacingOccurrences(of: ">", with: "&gt;")
+            .replacingOccurrences(of: "<", with: "&lt;")
+            .replacingOccurrences(of: "'", with: "&apos;")
+            .replacingOccurrences(of: "\"", with: "&quot;")
+        // Also, the `term` coming back out of the view model may contain backslash escape characters which we should not display.
+            .replacingOccurrences(of: "\\", with: "")
+    }
 }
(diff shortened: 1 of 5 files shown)


Sources/App/Views/Search/SearchShow+Model.swift
```diff
@@ -68,3 +68,16 @@ enum SearchShow {
...
+
+extension String {
+    func filteredForDisplay() -> String {
+        // This does not strip HTML tags as it's possible people would want to search for them.
+        // Instead, it changes angle brackets and quotes and when combined with our SQL sanitisation does everything we need.
+        self.replacingOccurrences(of: ">", with: "&gt;")
+            .replacingOccurrences(of: "<", with: "&lt;")
+            .replacingOccurrences(of: "'", with: "&apos;")
+            .replacingOccurrences(of: "\"", with: "&quot;")
+        // Also, the `term` coming back out of the view model may contain backslash escape characters which we should not display.
```
daveverwer, line 80: This is actually an unrelated fix. On the current site, if you search for anything that gets sanitised as part of `Search.sanitize`, the escape character was being echoed back to the output in HTML.

Unrelated fix, but worth doing at the same time!

Sources/App/Views/Search/SearchShow+Model.swift
```diff
@@ -24,8 +24,8 @@ enum SearchShow {
 
         internal init(page: Int, query: String, response: Search.Response, weightedKeywords: [WeightedKeyword]) {
             self.page = page
-            self.query = query
-            self.term = response.searchTerm
+            self.query = query.filteredForDisplay()
+            self.term = response.searchTerm.filteredForDisplay()
```
finestructure, line 28: I think it would be important to have some regression tests for this in place, in particular the actual offending string and what it looks like after.

finestructure: Just on the SearchShow init should be enough as it's the general entrypoint I think.

daveverwer: Good idea. I made it a snapshot test because the exploit injected an SVG which also broke the layout so it should verify that nothing changes visually, too.

Sources/App/Views/Search/SearchShow+Model.swift
```diff
@@ -68,3 +68,16 @@ enum SearchShow {
     }
 
 }
+
+extension String {
+    func filteredForDisplay() -> String {
```
finestructure, line 73: Nit: I'd call this `sanitized()` instead, matching out `Search.sanitize()` function.

finestructure (approved):

daveverwer:
Deploying as 2.79.26

PR #2212, opened 2023-01-09
Testing cla-bot

daveverwer:


Sherlouk:
let me know if you need another test subject 👀

daveverwer:
Thanks, James! Unfortunately I need one who hasn't signed a CLA!

finestructure:
Do we need to/want to disable this for the package list repository? It's showing up there (although it doesn't actually prevent merges - not sure if it would/should for other repos).

https://github.com/SwiftPackageIndex/PackageList/pull/3343

CleanShot 2023-01-10 at 10 26 11@2x

daveverwer:
Yes, this is no good. I don't see anything obvious for "all repositories except ..." in the documentation, so I'll disable it all for now.

The alternative is to manually enable it for every repo apart from PackageList, but that's no good as every time we add someone, we'd need to update *all* repository's allow lists.

Sherlouk:
Is there a way to whitelist the GitHub Actions bot?

Sherlouk:
Equally, we don't use it but there is an ability to specify the account used to create the automated PackageList PRs. We could update it to use an account which has signed the CLA (be it you or Sven)?

Just throwing ideas out.

daveverwer:
That works great. Thank you James.

Sherlouk:
Teamwork!

PR #2218, opened 2023-01-17
Issue 2217

### LINKED ISSUE #2217: link from twitter announcement of new package results in package not found
**Please describe the bug**

I follow the twitter feed for SPI on new packages. The sentry update was announced at https://twitter.com/SwiftPkgUpdates/status/1615057390272974873, and included a link to the site and page:
https://swiftpackageindex.com/getsentry/sentry-cocoa#releases

However, that link reports "Package not found".

The same links are provided on the main page currently for "recent releases". I tried a search for Sentry to see if was in there under a slightly different name, but to no avail...

1. Navigate to to '[twitter announcement](https://twitter.com/SwiftPkgUpdates/status/1615057390272974873)'
2. Click on link to package
3. See an error: Package Not Found

Link to a package index entry and details about Sentry

- Platform: iOS/iPad - NetNewsWire, Twitter, and direct Safari
- Browser: Safari
- Version: iOS 16.2

finestructure:
Fixes #2217 

I've checked and we can't really loosen our reliance on default branch presence that I contemplated in the issue. The underlying query is difficult to adjust and given how rare and typically short-lived this problem is (even before mitigation) it's not really worth the effort.

This change prevents throttling of default branches after name changes and should limit the exposure to a couple of hours at worst.

One other remedy we could seek if this continues to be an issue is to prioritise packages with default branch name changes for analysis, like we do with new packages. However, we can't simply make them "new" packages to piggy back on the existing priority, because it would also announce it as a new package (home page, social).

We'd have to introduce an additional priority, making this a bigger change, and one we shouldn't make at this time.

We can test this patch by deploying from the branch to `dev`, which has the same data problem as `prod`.

### DIFF
--- a/Sources/App/Commands/Analyze.swift
+++ b/Sources/App/Commands/Analyze.swift
@@ -389,7 +389,7 @@ extension Analyze {
         let incoming = try await getIncomingVersions(client: client, logger: logger, package: package)
 
         let throttled = throttle(
-            lastestExistingVersion: existing.latestBranchVersion,
+            latestExistingVersion: existing.latestBranchVersion,
             incoming: incoming
         )
         let origDiff = Version.diff(local: existing, incoming: incoming)
@@ -446,8 +446,8 @@ extension Analyze {
     }
 
 
-    static func throttle(lastestExistingVersion: Version?, incoming: [Version]) -> [Version] {
-        guard let existingVersion = lastestExistingVersion else {
+    static func throttle(latestExistingVersion: Version?, incoming: [Version]) -> [Version] {
+        guard let existingVersion = latestExistingVersion else {
             // there's no existing branch version -> leave incoming alone (which will lead to addition)
             return incoming
         }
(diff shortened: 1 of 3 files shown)


Sources/App/Commands/Analyze.swift
```diff
@@ -459,11 +459,20 @@ extension Analyze {
...
+            resultingBranchVersion = incomingVersion
+        } else {
+            // if existing version isn't older than our "window", keep it - otherwise
+            // use the latest incoming version
+            if ageOfExistingVersion < Constants.branchVersionRefreshDelay {
+                resultingBranchVersion = existingVersion
+            } else {
+                resultingBranchVersion = incomingVersion
+            }
+        }
```
finestructure, line 475: There's a bit of noise due to also fixing a typo. The essential change is in commit 710522880debfa70d5c8a06e69f37049d6b9dc76

finestructure:
It's back on `dev` after the fix:

CleanShot 2023-01-17 at 13 50 23@2x

daveverwer (approved):

Issue #2227, opened 2023-01-22
"Received signal 11" in app_analyze


finestructure:
I finally caught a glimpse as to why analysis sometimes hangs:

```
2023-01-21T18:57:54.937913401Z app_analyze.1.t1dix7fv4zph@p3    | [ INFO ] Analyzing (limit: 25) ... [component: analyze]
2023-01-21T18:57:54.983740764Z app_analyze.1.t1dix7fv4zph@p3    | [ INFO ] Checkout directory: /checkouts [component: analyze]
2023-01-21T18:57:54.987633060Z app_analyze.1.t1dix7fv4zph@p3    | [ INFO ] pulling https://github.com/yonaskolb/Stringly.git in /checkouts/github.com-yonaskolb-stringly [component: analyze]
2023-01-21T18:57:55.021965033Z app_analyze.1.t1dix7fv4zph@p3    | [ INFO ] pulling https://github.com/dagronf/DSFActionBar.git in /checkouts/github.com-dagronf-dsfactionbar [component: analyze]
2023-01-21T18:57:55.029915226Z app_analyze.1.t1dix7fv4zph@p3    | [ INFO ] pulling https://github.com/ordo-one/package-frostflake.git in /checkouts/github.com-ordo-one-package-frostflake [component: analyze]
2023-01-21T18:57:55.271462231Z app_analyze.1.t1dix7fv4zph@p3    | [ INFO ] pulling https://github.com/chojnac/Dumpling.git in /checkouts/github.com-chojnac-dumpling [component: analyze]
2023-01-21T18:57:55.273393230Z app_analyze.1.t1dix7fv4zph@p3    | [ INFO ] pulling https://github.com/raymccrae/swift-jsonpatch.git in /checkouts/github.com-raymccrae-swift-jsonpatch [component: analyze]
2023-01-21T18:57:55.277729726Z app_analyze.1.t1dix7fv4zph@p3    | Received signal 11. Backtrace:
2023-01-21T18:58:04.984287787Z app_analyze.1.t1dix7fv4zph@p3    | [ ERROR ] Connection request timed out. This might indicate a connection deadlock in your application. If you're running long running requests, consider increasing your connection timeout. [component: server, database-id: psql]
[...]

PR #2230, opened 2023-01-23
Add backtrace

finestructure:
Merge after #2229


daveverwer (approved):

Sherlouk:
Doesn't the initialiser to `Application` already install Backtrace as the first thing it does? Surprised (given the issue this supposedly addresses) that moving this forward two lines of code has such a dramatic impact.

finestructure:
> Doesn't the initialiser to `Application` already install Backtrace as the first thing it does? Surprised (given the issue this supposedly addresses) that moving this forward two lines of code has such a dramatic impact.

Ah, is Backtrace already a dependency? Just realised Package.resolved didn’t change.

Sherlouk:
Yup - Vapor itself adds it. They set it up right at the beginning 👍 

(just checked because this PR thought they changed it aha)

finestructure:
Sadly this is not fixing #2227 then. I'm not sure why the backtrace didn't show up in the logs.

finestructure:
Thanks for spotting this, James!

PR #2234, opened 2023-01-26
Pluralization tweaks

finestructure:
Merge after #2233 

Small clean-up to move the "pluralized" top level function into String and DefaultStringInterpolation extensions.

### DIFF
--- a/Sources/App/Commands/TriggerBuilds.swift
+++ b/Sources/App/Commands/TriggerBuilds.swift
@@ -262,7 +262,11 @@ func triggerBuildsUnchecked(on database: Database,
                             logger: Logger,
                             triggers: [BuildTriggerInfo]) -> EventLoopFuture<Void> {
     triggers.flatMap { trigger -> [EventLoopFuture<Void>] in
-        logger.info("Triggering \(trigger.pairs.count) builds for package name: \(trigger.packageName), ref: \(trigger.reference)")
+        if let packageName = trigger.packageName, let reference = trigger.reference {
+            logger.info("Triggering \(pluralizedCount: trigger.pairs.count, singular: "build") for package name: \(packageName), ref: \(reference)")
+        } else {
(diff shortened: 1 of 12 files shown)


daveverwer (approved):

Sources/App/Controllers/PackageController+ShowRoute.swift
```diff
@@ -87,10 +87,10 @@ extension PackageController {
                     return nil
                 }
                 let cl = Link(
-                    label: pluralizedCount(commitCount, singular: "commit"),
+                    label: "commit".pluralizedCount(commitCount),
```
finestructure, line 90: I think in case of the interpolation `"\(pluralizedCount: ...)"` I'm happy with with new version but here it's a downside that the count comes after the label. In that respect this is a regression.

Probably not worth spending a lot of time on it but perhaps this could be an extension on `Int` to enable

```swift
commitCount.labelled("commit")
```

Sources/App/Commands/TriggerBuilds.swift
```diff
@@ -262,7 +262,11 @@ func triggerBuildsUnchecked(on database: Database,
                             logger: Logger,
                             triggers: [BuildTriggerInfo]) -> EventLoopFuture<Void> {
     triggers.flatMap { trigger -> [EventLoopFuture<Void>] in
-        logger.info("Triggering \(trigger.pairs.count) builds for package name: \(trigger.packageName), ref: \(trigger.reference)")
+        if let packageName = trigger.packageName, let reference = trigger.reference {
+            logger.info("Triggering \(pluralizedCount: trigger.pairs.count, singular: "build") for package name: \(packageName), ref: \(reference)")
+        } else {
+            logger.info("Triggering \(pluralizedCount: trigger.pairs.count, singular: "build") for version ID: \(trigger.versionId)")
+        }
```
finestructure, line 269: This little rabbit hole opened up when fixing the "builds" label here and regretting that `pluralizedCount` was a top level function.

PR #2245, opened 2023-02-01
Adding the Supporters page

daveverwer:
This PR is not complete. Opening it to get initial feedback on the design/layout.

Screenshot 2023-02-01 at 21 33 20@2x


finestructure:
That looks great!

Sources/App/Views/Supporters/SupportersShow+Model.swift
```diff
@@ -0,0 +1,86 @@
...
+
+enum SupportersShow {
+    struct Model {
+        func corporateSupporters() -> Node<HTML.BodyContext> {
+            .section(
+                .class("corporate"),
+                .h3("Corporate Supporters"),
+                .ul(
+                    .group(
+                        Supporters.corporate.shuffled().map(\.listNode)
```
finestructure, line 25: We probably need to use `Current` here to control randomness in snapshots, or am I missing how that’s controlled?

finestructure (approved):

Sources/App/Views/Supporters/SupportersShow+Model.swift
```diff
@@ -0,0 +1,86 @@
...
+            )
+        }
+
+        func communitySupporters() -> Node<HTML.BodyContext> {
+            .section(
+                .class("community"),
+                .h3("Community Supporters"),
+                .ul(
+                    .group(
+                        Supporters.community.shuffled().map(\.listNode)
```
Sherlouk, line 37:
```jameslang
sorted by { if name is james sherlock { then move to front } else { screw them } }
```

Sources/App/Views/Supporters/SupportersShow+Model.swift
daveverwer: The mock repeats the same data repeated, which tests the layout without difficulties caused by `shuffled` and also without including real profiles in our tests.

daveverwer:
Final designs

Screenshot 2023-02-07 at 15 18 11@2x
Screenshot 2023-02-07 at 15 19 10@2x

daveverwer:
I'm going to merge this here as it's a big enough PR already. The next PR will remove the old sponsor CTA and link up this new page.

PR #2246, opened 2023-02-02
Doc report api

finestructure:
This adds the doc reporting API

```
POST /api/builds/{buildId}/docReport
```

I've also changed the build reporting API to be in line with this from

```
POST /api/versions/{versionId}/buildReport
```

This API used to be

```
POST /api/versions/{version}/builds
```

when it actually _created_ builds when the endpoint was hit. This has been different for a while now since we are creating the `Build` records on trigger and the the build reporting merely updates the record.

I think not going with "pure" REST endpoints makes more sense here because the reports do a little more than simply update the records. They also update relationships and so the dedicated `xxxReport` endpoint probably makes sense.

NB: I've left the old

```
POST /api/versions/{version}/builds
```

in place (with deprecation warnings) for now so we don't need to worry about deployment order. I'll take it out once the `builder` has transitioned.

### DIFF
--- a/Sources/App/Controllers/API/API+BuildController.swift
+++ b/Sources/App/Controllers/API/API+BuildController.swift
@@ -86,7 +86,24 @@ extension API {
         }
 
         static func docReport(req: Request) async throws -> HTTPStatus {
-            .noContent
+            let dto = try req.content.decode(PostDocReportDTO.self)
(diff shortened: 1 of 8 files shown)


finestructure:
Ah yes, absolutely. I meant to bring this up which case we should be using. I'll change them!

PR #2277, opened 2023-02-28
Issue 1888 doc build flag

### LINKED ISSUE #1888: Set a flag on a build record indicating if it ran the doc generation step

finestructure:
Fixes #1888 

CleanShot 2023-02-28 at 12 08 35@2x


Sources/App/Controllers/PackageController+BuildsRoute.swift
```diff
@@ -52,20 +52,19 @@ extension PackageController {
...
+            var docStatus: DocUpload.Status?
 
             static func query(on database: Database, owner: String, repository: String) async throws -> [BuildInfo] {
-                try await Joined4<Build, Version, Package, Repository>
-                    .query(on: database)
-                    .filter(Version.self, \Version.$latest != nil)
-                    .filter(Repository.self, \.$owner, .custom("ilike"), owner)
-                    .filter(Repository.self, \.$name, .custom("ilike"), repository)
+                try await Joined5<Build, Version, Package, Repository, DocUpload>
+                    .query(on: database, owner: owner, repository: repository)
```
finestructure, line 59: This means an additional join, and not only on the builds page but also on the package-show page, which reuses this query for the matrix.

I've tested the query with `EXPLAIN ANALYZE` locally and the perf hit is negligible due to the cardinality of max 1 in the join.

I also looked into adding a perf test for that query to our perf test suite but that's not straightforward for Fluent queries. We'd need to find a way to render them into SQL which we can then run through `EXPLAIN ANALYZE`.

Sources/App/Views/PackageController/Builds/BuildIndex+Model.swift
```diff
@@ -89,20 +88,21 @@ extension BuildIndex.Model {
         var platform: App.Build.Platform
         var status: App.Build.Status
         var swiftVersion: App.SwiftVersion
-        var generatedDocs: Bool
+        var docStatus: DocUpload.Status?
 
         var isCompleted: Bool { status.isCompleted }
+        var isDocBuild: Bool {  docStatus != nil }
```
finestructure, line 94: I've fed through the actual doc status instead of coercing into a `Bool` early on. That way we can choose to make the icon carry status info as well if we so choose later on.

I've also renamed to `isDocBuild` instead of `generatedDocs`, because we can't actually be sure that docs have been generated.

daveverwer (approved):
👍

PR #2289, opened 2023-03-09
Path encode reference when constructing S3 urls

### LINKED ISSUE #2287: Doc build fails for `SelectionKit` reference `feature/2.0.0`
Umbrella issue to track the full fix

- [x] doc uploader: https://github.com/SwiftPackageIndex/DocUploader/pull/11
- [x] builder: [MR 186](https://gitlab.com/finestructure/swiftpackageindex-builder/-/merge_requests/186)
- [x] re-trigger doc build
- [x] server url

finestructure:
I'll manually deploy from the branch to test this on dev.

Fixes #2287

To do:

- [x] add tests

### DIFF
--- a/Tests/AppTests/ReferenceTests.swift
+++ b/Tests/AppTests/ReferenceTests.swift
@@ -34,6 +34,30 @@ class ReferenceTests: XCTestCase {
         }
     }
 
+    func test_Equatable() throws {
+        // setup
+        let b1 = Reference.branch("b/1")
+        let b2 = Reference.branch("b/2")
+        let b1dash = Reference.branch("b-1")
+        let b2dash = Reference.branch("b-2")
+        let t1 = Reference.tag(.init("1.0.0")!)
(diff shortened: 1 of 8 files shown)


daveverwer (approved):

Tests/AppTests/ReferenceTests.swift
```diff
@@ -34,6 +34,30 @@ class ReferenceTests: XCTestCase {
         }
     }
 
+    func test_Equatable() throws {
```
finestructure, line 37: I'm having second thoughts whether pushing `pathEncoded` all the way into `==` is a good idea. It may be too broad and lead to unrelated issues if users were to have distinct `feature/a` and `feature-a` branches.

It's probably better to restrict this widened equality only to the area of documentation url/target matching.

Will revisit.

finestructure:
This is good to go now, tested against dev db-dump.

PR #2313, opened 2023-03-30
Swift 5.8

finestructure:


daveverwer (approved):

Package.swift
```diff
@@ -26,12 +26,11 @@ let package = Package(
         .package(url: "https://github.com/JohnSundell/Plot.git", from: "0.10.0"),
         .package(url: "https://github.com/MrLotU/SwiftPrometheus.git", from: "1.0.0-alpha"),
         .package(url: "https://github.com/SwiftPackageIndex/DependencyResolution", from: "1.0.0"),
-        .package(url: "https://github.com/SwiftPackageIndex/SPIManifest", from: "0.10.1"),
+        .package(url: "https://github.com/SwiftPackageIndex/SPIManifest", from: "0.17.0"),
         .package(url: "https://github.com/SwiftPackageIndex/SemanticVersion", from: "0.3.0"),
         .package(url: "https://github.com/SwiftPackageIndex/ShellOut.git",
                  revision: "db112a2104eae7fa8412ea80210d0f60b89a377e"),
-        .package(url: "https://github.com/apple/swift-package-manager.git", branch: "release/5.7"),
-        .package(url: "https://github.com/handya/OhhAuth.git", from: "1.4.0"),
```
finestructure, line 34: We need to inline OhhAuth, because it pins swift-crypto to < 2.0.0 while swift-package-manager needs >= 2.0.0.

We could also ditch Twitter integration entirely, which is the only thing using OhhAuth (to sign the requests).

```
Failed to resolve dependencies Dependencies could not be resolved because root depends on 'ohhauth' 1.4.0..<2.0.0 and 'swift-package-manager' depends on 'swift-crypto' 2.2.3..<2.3.0.
'ohhauth' >= 1.4.0 practically depends on 'swift-crypto' 1.0.0..<2.0.0 because 'ohhauth' 1.4.0 depends on 'swift-crypto' 1.0.0..<2.0.0 and no versions of 'ohhauth' match the requirement 1.4.1..<2.0.0.
```

Makefile
```diff
@@ -35,7 +35,7 @@ run:
 
 test: xcbeautify
 	set -o pipefail \
-	&& swift test --disable-automatic-resolution --sanitize=thread \
+	&& swift test --disable-automatic-resolution \
```
finestructure, line 38: Getting new TSAN errors, reminder to re-enable set here: #2321

PR #2315, opened 2023-03-31
Missing Package.resolved

finestructure:
Should have been part of #2314


daveverwer (approved):

finestructure:
Actually, #2314 was correct since we're still building CI on 5.7. This diff would only come in once we update to the new 5.8 base image.

Keeping it open for now as a reminder. This will probably fold into #2313

PR #2342, opened 2023-04-17
Validate spi yml

### LINKED ISSUE #1782: No way for package authors to detect a syntax error in their `.spi.yml` file
I had an email from the author of [bytes](https://swiftpackageindex.com/tbointeractive/bytes) asking why their documentation hadn't shown up and I noticed an [indentation error](https://github.com/tbointeractive/bytes/pull/55) (which was actually my fault, since the early draft of the blog post was wrong!) in the YML.

Analysis hadn't picked up the manifest and so the `spi_manifest` field was `NULL`.

We should add something to the "Information for package authors" page that says whether there's a parsing error or if any key names are incorrect and maybe also show a "What we understand from your manifest" type section where we humanise the information we have extracted.

finestructure:
Fixes #1782 

I've spent very little time trying to make this look nice but it's a functional checker.

It's a bit unfortunate that the error messages in the failure case will tend to be a bit cryptic but I don't see how we'd fix that unless we used something other than decoding the input to validate it. It's not great but it should give enough information to users to figure out what's wrong.

CleanShot 2023-04-17 at 13 25 16@2x

CleanShot 2023-04-17 at 13 24 58@2x


daveverwer:
Ah this is great! 👍

I can add a little style to it. I'll take a look today.

daveverwer:
OK, I did some work on this and it’s now a “regular” form (well, a Turbo form, see the `turbo-frame` tag) and is styled, etc… Screenshots below.

A couple of observations from while I was working on this:

* Key name validation is a little misleading. For example if you put `doc_targets` instead of `documentation_targets`, it says the manifest is valid. It technically is valid YML and a valid manifest, but it gives the wrong impression.
* As you mentioned, the validation messages are awful 😬

I wonder if we should, as well as the error message, print out what we believe the file actually represents? Like we should show what we have discovered for each key?

I also wonder if we should have a more comprehensive default that shows more of the options we support in the file?

Screenshot 2023-04-18 at 19 49 03@2x
Screenshot 2023-04-18 at 19 49 31@2x
Screenshot 2023-04-18 at 19 50 03@2x
Screenshot 2023-04-18 at 19 50 17@2x

finestructure:
I've added the `dump` of the parsed struct to the result. This should help a little with cases where users insert wrong keys that are being ignored.

CleanShot 2023-04-19 at 10 48 22@2x

With respect to the default placeholder, I chose the one that adds only docs because it should be by far the most common `.spi.yml` people are using. There are more examples in the SPIManifest docs, which should take care of covering other keys and properties, I'd say.

finestructure:
I've run the snapshot test a few times but there's a chance that this will flicker if `dump` isn't stable with respect to attribute ordering across platforms/runs.

If that's the case, we can use https://swiftpackageindex.com/pointfreeco/swift-custom-dump to render a stable output (but I wanted to try the simple solution first).

finestructure:
Actually, I think we should use `customDump`, it generates much more useful output:

CleanShot 2023-04-19 at 11 06 42@2x

finestructure:
What we could do instead of all this is of course to just round-trip it back to `yml`. That'd strip out any unrecognised attributes as well and present the result in a familiar format.

finestructure:
I just remembered why I didn't do that initially: In the happy path it'll just echo the input back to the user and it may be unclear why that is if it looks identical (and too subtle even if there are diffs unless we highlight them). There's valuable feedback in showing the data in an obviously parsed format.

daveverwer:
Yes, I like it with the dump format. I'm just going to make one small tweak and then I think it's ready to go.

daveverwer (approved):

PR #2348, opened 2023-04-20
Issue 2346 remove rollbar

finestructure:
Not for merging, this is for reference. I will redo the changes in this PR as incremental changes. This is just to record the overall progress (and to possibly deploy it to `dev` for testing) and to see if/how it would work.

Changes:

- remove Rollbar
- move from explicit `logger:` parameter across many, many function to `Current.logger()`
- introduce `CapturingLogger` in tests to be able to validate logging output
- adjust tests that checked for errors reported via Rollbar to captured logs

I had started by removing Rollbar and then realised if we remove that we'll need to recover the error report validation. That led to the CapturingLogger change.

I probably moved to Current.logger() prematurely, I think that part of that diff can be deferred. (We should do it eventually, not only because it simplifies interfaces but because we already have `Current.logger()` - we're just not using it universally.)

So what I'll try and do in separate PRs:

1. introduce `CapturingLogger` + move Rollbar error tests over
2. remove Rollbar
3. move to `Current.logger()` universally (perhaps far later)


finestructure:
Superseded by #2350

PR #2357, opened 2023-04-27
Encode parameter requirements in Query type

finestructure:
Pulling this tiny change out into its own PR because it's a critical piece that shouldn't get lumped in with other changes.


daveverwer (approved):

finestructure:
Ended up putting more into this PR:

- move search from ELF to a/a
- make `API.search` based on `Query` as the input parameter rather than `query: String, page: Int`

PR #2360, opened 2023-04-27
Include a canonical URL in all documentation pages

daveverwer:
This is WIP as it still needs the remainder of the path appended on the end of the canonical URL but I wanted to open the PR now as I feel like the back of it is broken now 😅

### DIFF
--- a/Sources/App/Core/DocumentationVersion.swift
+++ b/Sources/App/Core/DocumentationVersion.swift
@@ -17,17 +17,11 @@ import Foundation
 import Fluent
 import SemanticVersion
 
-
-struct DocumentationVersion: Equatable {
-    var reference: Reference
-    var ownerName: String
-    var packageName: String
-    var docArchives: [DocArchive]
-    var latest: Version.Kind?
-    var updatedAt: Date
+struct DocumentationVersionMetadata {
     var canonicalTarget: DocumentationTarget?
+    var versions: [DocumentationVersion]
 
-    static func query(on database: Database, owner: String, repository: String) async throws -> [Self] {
+    static func query(on database: Database, owner: String, repository: String) async throws -> Self {
         let results = try await Joined3<Version, Package, Repository>
             .query(on: database,
                    join: \Version.$package.$id == \Package.$id, method: .inner,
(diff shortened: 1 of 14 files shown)


Sources/App/Core/DocumentationVersion.swift
```diff
@@ -17,17 +17,12 @@ import Foundation
 import Fluent
 import SemanticVersion
 
+struct DocumentationVersionMetadata {
```
finestructure, line 20: Is this perhaps just `DocumentationMetadata` with `versions` being an aspect (property) of it?

daveverwer:
OK this is ready for review.

Note that this does _not_ cover the situation where the canonical URL becomes incorrect if someone visits the page and then navigates to a link using the Vue links. For example:

1. Person visits `http://swiftpackageindex.com/owner/repo/non-canonical-ref/documentation/archive` and the canonical `link` in the `head` gets rendered as `http://swiftpackageindex.com/owner/repo/canonical-ref/documentation/archive`. ✅
2. They navigate to `http://swiftpackageindex.com/owner/repo/non-canonical-ref/documentation/archive/symbol` by clicking on a link within the Vue documentation, and the canonical URL is still in the `head` as `http://swiftpackageindex.com/owner/repo/canonical-ref/documentation/archive` rather than `http://swiftpackageindex.com/owner/repo/canonical-ref/documentation/archive/symbol` ❌
3. If they then refresh the page from the server, the canonical link is fixed and renders as `http://swiftpackageindex.com/owner/repo/canonical-ref/documentation/archive/symbol` ✅

This either will or will not be a problem depending on how Google looks at canonical URLs. We know that it uses Chrome to render pages and waits for all JavaScript to execute before indexing, but there’s no definitive answer to whether it navigates pages in the same browsing session or whether it gathers links from rendered pages and requests those pages with separate requests. There is some information [here](https://developers.google.com/search/docs/crawling-indexing/javascript/javascript-seo-basics#properly-inject-canonical-links) but it does not cover the situation of browsing after rendering, and [here](https://webmasters.stackexchange.com/questions/138519/does-googlebot-call-every-link-separately-with-a-new-request) where the consensus is that every page is a separate request, but those answers are not definitive or confirmed by Google.

I think we’re probably fine here, and that every page would be a separate request, but it’s something to keep in mind as we test how this performs.

daveverwer:
Ready for review but don't deploy to production as I need to take measurements from search console before roll this out so I can measure if it works.

finestructure:
> Ready for review but don't deploy to production as I need to take measurements from search console before roll this out so I can measure if it works.

Best we don't merge it yet then or it'll block us from deploying. How are you measuring? Via the dev env? We can ad hoc deploy in that case, that'll keep our options open.

finestructure (approved):

daveverwer:
> Best we don't merge it yet then or it'll block us from deploying. How are you measuring? Via the dev env? We can ad hoc deploy in that case, that'll keep our options open.

Absolutely. I'm going to take the measurements now and then deploy through to production. It can only be tested in production as we hide the staging site from Google completely.

daveverwer:
Deploying as 2.83.15

daveverwer:
Tested and working in production.

PR #2365, opened 2023-04-28
Make AuthorMetadata model associated values non-optional

finestructure:


Sources/App/Views/PackageController/PackageShow+Model.swift
```diff
@@ -20,8 +20,8 @@ import SPIManifest
 
 
 enum AuthorMetadata : Equatable {
-    case fromSPIManifest(String?)
-    case fromGitRepository(PackageAuthors?)
+    case fromSPIManifest(String)
+    case fromGitRepository(PackageAuthors)
```
finestructure, line 24: `. fromSPIManifest(nil)` or `. fromGitRepository(nil)` don't really make sense. It's the `AuthorMetadata` as a whole that's already optional on `PackageShow.Model`, i.e. _that's_ where `nil` "happens".

Alternatively we could have

```swift
enum AuthorMetadata : Equatable {
    case fromSPIManifest(String)
    case fromGitRepository(PackageAuthors)
    case none
}
```

daveverwer: Yep, this looks good 👍

daveverwer (approved):

PR #2408, opened 2023-06-02
Switch from api to pageview event

finestructure:


finestructure:
Once this is deployed we should reset stats on test.packageindex.com to have an easier time validating the change.

daveverwer (approved):

PR #2412, opened 2023-06-05
Change repository creation date in snapshot tests

daveverwer:
This is more interesting than I originally thought. I'm honestly not sure why these changed today, or why they have different results when run on Linux today.

Changed to not span a year boundary for now, but this is odd.


daveverwer:
Not ready for merge

PR #2427, opened 2023-06-12
Add macro target type

finestructure:
This adds support for parsing and storing `macro` targets.

There is currently no `macro` product type in SwiftPM, which impacts our ability to apply our `product:` search filter.

This PR makes no attempt to address that, it simply add support for the new target type and saves it. Addressing the search filter is a separate issue.

This PR is complete, I'm just going to wait for Yim's response before setting it non-draft and merging.


Sources/App/Core/PackageCollection+generate.swift
```diff
@@ -295,3 +299,13 @@ private extension PackageCollection.Platform {
...
     }
 }
+
+
+extension PackageCollectionModel.V1.Signer {
+    static var spi: Self {
+        .init(type: "ADP",
+              commonName: "Swift Package Index",
+              organizationalUnitName: "Swift Package Index",
+              organizationName: "Swift Package Index")
```
finestructure, line 309: @yim-lee Hi Yim, I'm adapting our package collection handling to the changes in the release/5.9 branch of SwiftPM and I was wondering what the recommended values are for these fields, in particular `type`. I saw mention of `ADP` in the doc string but found no other list of options or anything. I'm assuming `ADP` is short for `Apple Developer Portal` and would be what we should be using for our dev portal issued cert?

daveverwer (approved):

Sources/App/Core/PackageCollection+generate.swift
yim-lee: Hi Sven, the `signer` field is optional and the only supported signer type is `ADP` (Apple Developer Program), so only fill this in if you are signing with an ADP cert (which you are). `commonName` etc. should match the corresponding subject fields in the certificate. Please refer to the [docs](https://github.com/apple/swift-package-manager/blob/main/Sources/PackageCollectionsModel/Formats/v1.md) and let me know if any of it is unclear.

PR #2428, opened 2023-06-12
Optional version array canonical target

daveverwer:


Tests/AppTests/ArrayVersionExtensionTests.swift
```diff
@@ -0,0 +1,178 @@
...
+                nil,
+                nil
+            ]
+
+            XCTAssertEqual(versions.canonicalDocumentationTarget(),
+                           .internal(reference: "main", archive: "foo"))
+        }
+
+        do {
+            // [ Default branch (with docs), Release (no docs), nil ] = Release
```
finestructure, line 47: Comment looks wrong, = default branch says the test (I'd drop these comments and let the tests "speak" tbh, that avoids drift)

finestructure: well actually, given how big the Version init is, the comment does help outline the gist of the array input, so it does help

PR #2430, opened 2023-06-13
Create PG extensions `IF NOT EXISTS`

RemiBardon:
Running the migrations on a PostgreSQL database already containing the needed extensions results in a failure. This commit fixes it.

I added `IF NOT EXISTS` in the `CREATE EXTENSION` queries, as explained by [the PostgreSQL docs](https://www.postgresql.org/docs/current/sql-createextension.html).

FYI I'm working on adding support for Swift on [Clever Cloud](https://clever-cloud.com). I tried to deploy this Vapor app to detect deployment issues, and found this while running migrations after a successful build. The extension you need are [enabled by default in our managed PostgreSQL add-ons](https://www.clever-cloud.com/doc/deploy/addon/postgresql/postgresql/#default-extensions), which is why I got a failure.


RemiBardon:
I rebased on `main` and fixed an issue. Now the PR should be clean.

finestructure (approved):

daveverwer (approved):
Thanks so much Remi! We received your CLA and you're all set.

Thank you for your contribution 👍

PR #2432, opened 2023-06-14
Search page uses all query parameters in the Canonical URL

tobyaherbert:
As discussed with @daveverwer, I noticed that the search page includes all query parameters in the `<link rel="canonical">` in the page's head element.

As I understand it, only the `query` query parameter should be included so that incoming links don't get to set a canonical URL with arbitrary parameters.

As an example, were someone to link to a search page with added UTM parameters, this would pollute the canonical URL. 

For example:
`/search?query=semantic&utm_campaign=test`


Tests/AppTests/SearchShowModelTests.swift
```diff
@@ -143,5 +144,13 @@ class SearchShowModelTests: XCTestCase {
 
         XCTAssertEqual(matchingKeywords, ["keyword2", "keyword4"])
     }
+    
+    func test_SearchShow_Model_canonicalURLAllowList() throws {
```
daveverwer, line 148: We needed to write an `app.test` test here as this situation only occurs with query parameters not included in `API.SearchController.Query `. Instead, these situations would come from incoming links or someone just typing into a browser address bar.

This seemed like a reasonable way to test it, would you agree @finestructure?

Tests/AppTests/__Snapshots__/SearchShowModelTests/test_SearchShow_Model_canonicalURLAllowList.1.html
```diff
@@ -0,0 +1 @@
...ta-focus-query-field="true"></ul><ul class="pagination"><li class="previous"><a href="/search?query=alamo&page=1">Previous Page</a></li></ul></section></section></div></main><footer><div class="inner"><nav><ul><li><a href="https://blog.swiftpackageindex.com">Blog</a></li><li><a href="https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server">GitHub</a></li><li><a href="/privacy">Privacy and Cookies</a></li><li><a href="https://swiftpackageindex.statuspage.io">Uptime and System Status</a></li><li><a href="/build-monitor">Build System Monitor</a></li><li><a href="https://twitter.com/swiftpackages">Twitter</a></li></ul><small>The Swift Package Index is entirely funded by community sponsorship. Thank you to <a href="https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server#funding-and-sponsorship">all our sponsors for their generosity</a>.</small><small>The Swift Package Index is operated by SPI Operations Limited, a company registered in the UK with company number 13466692.</small><a rel="me" href="https://mas.to/@SwiftPackageIndex"></a><a rel="me" href="https://mas.to/@SwiftPackageUpdates"></a></nav></div></footer><div class="staging">Staging / Development</div></body></html>
```
daveverwer, line 1: This snapshot is less than ideal given the compressed HTML we output, but again it seems reasonable to me.

daveverwer (approved):
Thank you so much for your contribution, @tobyaherbert! Looks good to me, but I just want to check the one comment below with @finestructure.

Tests/AppTests/SearchShowModelTests.swift
```diff
@@ -14,9 +14,10 @@
 
 @testable import App
 
+import SnapshotTesting
 import XCTVapor
 
-class SearchShowModelTests: XCTestCase {
+class SearchShowModelTests: AppTestCase {
```
finestructure, line 20: Better to use `SnapshotTestCase` here instead, it pins dates and makes it easy to switch on recorded if needed.

Note though that changing from `XCTestCase` to `AppTestCase` (or `SnapshotTestCase`, which inherits from `AppTestCase`) means all tests now run through a db tearDown and setUp phase, making them much slower (0.90s vs 0.01s).

It's not a huge deal with only six other tests in this file but it would be better to leave this as `XCTestCase` and pull out the test that needs the snapshot into its own class (even if it's in the same file). E.g.

```swift
class SearchShowModelSnapshotTests: SnapshotTestCase {
    func test_SearchShow_Model_canonicalURLAllowList() throws {
        try app.test(.GET, "search?query=alamo&page=2&utm_campaign=test&utm_source=email", afterResponse: { res in
            XCTAssertEqual(res.status, .ok)

            assertSnapshot(matching: res.body.string, as: .html)
        })
    }
}
```

That way if more tests are added that don't need the db they don't automatically pay the setup cost.

Tests/AppTests/SearchShowModelTests.swift
finestructure: Actually, there's a slightly better way to get the snapshot and even render it in a more readable way:

```swift
    @MainActor
    func test_SearchShow_Model_canonicalURLAllowList() async throws {
        let req = Vapor.Request(application: app,
                                url: "search?query=alamo&page=2&utm_campaign=test&utm_source=email",
                                on: app.eventLoopGroup.next())
        let html = try await SearchController.show(req: req)
        assertSnapshot(matching: html.render(indentedBy: .spaces(2)), as: .html)
    }
```

→

```
<!DOCTYPE html>
<html lang="en"><!--Version: test--><!--DB Id: db-id-->
  <head>
    <meta name="robots" content="noindex"/>
    <meta name="viewport" content="width=device-width, initial-scale=1.0"/>
    <meta charset="UTF-8"/>
    <meta property="og:site_name" content="The Swift Package Index"/>
    <link rel="canonical" href="http://localhost:8080/search?query=alamo&page=2"/>
    <meta name="twitter:url" content="http://localhost:8080/search?query=alamo&page=2"/>
    <meta property="og:url" content="http://localhost:8080/search?query=alamo&page=2"/>
```

finestructure: If we really wanted to make the test explicit (rather than in a roundabout way via a full snapshot), we could probably drill into the `HTML` node we're getting back and inspect the elements directly but it's probably not worth the effort.

Tests/AppTests/__Snapshots__/SearchShowModelTests/test_SearchShow_Model_canonicalURLAllowList.1.html
finestructure: Oh, I hadn't even seen this comment yet - this should be fixed by my suggestion above (and could even be avoid entirely by inspecting the `HTML` node object).

finestructure (approved):

Tests/AppTests/SearchShowModelTests.swift
finestructure: Sorry for the spam, I had another thought. We could also write this as follows, without any snapshotting:

```swift
    func test_SearchShow_Model_canonicalURLAllowList() async throws {
        let req = Vapor.Request(application: app,
                                url: "search?query=alamo&page=2&utm_campaign=test&utm_source=email",
                                on: app.eventLoopGroup.next())
        let html = try await SearchController.show(req: req).render(indentedBy: .spaces(2))
        XCTAssert(
            html.contains(#"<link rel="canonical" href="http://localhost:8080/search?query=alamo&page=2"/>"#),
            "was: \(html)"
        )
    }
```

This way it's very explicit what we're looking for. It has the downside that (unless we added another check to constrain the number of `link rel` snippets to 1) we're not asserting that there's a second one with the extra parameters.

tobyaherbert:
Thanks for the tips @finestructure!

I've extracted the test into it's own file to restore the performance of the original tests, and have refactored the test to use `SwiftSoup` instead so we aren't relying on the exact order of attributes on the `<link>` element.

This is very slightly slower than just checking if the HTML contains the `<link>` element, but it's negligible (about 0.02s from my testing). I did also run both variants in a loop of 100 attempts to confirm it has minimal impact.

finestructure:
> Thanks for the tips @finestructure!
> 
> I've extracted the test into it's own file to restore the performance of the original tests, and have refactored the test to use `SwiftSoup` instead so we aren't relying on the exact order of attributes on the `<link>` element.
> 
> This is very slightly slower than just checking if the HTML contains the `<link>` element, but it's negligible (about 0.02s from my testing). I did also run both variants in a loop of 100 attempts to confirm it has minimal impact.

Ah yes, none of this will matter - the db stuff will dwarf anything else we do.

Just to explain in more detail: the "problem" is that any db-related test requires `AppTestCase.setUp` to run and that runs a number of SQL queries to truncate all tables - which is very slow (for a test). So we should try and use `AppTestCase` as a base test class only when we actually use the db.

daveverwer:
> Just to explain in more detail: the "problem" is that any db-related test requires `AppTestCase.setUp` to run and that runs a number of SQL queries to truncate all tables - which is very slow (for a test). So we should try and use `AppTestCase` as a base test class only when we actually use the db.

I think this still needs to be an AppTestCase, not for the DB but because of this line:

```swift
let request = Vapor.Request(application: app,
```

PR #2433, opened 2023-06-14
Package pages should use the canonical capitalisation for URLs from the database

tobyaherbert:
Package page URLs are case-insensitive but the canonical URL shouldn't change when only the repository owner or repository name have a different case in the URL. Instead, we should use the capitalisation from the repository owner and repository name fields stored in the database.

### DIFF
--- a/Tests/AppTests/__Snapshots__/WebpageSnapshotTests/test_PackageShowView.1.html
+++ b/Tests/AppTests/__Snapshots__/WebpageSnapshotTests/test_PackageShowView.1.html
@@ -5,9 +5,9 @@
     <meta name="viewport" content="width=device-width, initial-scale=1.0"/>
     <meta charset="UTF-8"/>
     <meta property="og:site_name" content="The Swift Package Index"/>
-    <link rel="canonical" href="http://localhost:8080/"/>
-    <meta name="twitter:url" content="http://localhost:8080/"/>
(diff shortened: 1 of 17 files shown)


daveverwer (approved):
This is so well spotted, @tobyaherbert. Thanks so much for the contribution!

Tests/AppTests/__Snapshots__/WebpageSnapshotTests/test_PackageShowView.1.html
```diff
@@ -5,9 +5,9 @@
     <meta name="viewport" content="width=device-width, initial-scale=1.0"/>
     <meta charset="UTF-8"/>
     <meta property="og:site_name" content="The Swift Package Index"/>
-    <link rel="canonical" href="http://localhost:8080/"/>
-    <meta name="twitter:url" content="http://localhost:8080/"/>
-    <meta property="og:url" content="http://localhost:8080/"/>
+    <link rel="canonical" href="http://localhost:8080/Alamo/Alamofire"/>
+    <meta name="twitter:url" content="http://localhost:8080/Alamo/Alamofire"/>
+    <meta property="og:url" content="http://localhost:8080/Alamo/Alamofire"/>
```
finestructure, line 10: Oh wow, did we really have wrong canonical URLs on all our package pages this whole time?! 😅

finestructure (approved):

PR #2434, opened 2023-06-14
Use `.url` instead of manually adding a `.link` for the canonical URL

tobyaherbert:
On documentation pages, we should use `.url`  instead of `.link` and manually constructing a `<link rel="canonical" />` element. This matches the behaviour of the `PublicPage` model for other static pages served by the site.


daveverwer (approved):
Thanks so much, Toby!

PR #2436, opened 2023-06-14
Use the canonical owner and repository capitalisation for documentation canonical URLs

tobyaherbert:
Merge After #2435

Documentation pages should use the canonical owner and repository capitalisation from the database, as the canonical URLs must not change when the page is loaded with a different case owner or repository.

### DIFF
--- a/Sources/App/Controllers/PackageController+routes.swift
+++ b/Sources/App/Controllers/PackageController+routes.swift
@@ -228,10 +228,14 @@ enum PackageController {
         }
 
         let canonicalUrl: String? = {
-            guard let canonicalTarget = documentationMetadata.canonicalTarget else { return nil }
+            guard let canonicalOwner = documentationMetadata.owner,
+                  let canonicalRepository = documentationMetadata.repository,
+                  let canonicalTarget = documentationMetadata.canonicalTarget
+            else { return nil }
+            
             return Self.canonicalDocumentationUrl(from: "\(req.url)",
-                                                  owner: owner,
-                                                  repository: repository,
+                                                  owner: canonicalOwner,
+                                                  repository: canonicalRepository,
                                                   fromReference: reference,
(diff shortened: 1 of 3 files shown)


Sources/App/Controllers/PackageController+routes.swift
```diff
@@ -227,11 +227,23 @@ enum PackageController {
...
+                .field(\.$owner)
+                .field(\.$name)
+                .filter(Repository.self, \.$owner, .custom("ilike"), owner)
+                .filter(Repository.self, \.$name, .custom("ilike"), repository)
+                .first()
+            
+            guard let repositoryName = repository?.name,
+                  let repositoryOwner = repository?.owner,
+                  let canonicalTarget = documentationMetadata.canonicalTarget
+            else { return nil }
```
finestructure, line 242: Could we not simply make a `lowercased()` owner/repo pair the canonical URL? I'd hate to have to run an extra query on all these requests just for that lookup.

finestructure: Alternatively, we should pull the fields from the upstream call to `DocumentationMetadata.query` and pass them through instead of the query parameter values.

finestructure: (Sorry, just realised this was still a draft)

tobyaherbert:
How does this look @finestructure?

We've pulled the info from the existing `DocumentationMetadata.query` instead of executing another DB query.
There's still a test to write at some point.

finestructure:
> How does this look @finestructure?
> 
> We've pulled the info from the existing `DocumentationMetadata.query` instead of executing another DB query. There's still a test to write at some point.

That's great, that's exactly what I had in mind!

daveverwer (approved):
Thank you @tobyaherbert!

PR #2437, opened 2023-06-15
Fix build on Xcode 15

finestructure:


Package.resolved
```diff
@@ -344,10 +344,9 @@
     {
       "identity" : "swift-snapshot-testing",
       "kind" : "remoteSourceControl",
-      "location" : "https://github.com/pointfreeco/swift-snapshot-testing.git",
+      "location" : "https://github.com/finestructure/swift-snapshot-testing.git",
```
finestructure, line 347: I've not pushed this into the `SwiftPackageIndex` org because it should be short-lived. There's a PR open to upstream it: https://github.com/pointfreeco/swift-snapshot-testing/pull/737

daveverwer (approved):

PR #2416, opened 2023-06-07
Convert the site map to be a site map index

daveverwer:
This is early days but I wanted to get this documented:

- [x] `sitemap.xml` should be a site map index instead of a plain site map
- [x] Create a new static site map containing all of the top level pages like the home page, FAQ, etc…
- [x] Create a site map for every package on `/owner/repo/sitemap.xml`
- [x] Reference all package site maps in the index
- [x] Include all documentation page URLs from the data generated by `--emit-digest`

The plan is to get everything apart from the last task in the above checklist into this PR and then create a new PR for the `–emit-digest` changes that merges into this one. This is already a big PR, and that one could well be big, too.

### Go Live Plan

We should think about how we want this to go live as we want Google to see an entirely new set of URLs all at once.

- [x] Re-generate documentation for the canonical docs version so that every package that has documentation is covered by these changes
- [ ] Add some of the new sitemap files to the smoke test
- [ ] Merge PR and deploy

PR #2438, opened 2023-06-15
Track rss and sitemap

finestructure:


Sources/App/Core/AppEnvironment.swift
```diff
@@ -179,7 +189,7 @@ extension AppEnvironment {
         },
         mastodonPost: Mastodon.post(client:message:),
         metricsPushGatewayUrl: { Environment.get("METRICS_PUSHGATEWAY_URL") },
-        plausibleAPIReportingSiteID: { Environment.get("PLAUSIBLE_API_REPORTING_SITE_ID") },
+        plausibleBackendReportingSiteID: { Environment.get("PLAUSIBLE_BACKEND_REPORTING_SITE_ID") },
```
finestructure, line 192: I've added the corresponding new variables to `dev` and `prod` and will remove the old one after deployment.

Sources/App/Core/Plausible.swift
```diff
@@ -31,16 +31,9 @@ enum Plausible {
...
-    }
-
-    enum Identifier {
-        case anonymous
-        case identifier(String)
-
-        var props: [String: String] {
-            fatalError("fix me")
-        }
+        case sitemap = "/sitemap"
```
finestructure, line 36: These (`/rss` and `/sitemap`) aren't real paths, they're just for filtering in Plausible. I've just picked something that seemed sensible.

Sources/App/routes.swift
```diff
@@ -259,16 +259,20 @@ func routes(_ app: Application) throws {
...
+            
+            $0.get(SiteURL.rssReleases.pathComponents, use: RSSFeed.showReleases)
+                .excludeFromOpenAPI()
+        }
 
-        app.get(SiteURL.siteMap.pathComponents) { req in
-            SiteMap.fetchPackages(req.db)
-                .map(SiteURL.siteMap)
-        }.excludeFromOpenAPI()
+        app.group(BackendReportingMiddleware(path: .sitemap)) {
```
daveverwer, line 270: When I rebase #2416 on top of this I'll create separate tracking for the three different site map types as we want to know how often it hits the index vs the package-specific site maps.

daveverwer (approved):

Sources/App/Core/Plausible.swift
daveverwer: Yes this is fine. No need for the `.xml`.

Sources/App/routes.swift
```diff
@@ -259,16 +259,20 @@ func routes(_ app: Application) throws {
...
-
-        app.get(SiteURL.rssReleases.pathComponents, use: RSSFeed.showReleases)
-            .excludeFromOpenAPI()
+        app.group(BackendReportingMiddleware(path: .rss)) {
+            $0.get(SiteURL.rssPackages.pathComponents, use: RSSFeed.showPackages)
+                .excludeFromOpenAPI()
+            
+            $0.get(SiteURL.rssReleases.pathComponents, use: RSSFeed.showReleases)
+                .excludeFromOpenAPI()
+        }
```
finestructure, line 268: I've collapsed both rss feed paths (the actual ones are `/packages.rss` and `releases.rss` into `/rss` for sake of simplicity. We could also report them individually under some path, perhaps with a `/rss/` prefix to make filtering easier 🤷‍♂️

daveverwer: I think this is fine as one trigger. No benefit in splitting these out. 👍

PR #2447, opened 2023-06-18
Swapped CSS/JS test suite function names

tobyaherbert:
I just noticed that the CSS/JS tests for documentation pages appear to have the inverse function names.

- `test_documentation_css` is testing JS related URLs.
- `test_documentation_js` is testing CSS related URLs.

I've swapped these round so they test the expected features.


daveverwer (approved):
Well spotted, @tobyaherbert! Thank you

PR #2452, opened 2023-06-19
Add package specific site maps with documentation URLs where available

daveverwer:
Not ready for merge, but ready for discussion. @finestructure let's chat about it tomorrow, especially the WIP tests. I've also left some comments inline on some other aspects.


Sources/App/Controllers/PackageController+routes.swift
```diff
@@ -284,6 +287,75 @@ enum PackageController {
         return response
     }
 
+    static func siteMap(req: Request) async throws -> Response {
```
daveverwer, line 290: This method is too long. May need a refactor before merging, but let's discuss.

Sources/App/routes.swift
```diff
@@ -101,6 +101,9 @@ func routes(_ app: Application) throws {
             app.get(":owner", ":repository", ":reference", "js", "**") {
                 try await PackageController.documentation(req: $0, fragment: .js)
             }.excludeFromOpenAPI()
+            app.get(":owner", ":repository", ":reference", .fragment(.linkableEntities)) {
```
daveverwer, line 104: This is not strictly necessary, as it only allows SPI to serve the `linkable-entities.json` which is not necessary for us to fetch it for parsing. However, I added it because I bet we'll want to check what's in it occasionally and this is easier than figuring out the S3 URL.

We might want to add it here and remove it later, or maybe just remove it now.

Tests/AppTests/PackageController+routesTests.swift
```diff
@@ -697,18 +705,35 @@ class PackageController_routesTests: AppTestCase {
...
         // MUT
         try app.test(.GET, "/owner/package/1.2.3/theme-settings.json") {
             XCTAssertEqual($0.status, .ok)
-            XCTAssertEqual($0.content.contentType?.description, "application/octet-stream")
+            XCTAssertEqual($0.content.contentType?.description, "application/json")
             XCTAssertEqual($0.body.asString(), "/owner/package/1.2.3/theme-settings.json")
         }
     }
 
+    func test_linkableEntites() throws {
```
daveverwer, line 720: This test should also go if we don't end up serving the `linkable-entities.json`.

Sources/App/Controllers/PackageController+routes.swift
```diff
@@ -66,15 +66,18 @@ enum PackageController {
...
         case js
+        case linkableEntities = "linkable-entities.json"
         case themeSettings = "theme-settings.json"
         case tutorials
 
         var contentType: String {
             switch self {
                 case .css:
                     return "text/css"
-                case  .data, .faviconIco, .faviconSvg, .images, .img, .index, .themeSettings:
```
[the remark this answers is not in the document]
finestructure: I tossed them all into `application/octet-stream` because these files aren't really "served" in the sense that anything other than the DocC JS ingests them and that doesn't seem to care about content-types.

I guess it's fine to adjust it but the bigger question is why we should proxy linkable entities (which I suppose triggered this change). Is the DocC site using this in some way?

Sources/App/Controllers/PackageController+routes.swift
finestructure: TBH I think it's fine but it looks like there's a nested `func linkableEntityUrls ` that could easily be pulled out into a sibling `static func`?

Sources/App/routes.swift
finestructure: My gut feeling is not to serve it unless it serves an actual purpose for the DocC app. This way there's no doubt which parts it needs and which parts it doesn't, and we're not tempted to run any lower level things through the proxy when we should really be going to the source (say "fetch all `linkable-entities.json` files").

daveverwer: So I just started removing this and it's a little tricky. We use the same fragment to build the AWS URLs for fetching resources directly from S3. If we don't have a `linkableEntities` in the enum, we can't easily build the URL we need to fetch the file.

What I'd suggest is, given that this is already a large PR, we don't tackle solving this here but in a new issue afterwards.

Of course, I could be missing something and there may be an easy way to build the URL without a fragment.

daveverwer:
OK, I think this is ready for a final review and to get merged into #2416

Sources/App/routes.swift
finestructure: I think in that case we leave it, as it's become a required part of what's exposed from the S3 doc archive content. It's not the DocC app per se, but what _we_ use in the DocC-Proxy to host the DocC app.

What I was mainly getting at is that we should be mindful of which parts of the doc archive content we treat as opaque and which parts we explicitly reference.

finestructure (approved):

PR #2453, opened 2023-06-20
Review suggestions

finestructure:


Tests/AppTests/SitemapTests.swift
```diff
@@ -52,54 +52,109 @@ class SitemapTests: SnapshotTestCase {
...
+                            """)
+            )
+        }
+
+        // MUT
+        let urls = await PackageController.linkableEntityUrls(client: app.client, packageResult: packageResult)
+
+        XCTAssertEqual(urls, [
+            "https://spi.com/Owner/Repo0/default/documentation/foo/bar/1",
+            "https://spi.com/Owner/Repo0/default/documentation/foo/bar/2"
```
finestructure, line 108: I believe this highlights a potential issue - should these be relative URLs?

finestructure: Never mind, it looks like we have absolute URLs in sitemap.xml.

daveverwer: Sitemap URLs can be either, but in reading more about this it seems absolute ones are preferred. No ambiguity at all.

PR #2456, opened 2023-06-21
Add and persist target type

finestructure:
This add `Target.type` and saves it. We're not doing anything with it yet but I wanted to get this in early so we start picking up `macro` targets without having to re-analyze too many versions.

⚠️ schema change ⚠️

### DIFF
--- a/Sources/App/Migrations/064/UpdateTargetAddType.swift
+++ b/Sources/App/Migrations/064/UpdateTargetAddType.swift
@@ -1,0 +1,29 @@
+// Copyright Dave Verwer, Sven A. Schmidt, and other contributors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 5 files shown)


Sources/App/Migrations/064/UpdateTargetAddType.swift
```diff
@@ -0,0 +1,29 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import Fluent
+
+struct UpdateTargetAddType: AsyncMigration {
+    func prepare(on database: Database) async throws {
+        try await database.schema("targets")
+            .field("type", .json)
```
finestructure, line 20: We'll want to make this `.required` at some point but we'll have to have re-analysed all versions and populated the field for that.

finestructure:
Is it ok to merge this now?

daveverwer:
Yep! Thanks for holding it!

PR #2457, opened 2023-06-21
Metanoindex

### LINKED ISSUE #2448: Changes to how we `noindex` pages
We should think about `noindex`ing the "Information for site maintainers" page and the "build index" page, too. There is a lot of duplicated text on the maintainers page, and the build index provides nothing useful to Google, so let's hide them.

We should also move how we `noindex` the specific build pages out of the `robots.txt` and into meta tags.

daveverwer:
Fixes #2448

- [x] Allow `PublicPage` subclasses to opt-out of being indexed in production environments (staging will always `noindex`)
- [x] Opt out the Information for package maintainers and the two build pages (index and show)
- [x] Replace the `noindex`ing of the build detail pages from the `robots.txt`

### DIFF
--- a/Public/robots.txt
+++ b/Public/robots.txt
@@ -1,6 +1,1 @@
-# All robots allowed
-User-agent: *
-Disallow: /builds/*
-
-# Sitemap files
 Sitemap: https://swiftpackageindex.com/sitemap.xml
--- a/Sources/App/Views/PackageController/Builds/BuildIndex+View.swift
+++ b/Sources/App/Views/PackageController/Builds/BuildIndex+View.swift
@@ -26,6 +26,11 @@ enum BuildIndex {
             super.init(path: path)
         }
 
+        override func allowIndexing() -> Bool {
+            // Block this page from being indexed by search engines.
+            return false
+        }
+
         override func pageTitle() -> String? {
             "\(model.packageName) &ndash; Build Results"
         }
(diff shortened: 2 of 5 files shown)


finestructure (approved):
Surely there are snapshot diffs that need updating?

finestructure:
Ah, I see. That's actually fixable if we want to:

```swift
struct AppEnvironment {
    ...
    var environment: () → Environment
    ...
}

// in live
    ...
    environment: { (try? Environment.detect()) ?? .development }
    ...

// in PublicPage
    if Current.environment() == .production && allowIndexing() {
   ...

// in snapshot tests
    Current.environment = { .production }
```

daveverwer:
> Ah, I see. That's actually fixable if we want to:
> 
> ```swift
> struct AppEnvironment {
>     ...
>     var environment: () → Environment
>     ...
> }
> 
> // in live
>     ...
>     environment: { (try? Environment.detect()) ?? .development }
>     ...
> 
> // in PublicPage
>     if Current.environment() == .production && allowIndexing() {
>    ...
> 
> // in snapshot tests
>     Current.environment = { .production }
> ```
> 
> Haven't tried it but it should work.

Great idea, we should probably run webpage snapshots in production for this reason. I'll open an issue.

daveverwer:
Tested on production and working fine 👍

PR #2466, opened 2023-06-23
Rename get route.model.platform results

finestructure:
Last one? 🤞


daveverwer (approved):
Is it too late to say that I think they should be `ios`, `macos`, `tvos`, etc? 😂

I'll be running away, now...

finestructure:
💨⛵

PR #2467, opened 2023-06-23
Add vision os

finestructure:
To do before merging:

- [x] install Xcode 15b2 on mb6
- [x] validate builder [Merge request](https://gitlab.com/finestructure/swiftpackageindex-builder/-/merge_requests/211)
- [x] put visionOS after macOS
- [x] plan deployment sequence and rollback actions
- [x] Tag new release of SPIManifest
- [x] Adopt new release of SPIManifest in builder
- [x] Adopt new release of SPIManifest in server
- [ ] deploy builder
- [ ] deploy server

CleanShot 2023-06-28 at 10 05 01@2x


finestructure:
I wonder if we should swap `visionOS` and `tvOS` - it'll look a bit more balanced when it's not next to the equally long `watchOS`.

finestructure:
CleanShot 2023-06-23 at 12 14 26@2x

daveverwer:
> I wonder if we should swap `visionOS` and `tvOS` - it'll look a bit more balanced when it's not next to the equally long `watchOS`.

Maybe even iOS, macOS, visionOS, watchOS, tvOS, Linux

The spacing is quite tight on those. I might play with one point less on the font size or maybe a 600 weight or something to to relieve it a bit.

finestructure:
We've already discussed tweaking the font sizes etc in a follow-up but I wanted to explicitly call out what it looks like on the mobile screens:

CleanShot 2023-06-28 at 15 05 33@2x

finestructure:
Only just noticed there wasn't a review tick on this yet. I'm going to "sudo merge" this in order to complete the rollout since we've discussed all this.

finestructure:
Live and processing:

CleanShot 2023-06-29 at 09 13 10@2x

PR #2473, opened 2023-06-30
Persist linkable paths count

finestructure:
This saves the new doc report property `linkablePathsCount` to the database. The change is backwards compatible (but the new builder is already live, too).

⚠️ schema change ⚠️


daveverwer (approved):
This is great. Thank you. Once visionOS processing is complete, we should run a re-generation of documentation to get these files in order on prod.

Thanks!

PR #2481, opened 2023-07-04
Macro search

### LINKED ISSUE #2425: Add support for `product:macro` searches
Aren't turning up yet although we definitely have some in the index, for example [FullQueueDeveloper/StaticMemberIterable](https://swiftpackageindex.com/FullQueueDeveloper/StaticMemberIterable):

CleanShot 2023-06-12 at 09 11 36@2x

finestructure:
Merge after #2480 

This add support for `product:macro` searches via "virtual" macro product types: we insert macro targets as "virtual" macro products into our search view (and there only, we don't change any of the underlying source data).

Fixes #2425

### DIFF
--- a/Tests/AppTests/QueryPerformanceTests.swift
+++ b/Tests/AppTests/QueryPerformanceTests.swift
@@ -115,7 +115,6 @@ class QueryPerformanceTests: XCTestCase {
     }
 
     func test_12_Search_refresh() async throws {
-        throw XCTSkip("run only after schema change with new index has been deployed")
         // We can't "explain analyze" the refresh itself so we need to measure the underlying
         // query.
         // Unfortunately, this means it'll need to be kept in sync when updating the search
@@ -155,7 +154,7 @@ class QueryPerformanceTests: XCTestCase {
               JOIN versions v ON v.package_id = p.id
             WHERE v.reference ->> 'branch' = r.default_branch
             """)
-        try await assertQueryPerformance(query, expectedCost: 31_300, variation: 500)
+        try await assertQueryPerformance(query, expectedCost: 55_000, variation: 500)
     }
 
 }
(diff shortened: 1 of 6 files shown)


Tests/AppTests/QueryPerformanceTests.swift
```diff
@@ -115,6 +115,7 @@ class QueryPerformanceTests: XCTestCase {
     }
 
     func test_12_Search_refresh() async throws {
+        throw XCTSkip("run only after schema change with new index has been deployed")
```
finestructure, line 118: I'll update the search figure once I've deployed this change from the branch. It needs to have the index in place. Without it, the query takes 30s locally on my machine vs 200ms with it in place.

finestructure: Actually, I don't need to deploy, I can just temporarily create the index manually for this test.

Tests/AppTests/QueryPerformanceTests.swift
```diff
@@ -139,15 +139,22 @@ class QueryPerformanceTests: XCTestCase {
...
+              ) AS product_types,
               ARRAY(SELECT DISTINCT name FROM products WHERE products.version_id = v.id) AS product_names,
               TO_TSVECTOR(CONCAT_WS(' ', COALESCE(v.package_name, ''), r.name, COALESCE(r.summary, ''), ARRAY_TO_STRING(r.keywords, ' '))) AS tsvector
             FROM packages p
               JOIN repositories r ON r.package_id = p.id
               JOIN versions v ON v.package_id = p.id
             WHERE v.reference ->> 'branch' = r.default_branch
             """)
-        try await assertQueryPerformance(query, expectedCost: 31_300, variation: 500)
+        try await assertQueryPerformance(query, expectedCost: 55_000, variation: 500)
```
finestructure, line 157: That's a big increase in search view creation cost. The actual query time on `dev` is ~1s now. We run this every every 25-30s, so it's taking significant overall time.

I think it's still ok but we should reconsider doing this once we've gathered some experience with the feature or in case we make further extensions to search.

finestructure: Search view updates are linked to analysis passes. We could also reduce relative cost by not updating search on every analysis pass but on a "once every 5min" global timer.

PR #2482, opened 2023-07-04
Adopt linkable paths

### LINKED ISSUE #2471: Improving the performance of documented package sitemaps
A package’s `linkable-entities.json` can be large and needs parsing every time a sitemap for that package is requested. One [specific package](https://swiftpackageindex.com/GetStream/stream-chat-swiftui/4.33.0/documentation/streamchatswiftui) has a sitemap with over 48,000 URLs, and the sitemap takes almost a minute to generate from a `linkable-entities.json` of over 80Mb.

As discussed, we should pre-process and only upload the list of URLs rather than the full `linkable-entities.json` file.

### In the builder

After documentation generation, we should parse `linkable-entities.json` and extract every `path` from the top-level dictionary. That is all the data we need to build the sitemap. These URLs need no processing and can be stored as a JSON string array or a text file with one URL per line.

`linkable-entities.json` and other `--emit-digest` files can all be deleted before the documentation is uploaded to S3.

### In the server

Rather than fetching and parsing `linkable-entities.json`, we fetch the newly generated file and generate the sitemap from that.

### Other notes

In the final rendered sitemap, we should clip the maximum number of URLs to 49,000 (we need a couple of URLs for the package itself, and the maximum is 50,000. We can clip this either in the builder or as we serve the sitemap, but we need to be able to report on any packages that are going over this limit.

It might be useful to store a count of the total number of documentation pages in the database against the associated `version` record so we can query packages that have large amounts of documentation. This may even be useful as an indicator for package score.

If we don’t want to do that, some kind of reporting of this should be built into this mechanism so we can investigate very large documentation sets.

daveverwer:
Fixes #2471

Hard to test the performance on dev as I don't think we have any of the huge packages with the files there, but we can test it on prod once it's deployed and the documentation re-generation has completed.

### DIFF
--- a/Sources/App/Controllers/PackageController+routes.swift
+++ b/Sources/App/Controllers/PackageController+routes.swift
@@ -294,7 +294,7 @@ enum PackageController {
         else { throw Abort(.notFound) }
 
         let packageResult = try await PackageResult.query(on: req.db, owner: owner, repository: repository)
-        let urls = await linkableEntityUrls(client: req.client, packageResult: packageResult)
+        let urls = await linkablePathUrls(client: req.client, packageResult: packageResult)
 
         return try await siteMap(packageResult: packageResult, linkableEntityUrls: urls)
             .encodeResponse(for: req)
@@ -325,7 +325,7 @@ enum PackageController {
         )
     }
 
-    static func linkableEntityUrls(client: Client, packageResult: PackageResult) async -> [String] {
+    static func linkablePathUrls(client: Client, packageResult: PackageResult) async -> [String] {
         guard let canonicalTarget = [packageResult.defaultBranchVersion.model,
                                      packageResult.preReleaseVersion?.model,
                                      packageResult.releaseVersion?.model].canonicalDocumentationTarget(),
--- a/Sources/App/Controllers/PackageController+routes.swift
+++ b/Sources/App/Controllers/PackageController+routes.swift
@@ -343,14 +343,10 @@ enum PackageController {
                                                     reference: reference, fragment: .linkablePaths, path: "")
             guard let body = awsResponse.body else { return [] }
 
-            struct LinkableEntity: Decodable {
-                var path: String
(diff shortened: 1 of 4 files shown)


daveverwer:
> If you trigger a build for the doc build version via the `Run trigger-builds -v ... -p ... -s ...` it won't be affected by the downscaling.

I'm deploying this branch to dev to test that and I'll kick off a SwiftSyntax build, too

finestructure (approved):

daveverwer:
> Slightly more worrying, and why I am not going to merge this yet is that the files from prod/staging are different sizes. I am struggling to find anything that can diff them, but I'll write a little script tomorrow to extract just the urls and diff those.

OK I wrote the script tonight as it was bugging me why these would be different, and the good news is that they are not. There were just `staging.` prefixes to all the URLs in one of the files.

How do I even function as a human being? 🙄

daveverwer:
I'll merge and deploy this tomorrow.

daveverwer:
Deploying as 2.88.1

daveverwer:
Only a negligible increase in performance.

Before:

```
$ curl https://swiftpackageindex.com/GetStream/stream-chat-swiftui/sitemap.xml > stream-old.xml
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 9387k  100 9387k    0     0   182k      0  0:00:51  0:00:51 --:--:-- 2363k
```

```
$ curl https://swiftpackageindex.com/GetStream/stream-chat-swiftui/sitemap.xml > stream-new.xml
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 9387k  100 9387k    0     0   188k      0  0:00:49  0:00:49 --:--:-- 2163k
```

PR #2485, opened 2023-07-05
Fix sitemap performance

finestructure:


daveverwer:
Date formatters strike again!

PR #2492, opened 2023-07-08
Full yarn upgrade, including moving to new major versions

daveverwer:


finestructure (approved):

daveverwer:
There's an issue with the `esbuild` major version upgrade that I'll need to fix later. This is no longer ready for merge!

PR #2494, opened 2023-07-12
Store readmes

### LINKED ISSUE #1095: Adopt the GitHub API for rendering README files
I wasn't aware that the [GitHub README API could respond to different content types](https://docs.github.com/en/rest/reference/repos#get-a-repository-readme) and using that API is a better way to fetch README content than we're currently using in #1090.

However, it's not as simple as switching. We're currently requesting page content on a per-request basis, but that doesn't work as the API call is rate limited. There are several options on how we could adopt this API, but the one that we have agreed is best is to cache README content as part of our ingestion process.

Steps to adopt this API:

- [ ] Fetch the README API using an `Accept` header of `application/vnd.github.v3.html+json` to get a rendered HTML version of the README file.
- [ ] Cache that HTML content within our Infrastructure, either on Azure BLOB storage, or in another local cache, but not in the main database. If using BLOB storage, implement after #1094.
- [ ] Change `PackageController.readme(req:)` to fetch from the cache rather than from the GitHub page directly.

finestructure:
This is almost done, I've tested it by locally ingesting SemanticVersion

https://s3.console.aws.amazon.com/s3/buckets/spi-dev-readmes?region=us-east-2&prefix=SwiftPackageIndex/SemanticVersion/&showversions=false

Left to do:

- [x] add test(s) for `S3Store.readme`
- [x] add test(s) for S3 storage dependent on etag
- [x] update (or remove) `S3StoreTests.test_copy`

⚠️ schema change ⚠️

Fixes #1095


Sources/App/Controllers/PackageController+routes.swift
```diff
@@ -355,25 +355,21 @@ enum PackageController {
...
-            .map(PackageReadme.Model.init(url:readme:))
-            .map(PackageReadme.View.init(model:))
-            .map { $0.document() }
+        let pkg = try await Joined<Package, Repository>
+            .query(on: req.db, owner: owner, repository: repository).get()
+        guard let readmeHtmlUrl = pkg.repository?.readmeHtmlUrl else {
+            return PackageReadme.View(model: .init(url: nil, readme: nil)).document()
+        }
+        let readme = try await Current.fetchS3Readme(req.client, owner, repository)
+        return PackageReadme.View(model: .init(url: readmeHtmlUrl, readme: readme)).document()
```
finestructure, line 372: NB: I'm already feeing the new readme content into `PackageReadme.View` here, so any remaining changes would have to happen within that class.

daveverwer: Thank you, that's perfect.

I'll make a PR to merge into this so it merges atomically.

finestructure:
Ok, I think I'm done with this but I'm going to want to take another pass with fresh eyes tomorrow.

finestructure:
Working locally:

CleanShot 2023-07-13 at 08 13 23@2x

finestructure:
I'll ad-hoc deploy this to give it a spin on dev.

finestructure:
I stopped processing on dev and reverted the change for the moment. We'll want to ensure all the readme paths we store are lowercased:

CleanShot 2023-07-13 at 08 49 01@2x

finestructure:
Couple of additional things to take into account:

- [ ] improve how we record S3 cache state
- [ ] test S3 overwrite behaviour

1. We need to make sure that if a readme is removed from S3 for some reason a failure to fetch it in `PackageController.readme` resets the readme field on repositories so it's being re-fetched.
2. We need to check that if a readme is cached again for some reason it doesn't lead to constant errors because an overwrite fails.

daveverwer (approved):

finestructure:
Re-deploying from branch to ingest on dev.

finestructure:
Ok, we've ingested >10% on dev, it's looking good and I'm going to merge and deploy to prod now.

PR #2495, opened 2023-07-12
Add front-end changes for processing the new readme files from S3

daveverwer:
For discussion in the morning.

### DIFF
--- a/Sources/App/Core/Extensions/S3Store+ext.swift
+++ b/Sources/App/Core/Extensions/S3Store+ext.swift
@@ -45,7 +45,7 @@ extension S3Store {
 
             let store = S3Store(credentials: .init(keyId: accessKeyId, secret: secretAccessKey))
             let key = S3Store.Key(bucket: readmeBucket, path: "\(owner)/\(repository)/readme.html")
-            Current.logger().debug("Copying \(tempfile) to \(key.url) ...")
+            Current.logger().debug("Copying \(tempfile) to \(key.s3Uri) ...")
             try await store.copy(from: tempfile, to: key, logger: Current.logger())
         }
     }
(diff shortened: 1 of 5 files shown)


Sources/App/Core/Extensions/S3Store+ext.swift
```diff
@@ -22,8 +22,8 @@ extension S3Store {
         guard let bucket = Current.awsReadmeBucket() else {
             throw Error.genericError("AWS_README_BUCKET not set")
         }
-        let key = Key(bucket: bucket, path: "\(owner)/\(repository)/readme.html")
-        guard let body = try await client.get(URI(string: key.url)).body else {
+        let url = "https://\(bucket).s3-us-east-2.amazonaws.com/\(owner)/\(repository)/readme.html"
+        guard let body = try await client.get(URI(string: url)).body else {
```
finestructure, line 26: We should push that url generation into `S3Store.Key`, using `S3Store.region`. I'll fix that up tomorrow and then merge it.

Sources/S3Store/S3Store.swift
```diff
@@ -67,10 +67,11 @@ extension S3Store {
 
         public init(bucket: String, path: String) {
             self.bucket = bucket
-            self.path = path
+            self.path = path.droppingLeadingSlashes
         }
 
-        public var url: String { "s3://\(bucket)/\(path)" }
+        public var objectUrl: String { "https://\(bucket).s3.\(S3Store.region).amazonaws.com/\(path)" }
+        public var s3Uri: String { "s3://\(bucket)/\(path)" }
```
finestructure, line 74: These now match the terms in the S3 info panel:

CleanShot 2023-07-13 at 08 00 57@2x

Note that you had `s3-us-east-2.amazonaws.com`, which I've adjusted to `s3.us-east-2.amazonaws.com` (note the `.` after `s3`). Both work, this just matches what's in the info panel.

finestructure (approved):

Sources/S3Store/S3Store.swift
daveverwer: That's great, thanks.

This is one thing I was going to discuss with you this morning. I also had to apply a policy to the bucket, but let's chat about it.

PR #2496, opened 2023-07-13
Add s3readme type

finestructure:
This changes `repositories.readme_etag` into a proper `repositories.s3_readme` object to better allow us to track caching status. We'll want to be able to see why it failed, not just have `NULL` in the field.

This also ensure we reset the field to an error state if we can't fetch it in the `PackageController`. Without that, a failing read would potentially never resolve if the readme doesn't change to a new etag.


daveverwer (approved):
👍

PR #2511, opened 2023-07-19
Fix relative URLs in README files

### LINKED ISSUE #2502: Relative links are not working in README files
This worked previously but must have been broken by #2497

See Leftpad for an example:

Screenshot 2023-07-14 at 13 28 24@2x

Both the relative image and the link below are broken. The relative correction that we have implemented needs an update.

daveverwer:
Fixes #2502

This turned out to be a real bug rather than a phantom. The original rendered README files we got from scraping GitHub included all parts of the URL from a `github.com` base URL. The README file API gives back URLs without owner/repo/branch parts.

We also get one additional bit of functionality here, which would have been difficult to do previously. We now get to fix up truly relative URLs as well as root relative ones.

### DIFF
--- a/Tests/AppTests/__Snapshots__/PackageReadmeModelTests/test_processReadme_processRelativeImages.1.txt
+++ b/Tests/AppTests/__Snapshots__/PackageReadmeModelTests/test_processReadme_processRelativeImages.1.txt
@@ -1,7 +1,7 @@
 <p>README content.</p> 
 <img src="https://example.com/absolute/image/url.png"> 
-<img src="https://github.com/root/relative/image/url.png"> 
-<img src="relative/image/url.png"> 
-<img src="https://github.com/url/with/encoded%20spaces.png"> 
-<img src="https://github.com/url/with/unencoded%20spaces.png"> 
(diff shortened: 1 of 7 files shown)


Tests/AppTests/__Snapshots__/PackageReadmeModelTests/test_processReadme_processRelativeImages.1.txt
```diff
@@ -1,7 +1,7 @@
 <p>README content.</p> 
 <img src="https://example.com/absolute/image/url.png"> 
-<img src="https://github.com/root/relative/image/url.png"> 
-<img src="relative/image/url.png"> 
-<img src="https://github.com/url/with/encoded%20spaces.png"> 
-<img src="https://github.com/url/with/unencoded%20spaces.png"> 
+<img src="https://github.com/owner/repo/raw/main/root/relative/image/url.png"> 
+<img src="https://github.com/owner/repo/raw/main/relative/image/url.png">
```
daveverwer, line 4: This is the new functionality we get. 👍

finestructure (approved):

daveverwer:
Deploying as 2.89.3

PR #2513, opened 2023-07-21
Refactor readme model

finestructure:
When I added S3 readme fetches I noticed that the `PackageReadme.Model` state is getting a bit complicated with its state representation. We've had

```swift
    struct Model: Equatable {
        var url: String?
        var repositoryOwner: String?
        var repositoryName: String?
        var defaultBranch: String?
		...
    }
```

but it's not clear when and why these properties would be `nil`, and whether that'd only ever happen collectively. I.e. it's easy to represent impossible state like `Model(url: "http://...", repositoryOwner: nil, repositoryName: "bar", defaultBranch: nil)`

What we really have is three cases:

- no readme (all values are `nil`)
- cache miss (url is non-nil, the other fields are irrelevant)
- cache hit (all fields are non-nil, owner, repo, branch get converted into an optional readme Element (optional, because the conversion can fail))

The purpose of this PR is to actually have the Model reflect that like

```swift
    enum Model {
        case noReadme
        case readme(url: String, readmeElement: Element?)
        case cacheLookupFailed(url: String)
    }
```

Once I'd started pulling that thread I reshuffled a couple of other things, mainly because the change to an enum forced feeding the `RepoTriple` (`(owner: String, name: String, branch: String)`) through to the image and link fixing methods. It made some of the methods and tests a little more focused as a nice side-effect.

No functional changes.


Sources/App/Controllers/PackageController+routes.swift
```diff
@@ -368,30 +368,28 @@ enum PackageController {
...
+            return PackageReadme.View(model: .noReadme).document()
         }
 
         do {
             let readme = try await Current.fetchS3Readme(req.client, owner, repository)
+            guard let owner = pkg.repository?.owner,
+                  let repo = pkg.repository?.name,
+                  let branch = pkg.repository?.defaultBranch
+            else {
+                return PackageReadme.View(model: .cacheLookupFailed(url: readmeHtmlUrl)).document()
```
finestructure, line 380: I'm using `cacheLookupFailed` here, because it kinda is (we can't look up the cache without them) - although this could never really happen. The problem (that perhaps we could fix one day) is that `owner`, `repo`, and `branch` could be `nil` technically but never really are. That's due to how the fields are modelled in the db, which we could address ultimately.

Sources/App/Controllers/PackageController+routes.swift
```diff
@@ -368,30 +368,25 @@ enum PackageController {
...
-                                                   repositoryName: nil,
-                                                   defaultBranch: nil,
-                                                   readme: nil)).document()
+            return PackageReadme.View(model: .noReadme).document()
         }
 
         do {
             let readme = try await Current.fetchS3Readme(req.client, owner, repository)
+            guard let branch = pkg.repository?.defaultBranch else {
+                return PackageReadme.View(model: .cacheLookupFailed(url: readmeHtmlUrl)).document()
```
finestructure, line 377: I'm using `cacheLookupFailed` here, because it kinda is (we can't hook up the cache without the branch) - although this could never really happen. I still want to model the fields as non-optional in the db one day...

daveverwer (approved):

Alexandr787 (commented):
- [ ] ****

Alexandr787 (commented):
J

PR #2516, opened 2023-07-23
Use package product clause

### LINKED ISSUE #2197: Improve the "Use this Package" snippets
We're currently showing the following info:

CleanShot 2022-12-20 at 12 54 06@2x

However,

```swift
.package(url: "https://github.com/soto-project/soto-s3-file-transfer.git", from: "1.1.0")
```

is only part of what's needed for the package manifest.

In your target, you also need to specify

```swift
.product(name: "SotoS3FileTransfer", package: "soto-s3-file-transfer"),
```

to use the dependency.

We actually have all the information to provide these bits as well, per product. I.e. we could generate a `.product(name: "\(product)", package: "\(package)")` line per product (typically just one).

We can cap this at some number to avoid an overflow for packages with lots of products.

We would _not_ need to do this per significant version, it could be a common section below. (Yes, theoretically the default branch could have different products defined than the stable release but I don't think we need to take that sort of nuance into account.)

finestructure:
There was a [little discussion on the M'don](https://mastodon.social/@finestructure/110753287810928063) about package instructions and that  prompted me to look into the long planned extension of the "Use this package" popover.

I've added in the data we'd need in the model to do this and mocked up the UI bit of it. Obviously needs some tweaking, the biggest thing left to do being the "copyable input". It would need to pull the selected item from the `select` tag via JavaScript - or push it when a selection is made.

CleanShot 2023-07-23 at 17 22 56@2x

Don't think we're far off here and it would make a nice addition to handle installation instructions.

When we ship this, we should look into making this linkable, too. Not in this PR but it'd be nice if package authors could link here for installation instructions.

Fixes #2197


heckj:
Looks great, and I love the copy code snippet - but the other side of this equation isn't represented in this panel - where to put the relevant snippets. (grumble, grumble - wishing Package.swift _wasn't_ code at all) Is there a possibility of showing additional text with snippets of a plain swift package with indicators of where they do?

In my head, I think of it as building a configuration structure - so maybe there's a diagram possibility here, but I'm not sure others see it in the same fashion.

finestructure:
Yeah, I was thinking of adding a larger snippet with a sample manifest or an image with pointers. But that’s perhaps a follow up PR where we make this its own page or something. I think it’s probably too much for a popover.

daveverwer:
I think this is ready for a little more testing:

Screenshot 2023-07-26 at 22 55 16@2x

I tried to keep the vertical height of this panel under control, partly because the manifest section was already big without products, but also in case we decide to add a manifest preview, which will be quite a challenge.

I removed the opening instruction "How you add this package to your project depends on what kind of project you're developing." I also ended up placing the reference labels on the same lines as the snippets, which doesn't work so well for some packages … well, what I mean is it doesn't work very well with Swift Syntax 😂 

Screenshot 2023-07-26 at 23 00 18@2x

It is still functional, though.

I tested it with packages that only have one product and it works OK, although I might give the `select` a `min-width` as it's small with only one product with a short name.

The code should work for packages without products, too, but I've not tested that yet.

daveverwer:
Tested in Chrome, Safari, and Firefox and all look good and are functional.

finestructure:
Looks great. I think keeping the references stacked above the text field would have been fine, too. We're not that constrained for vertical space. Either way, this is a big improvement.

daveverwer (approved):

PR #2531, opened 2023-07-31
Add a macro target count to the Package page

### LINKED ISSUE #2426: Add macros support to the Package page
- [ ] Rethink how the package page reports product types. It was fine to have them separated our when there were two types, acceptable with three (as it is now) but I'm not sure if it works for four or more.
- [ ] Add the number of macros.

Screenshot 2023-06-12 at 08 45 24@2x

daveverwer:
Fixes #2426

Adds a count of macro targets to the Package page alongside libraries, executables, and plugins.

Screenshot 2023-07-31 at 19 23 55@2x

Screenshot 2023-07-31 at 19 24 41@2x

Notes:

* I’m pulling target information in with a separate query similar to what we do with products. You may want us to consolidate some of these queries @finestructure, but I thought it best to get it working this way first so we know what we’re working with.
* I’ve deliberately not added tests for the target query code in case we consolidate the queries. Tests exist for all the other parts of this change, though.
* I am only pulling `macro` targets through to the API level as we do with products. We should probably consider pulling everything through to the API at some point, but that should be consistent across products and targets rather than something we do in this PR.
* This PR does not change the layout, as I suggested it should in #2426. I’d like to see how this feels and will open a new issue if it needs changes.
* I also took the opportunity to fix up the plugins icon, which has been bugging me since I made it, as its “weight“ differs significantly from the other icons here. This probably should have been a separate PR, but it’s such a minor change I bundled it together.


Sources/App/Views/PackageController/GetRoute.Model+ext.swift
```diff
@@ -298,6 +298,17 @@ extension API.PackageController.GetRoute.Model {
         )
     }
 
+    func targetTypeListItem(_ type: Target.TargetType) -> Node<HTML.ListContext> {
+        guard let targets = targets
+        else { print("hello"); return .empty }
```
finestructure, line 303: `print("hello")` eh? 😅

```suggestion
        else { return .empty }
```

finestructure (approved):

Sources/App/Views/PackageController/GetRoute.Model+ext.swift
daveverwer: What do you mean my debugging techniques are primitive and ridiculous? That's a scandalous accusation!

😬

daveverwer:
Deploying as 2.89.8

PR #2555, opened 2023-08-14
Improve local development setup

RemiBardon:
While setting up the project for local development, I made two improvements to the documentation:

- Make `http://127.0.0.1:8080` link clickable for quicker access
- Improve installation steps for the `psql` command
  - @daveverwer already added a note about installing the PostgreSQL CLI in <https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/pull/2543> but it wasn't enough on my machine

### DIFF
--- a/LOCAL_DEVELOPMENT_SETUP.md
+++ b/LOCAL_DEVELOPMENT_SETUP.md
@@ -40,7 +40,11 @@ Close the scheme editor and run the application by selecting "Run" from the Xcod
 
 When working locally, it's helpful to have a database with pre-populated data from the live system. [Talk to us on Discord](https://discord.gg/vQRb6KkYRw), and we'll supply you with a recent database dump that you can load with `./scripts/load-db.sh`.
 
-**Note:** Running the `load-db.sh` script requires that the Postgres command line tools are available on your system. The easiest way to get these is with [brew](https://brew.sh/) by running `brew install postgresql`.
+**Note:** Running the `load-db.sh` script requires that the PostgreSQL command line tools are available on your system and in your `PATH`. The easiest way to get these is with [brew](https://brew.sh/) by running:
+
+```bash
+brew install postgresql@13 && brew install pgcli && brew link libpq --force
+```
 
 ### Setup the Front End
 
--- a/LOCAL_DEVELOPMENT_SETUP.md
+++ b/LOCAL_DEVELOPMENT_SETUP.md
@@ -70,7 +70,7 @@ Or, if you want to run either the CSS or JavaScript linting steps separately, ru
 
 ### Check Everything Works!
 
-Navigate to `http://127.0.0.1:8080` with a web browser, and the site should be up and running locally with CSS and JavaScript working!
+Navigate to <http://127.0.0.1:8080> with a web browser, and the site should be up and running locally with CSS and JavaScript working!
 
 ## Logging database queries
 
--- a/LOCAL_DEVELOPMENT_SETUP.md
+++ b/LOCAL_DEVELOPMENT_SETUP.md
@@ -40,11 +40,7 @@ Close the scheme editor and run the application by selecting "Run" from the Xcod
 
 When working locally, it's helpful to have a database with pre-populated data from the live system. [Talk to us on Discord](https://discord.gg/vQRb6KkYRw), and we'll supply you with a recent database dump that you can load with `./scripts/load-db.sh`.
 
-**Note:** Running the `load-db.sh` script requires that the PostgreSQL command line tools are available on your system and in your `PATH`. The easiest way to get these is with [brew](https://brew.sh/) by running:
-
-```bash
-brew install postgresql@13 && brew install pgcli && brew link libpq --force
-```
+**Note:** Running the `load-db.sh` script requires that the Postgres command line tools are available on your system. The easiest way to get these is with [brew](https://brew.sh/) by running `brew install postgresql`.
 
 ### Setup the Front End
 


LOCAL_DEVELOPMENT_SETUP.md
```diff
@@ -40,7 +40,11 @@ Close the scheme editor and run the application by selecting "Run" from the Xcod
 
 When working locally, it's helpful to have a database with pre-populated data from the live system. [Talk to us on Discord](https://discord.gg/vQRb6KkYRw), and we'll supply you with a recent database dump that you can load with `./scripts/load-db.sh`.
 
-**Note:** Running the `load-db.sh` script requires that the Postgres command line tools are available on your system. The easiest way to get these is with [brew](https://brew.sh/) by running `brew install postgresql`.
+**Note:** Running the `load-db.sh` script requires that the PostgreSQL command line tools are available on your system and in your `PATH`. The easiest way to get these is with [brew](https://brew.sh/) by running:
+
+```bash
+brew install postgresql@13 && brew install pgcli && brew link libpq --force
```
daveverwer, line 46: This is interesting, and certainly `brew install postgresql` was enough when running through it last week. I'm in no way doubting this is what was needed for you, but I wonder if there's more to it. Maybe the `postgresql@13` changed something as I think if you do a `brew install postgresql` you get `@14`.

On my system, the only Postgres package I have installed is `postgresql@14` and I do not have `pgcli` at all.

Did you happen to notice what version of `postgresql` was installed when you tried `brew install postgresql`?

RemiBardon: I am investigating this.

First thing I saw is

```log
$ brew info postgresql
Warning: Formula postgresql was renamed to postgresql@14.
```

Also, `postgresql@13` isn't installed in the `PATH` by default, which may be why I didn't get access to `psql` at first (`...` removing useless output):

```log
brew info postgresql@13
==> postgresql@13: stable 13.12 (bottled) [keg-only]
...
==> Caveats
This formula has created a default database cluster with:
  initdb --locale=C -E UTF-8 /usr/local/var/postgresql@13
For more details, read:
  https://www.postgresql.org/docs/13/app-initdb.html

postgresql@13 is keg-only, which means it was not symlinked into /usr/local,
because this is an alternate version of another formula.

If you need to have postgresql@13 first in your PATH, run:
  echo 'export PATH="/usr/local/opt/postgresql@13/bin:$PATH"' >> ~/.zshrc
...
```

I cleaned up everything:

```log
$ brew uninstall postgresql@13
Uninstalling /usr/local/Cellar/postgresql@13/13.12... (3,244 files, 42.4MB)
$ brew uninstall pgcli
Uninstalling /usr/local/Cellar/pgcli/3.5.0_1... (1,895 files, 15.3MB)
$ brew unlink libpq
Unlinking /usr/local/Cellar/libpq/15.4... 378 symlinks removed.
$ brew cleanup
$ which psql
psql not found
```

Then installed `postgresql@13` again (as the Makefile states `postgres:13.8-alpine` is the version used by the SPI):

```log
$ brew install postgresql@13
...
postgresql@13 is keg-only, which means it was not symlinked into /usr/local,
because this is an alternate version of another formula.

If you need to have postgresql@13 first in your PATH, run:
  echo 'export PATH="/usr/local/opt/postgresql@13/bin:$PATH"' >> ~/.zshrc
...
```

I didn't have access to `psql` so I tried the command proposed by the log:

```log
$ which psql
psql not found
$ export PATH="/usr/local/opt/postgresql@13/bin:$PATH"
$ which psql
/usr/local/opt/postgresql@13/bin/psql
```

In conclusion:

- `postgresql@13` differs from `postgresql@14` as it does not install in the `PATH`
- `pgcli` ("CLI for Postgres with auto-completion and syntax highlighting" says `brew info pgcli`) is not needed

I could update my proposition to replace `brew install pgcli && brew link libpq --force` by `echo 'export PATH="/usr/local/opt/postgresql@13/bin:$PATH"' >> ~/.zshrc`.

However I'm not fan of `>> ~/.zshrc` so I might just revert this proposition and leave your `brew install postgresql`. I don't know if unmatching versions can cause issues.

Tell me what you prefer.

finestructure: It should be fine to use Postgres 14. We're just using the v13 db in the docker image, because that's what our hosted servers are running. The client to interact with it doesn't have to be 13. I'll try an verify this (unless you beat me to it :))

RemiBardon: I would suppose we can always load a dump from an earlier version, so it would not be a problem.

You don't generate dumps locally to load them in production so the other way around doesn't happen.

daveverwer: Thanks so much for your detailed investigation into this. It now makes perfect sense why it worked for me and Cyndi.

The `postgresql` recipe in Homebrew tracks the latest (or not quite latest in this case) version, so I think we should stick with that in the setup document.

As Sven said, all we need on local development machines is for some version of the tools to be installed. It’s vaguely possible that a future version will break compatibility with the version we use as our server, but I’d be surprised if it did.

So, I think the instructions here can stay as is.

Let’s get the `localhost` link merged, though! Thanks again Remi.

finestructure:
I went ahead and reverted the Postgres instructions. This is now what we agreed on, isn't it?

daveverwer (approved):

daveverwer:
> I went ahead and reverted the Postgres instructions. This is now what we agreed on, isn't it?

Yep!

RemiBardon:
Oh yes sorry 🙈

PR #2575, opened 2023-08-22
Added a tooltip to the release and branch dates

daveverwer:
I keep wanting to know when 24h has passed for a package's default branch update date, this means you can hover over it for a precise time. 

The biggest downside of this is that it exposes the server's time to the world and that could confuse people. We could do it as an HTML comment instead to avoid that.


daveverwer:
This doesn't work as I expected it to! Closing

PR #2576, opened 2023-08-22
Analysis error logging detail

finestructure:
In order to understand what's going wrong with the ShellOut update, I'm adding it back in combination with:

- extra logging when there no tags found
- a safe-guard that aborts analysis when no tags are found but versions already exists

NB: this is for deployment to dev only. I'll ad hoc deploy this from the branch and leave the PR in "draft" so we don't push it through. I really just want to see what's in the logs when this runs.


daveverwer (approved):

Sources/App/Commands/Analyze.swift
```diff
@@ -372,6 +372,12 @@ extension Analyze {
             .filter(\.$package.$id == pkgId)
             .all()
         let incoming = try await getIncomingVersions(client: client, logger: logger, package: package)
+        // TODO: temporary safe-guard
+        if !existing.isEmpty && incoming.isEmpty {
+            // Sudden loss of versions is suspicious, warn and throw error
+            logger.error("Suspicious loss of versions for package \(pkgId) - aborting analysis")
+            throw AppError.genericError(pkgId, "Suspicious loss of versions")
+        }
```
finestructure, line 380: This might not even be temporary. I've noticed the following error in our logs:

CleanShot 2023-08-22 at 12 45 53@2x

If for some reason the shell `git tag` command returns just `"` with a status code 0, we'll fail to parse any tags and delete all versions. Our signal for failure is only the return code. I wonder if this is perhaps what's going wrong - we're getting nonsense from the command with a return code of 0.

Ideally we'd be doing our git stuff with a library to avoid this shell command hackery, but that's for another day.

It should be really rare and is highly likely a parsing error if a package loses all its releases. If we can't use a library to determine the tags, perhaps there are other ways to ensure we get all the tags. Perhaps the logs will tell us what exactly the shell is returning and solve this in the interim.

Sources/App/Commands/Analyze.swift
```diff
@@ -401,22 +409,34 @@ extension Analyze {
...
 
-        let defaultBranch = package.repository?.defaultBranch
-            .map { Reference.branch($0) }
-
-        let tags = try Current.git.getTags(cacheDir)
+        guard let defaultBranch = package.repository?.defaultBranch
+            .map({ Reference.branch($0) })
+        else {
+            throw AppError.genericError(package.model.id, "Package must have default branch - aborting analysis")
+        }
```
finestructure, line 416: This could have been the cause of default branch deletion: if default branch is nil, we omitted it in the list and scheduled it for deletion. However, that's quite a stretch. While our db field does allow null for historical reasons, there are no records that have it as null. Of course, that doesn't mean it never was.

finestructure:
Closing in favour of #2592

PR #2584, opened 2023-08-26
Fix versions latest update

finestructure:
Just leaving a note here that this whole fix should not be required, because the entire analysis section is wrapped in a transaction. So it shouldn't be possible for the `applyVersionDelta` to be committed without the `updateLatestVersions` also running. Reordering these calls should not make any difference whatsoever and if it does, it means the transaction isn't working as intended.

⛔ do not merge until this is understood ⛔

### DIFF
--- a/Sources/App/Commands/Analyze.swift
+++ b/Sources/App/Commands/Analyze.swift
@@ -220,6 +220,10 @@ extension Analyze {
                 }
 
                 try await applyVersionDelta(on: tx, delta: versionDelta)
+                do {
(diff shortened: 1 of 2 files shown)


daveverwer (approved):

finestructure:
Ok, this is definitely not a new problem. I shot myself in the foot in revision 5d13688873b460da1fbd984d73b4209c57b46835 where I _actively worked to disable the transaction_ for some bizarre reason:

CleanShot 2023-09-04 at 13 03 24@2x

Issue #317, opened 2020-06-07
Does a package include tests?


daveverwer:
If it does

* The score should be higher
* There should be a visual indicator of this on the package page

Precise details of how this will work still to be discussed, but this is the basics of the feature.

Sherlouk:
What is the source here? A few random examples (not saying I actually support all of these but good for conversation!)

* Code Coverage?
  * This could be picked up if the library supports something like Coveralls, but that's not a given?
* Number of Test Files?
  * Possibly a ratio of test files to source files?
* The existence of a test target/directory?

Should a package's score increase if it doesn't have CI running the tests? Perhaps +X points if it has tests +Y points if it also runs them on CI...

mattpolzin:
We could establish a precedent (much like that of “put a file in your repository named LICENSE.txt) or committing the JSON output of `swift test --enable-code-coverage` to the repository with a certain name. That’s a pain, though.

Sherlouk:
Had a brief conversation with Dave on Slack, it seems there may be a longer-term ambition to actually build and test packages automatically - so would likely end up calculating the code coverage from that.

Much to your sentiment, encouraging an entire community to commit coverage information like that is unlikely to gain massive traction IMO

mattpolzin:
> longer-term ambition to actually build and test packages automatically

That’s ambitious! Would be very cool.

mattpolzin:
Building on the idea that packages will be built by the indexer, it would be relatively easy to get the code coverage percentage (whether to display correctly, or even just to give each package some sort of letter grade specific to coverage). 

Maybe not even a letter grade because code coverage can mean different things for different packages, but something like "Test Coverage: Stellar" and similar. 

Subjective criteria like `0% -> "No Coverage"`, `<50% -> "Limited Coverage"`, `50%-75% -> "Moderate Coverage"`, `75%-85% -> "Decent Coverage"`, `85%-97% -> "Good Coverage"`, `97%-100% -> "Stellar Coverage"`.

Anyway, if useful, determining the percentage is a small lift (and barring other plans, feel free to use my library or even just snag the very small amount of code that currently is capable of deserializing the codecov JSON output by `swift test`: https://github.com/mattpolzin/swift-test-codecov/blob/master/Sources/SwiftTestCodecovLib/CodeCov.swift).

MaximBazarov:
[...]

PR #2588, opened 2023-08-29
add test target to package score

cyndichin:
Fixes issue #317

Discussed with @daveverwer

### DIFF
--- a/Sources/App/Commands/Analyze.swift
+++ b/Sources/App/Commands/Analyze.swift
@@ -239,16 +239,13 @@ extension Analyze {
 
             let versions = try await updateLatestVersions(on: tx, package: package)
 
-            let targets: [(String, TargetType)]? = try await {
-                guard let repo = package.repository,
-                      let owner = repo.owner,
-                      let repository = repo.name
-                else { return nil }
-
-//                let thisLineCausesProcessingTestFailure = try await API.PackageController.Target
-//                    .query(on: database, owner: owner, repository: repository)
-
-                return []
+            let targets: [(String, TargetType)]? = await {
+                guard let repo = package.repository, let owner = repo.owner, let repository = repo.name else {
+                    return nil
+                }
+                do {
+                    return try? await API.PackageController.Target.query(on: tx, owner: owner, repository: repository)
+                }
             }()
 
             updateScore(package: package, versions: versions, targets: targets)
--- a/Sources/App/Commands/Analyze.swift
+++ b/Sources/App/Commands/Analyze.swift
@@ -242,7 +242,7 @@ extension Analyze {
             let targets = await fetchTargets(on: tx, package: package)
 
             updateScore(package: package, versions: versions, targets: targets)
-            
+
             await onNewVersions(client: client, logger: logger, package: package, versions: newVersions)
         }
     }
@@ -259,7 +259,7 @@ extension Analyze {
         return try? await API.PackageController.Target.query(on: database, owner: owner, repository: repository)
     }
 
-    
+
     static func createCheckoutsDirectory(client: Client,
                                          logger: Logger,
                                          path: String) async throws {
--- a/Sources/App/Commands/Analyze.swift
+++ b/Sources/App/Commands/Analyze.swift
@@ -238,23 +238,28 @@ extension Analyze {
             }
 
             let versions = try await updateLatestVersions(on: tx, package: package)
-
-            let targets: [(String, TargetType)]? = await {
-                guard let repo = package.repository, let owner = repo.owner, let repository = repo.name else {
-                    return nil
-                }
-                do {
-                    return try? await API.PackageController.Target.query(on: tx, owner: owner, repository: repository)
-                }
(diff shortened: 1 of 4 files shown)


Sources/App/Commands/Analyze.swift
```diff
@@ -239,8 +239,20 @@ extension Analyze {
...
 
-            updateScore(package: package, versions: versions)
+            let targets: [(String, TargetType)]? = try await {
+                guard let repo = package.repository,
+                      let owner = repo.owner,
+                      let repository = repo.name
+                else { return nil }
+
+//                let thisLineCausesProcessingTestFailure = try await API.PackageController.Target
+//                    .query(on: database, owner: owner, repository: repository)
```
daveverwer, line 249: So, I pushed a debugging commit in 4dec651f8735fd729895305ff8a7bf84f770ccfe because I've narrowed the test failure down to this line of code. What we actually want to do here is return these target name/type pairs, but just executing that query causes `test_processing_pipeline` to fail.

With the commented-out line included, the test fails (even still returning the `[]`). When it’s commented out, the test passes.

I’ve traced that query back to its root, and it’s just a `SELECT` query. I can’t imagine how it’d affect the processing test in the way it is, except maybe a timing issue from adding the extra time for that query?

Any ideas, @finestructure?

finestructure: The problem is that this query is throwing (with `connectionRequestTimeout` for me, that's with the db up) and the test shows how this would impact processing. At the very least this should be

```swift
            let targets: [(String, TargetType)]? = await {
                guard let repo = package.repository, let owner = repo.owner, let repository = repo.name else {
                    return nil
                }
                do {
                    return try? await API.PackageController.Target.query(on: database, owner: owner, repository: repository)
                }
            }()

```

The timeout is happening, because it's using `database` and we're inside a transaction - leading to a deadlock until it times out.

So it needs to be

```swift
            let targets: [(String, TargetType)]? = await {
                guard let repo = package.repository, let owner = repo.owner, let repository = repo.name else {
                    return nil
                }
                do {
                    return try? await API.PackageController.Target.query(on: tx, owner: owner, repository: repository)
                }
            }()
```

(I'd also avoid the immediately executing close here but I suspect that's just WIP.)

cyndichin: thank you @daveverwer and @finestructure for debugging! 

@finestructure would you be able to clarify what you mean by the following? I would like to address it, but want to make sure I understand what is preferred.

> (I'd also avoid the immediately executing close here but I suspect that's just WIP.)

cyndichin: thank you @finestructure for the detail writeup and helping to improve this PR! this definitely helps clarify your thought process with me and I definitely agree with the points made. I also prefer option 1, so I updated the PR to use a private static function instead. let me know if there's anything else that should be addressed :)

Sources/App/Core/Score.swift
```diff
@@ -85,11 +86,13 @@ enum Score {
             case 5..<20: score += 5
             default: score += 10
         }
+        
+        if candidate.hasTestTargets { score += 15 }
```
daveverwer, line 90:
```suggestion
        if candidate.hasTestTargets { score += 5 }
```

Before merging this, I did a little experimentation with the newly analysed target data on production.

We have tagged **4,632** (~75%) packages as having a test target in their current version. Of course, this is so high because, depending on how the package author created the package, it might have included a default test target at creation time.

For example: https://github.com/iballan/BasicServiceLocator/blob/master/Tests/BasicServiceLocatorTests/BasicServiceLocatorTests.swift

The good news is that it took me quite a while to find a package with this default template test intact and unedited. This is going to lift some packages with a test target and without tests, but it’s also going to lift all the packages with a test target _and_ tests.

As we have discussed, there is more in-depth analysis we could do to find other metadata that would give a better sense of whether tests exist. We will tackle that as a separate PR.

If this number were 99% or even 90%, I’d want to consider whether this PR was making a helpful contribution. At 75%, I want to merge it, but 15 points is too high a score adjustment for this level of analysis. I suggest we adjust it to 5 points.

Tests/AppTests/ScoreTests.swift
```diff
@@ -153,8 +168,29 @@ class ScoreTests: AppTestCase {
...
+        XCTAssertEqual(Score.compute(.init(licenseKind: .compatibleWithAppStore,
+                                           releaseCount: 20,
+                                           likeCount: 20_000,
+                                           isArchived: false,
+                                           numberOfDependencies: 2,
+                                           lastActivityAt: Current.date().adding(days: -10),
+                                           hasDocumentation: false,
+                                           numberOfContributors: 0,
+                                           hasTestTargets: true)),
+                       122)
```
daveverwer, line 193: These tests will need fixing up to match the new scoring.

daveverwer (approved):
Thank you so much for this @cyndichin!

daveverwer:
🚀

Issue #2571, opened 2023-08-22
Version rediscovery incident


finestructure:
We've had a serious issue last night where we suddenly started re-discovering old releases of packages ([prod-alert channel notes](https://discord.com/channels/844491846452379649/1082208064515604610/1143290390590541955)).

![CleanShot_2023-08-21_at_23 00 372x](https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/assets/65520/bcd20bba-35e4-4907-bd25-2574e66a2199)
![CleanShot_2023-08-21_at_23 13 412x](https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/assets/65520/9fc96663-8f7f-4f9d-815e-7619d0c675b0)

I believe I understand mostly how this happened now, although the underlying reason is still unclear. As is often the case, this is (at least) two issues conspiring.

We deployed 2.89.16 at 17:00 CET. There's no unusual activity with respect to version discovery around that time, everything looks normal. However, going back in the logs noticed a new (?) error message: `[ WARNING ] The operation could not be completed. (App.GitError error 2.) [component: analyze]`.

`App.GitError error 2` is `invalidRevisionInfo`, which is thrown when we get bad revision info for a package's tags. Unfortunately, and this is the first bug, this leads to no tag versions to be discovered at all - which leads to a version diff where all existing tags are deleted.

This explains why the new version discovery only started happening ~3h after deployment: it's the _second_ pass of analysis that then rediscovers the tags (presumably, this `invalidRevisionInfo` error is intermittent).

That also explains why the rollback to 2.89.15 had no effect - the damage was already done, although it probably prevented further damage (version deletions) being done.

One thing that's not clear to me is how the ShellOut change in 2.89.16 could affect version discovery.

I wonder if #2567 is the reason. We've been getting crashes in analysis at times, which we may now be converting into properly thrown exceptions. So before the crash saved us from running into this version discovery bug. Now we throw instead of crashing and end up deleting valid tag versions.

We never noticed the crashes impacting processing beyond the error messages, because the error seems to be intermittent and it'll just sort itself out on the next pass.

finestructure:
Actions:

- [ ] Leave analysis disabled until we have a fix for the bad error handling.
[...]

PR #2594, opened 2023-09-04
Fix version transaction

finestructure:
This fixes the default branch `latest` flag flickering we observed as part of #2571 by properly re-instating the db transaction we had in place to avoid exactly this problem.

I broke this in 5d13688873b460da1fbd984d73b4209c57b46835 during the async/await rework.

We now have a regression test in place that will prevent this from breaking again.


finestructure:
Placing in draft just to avoid merging this until I'm done verifying that current main that's running staging is OK and can be deployed.

daveverwer (approved):

PR #2600, opened 2023-09-05
Add support for InjectionIII

daveverwer:
I've been chatting with @johnno1962 recently about Injection and how it might solve my annoying workflow with the HTML generation through the Vapor app while also working with CSS and/or JavaScript.

CSS/JavaScript files re-build immediately as soon as any of the SCSS or JS files change on disk, and so it's a case of saving a file, switching to the browser, refreshing the page, and seeing the result.

With HTML changes, it's a case of making the change in the Plot code, saving the file, forgetting to rebuild, switching to the browser and refreshing the page, wondering why nothing has changed, banging my head against the desk, and then switching back to Xcode to build and run, then refresh the browser. 😂

I've been back and forth with John, and with the latest pre-release build, this now works for our project. It doesn't work for tests, unfortunately, but it works for `App` classes, which is what we need for this workflow. 🎉

This PR does two things:

1. If the InjectionIII app is installed in `/Applications` it loads a bundle during `configure`. If the app is not installed, there’s no additional output, and the app is (obviously) not loaded.
2. Adds a linker setting of `-interposable` when building in debug.

I’m not concerned about the impact of point 1, although we could conditionally compile it only in debug to have both belt and braces.

I think the `interposable` flag is there to allow symbols to be overridden, and is only used in `debug` builds, which should only affect local development. I don’t know of any direct security implications of that, but I want to double-check that we don’t run a `debug` build on staging. I don’t think we do, but worth checking.


daveverwer:
Looks like the linker is failing in CI, too.

Screenshot 2023-09-05 at 19 50 45@2x

johnno1962:
This was a problem with adding `-interposable` with Xcode 15 up for beta 5

daveverwer:
> This was a problem with adding `-interposable` with Xcode 15 up to beta 5. You can leave it out but you won't be able to inject final methods, struct methods or functions.

Our CI runs on Linux, and I believe Sven did an update to the latest 5.9 nightly today or yesterday. It could be an issue of just Linux though.

johnno1962:
`-interposable` is a Apple linker thing. You Probably want:
```
    linkerSettings: [.unsafeFlags(["-Xlinker", "-interposable"],
              .when(platforms: [.macOS], configuration: .debug))]

**Comment by daveverwer** [comment]
> `.when(platforms: [.macOS], configuration: .debug))]`

Ah, that's great John. Thank you!

File: Sources/App/configure.swift [hunk 1]
```diff
@@ -18,6 +18,10 @@ import Vapor
 
 @discardableResult
 public func configure(_ app: Application) throws -> String {
+    // The bundle is only loaded if /Applications/InjectionIII.app exists on the local development machine.
+    // Requires InjectionIII 4.7.3 or higher to be compaible with Package.swift files.
+    let _ = Bundle(path: "/Applications/InjectionIII.app/Contents/Resources/macOSInjection.bundle")?.load()
```
finestructure, line 23: Since this is all for debug config only, it'd probably best to also wrap this in `#if DEBUG`, just to make that part explicit.

finestructure (approved):

Sources/App/configure.swift
finestructure: Or even `#if DEBUG && os(macOS)` actually.

daveverwer: I was also thinking we might want to gate it on an environment variable (or similar), too, so that it's possible to turn it off temporarily without uninstalling/moving the app.

johnno1962: You can set the INJECTION_DAEMON environment variable to force using the app which if you're not running it will disable injection.

daveverwer: Thanks John! I added a comment to let people know about that and it works great for me.

johnno1962: Using the App may better for you if you're not using the Xcode editor. If the code doesn't compile you get a change from orange (ready) to yellow (error) of the icon on the menu bar. The other states are not connected (blue) and compiling (compiling),

PR #2603, opened 2023-09-06
Cope with case mismatches in the package denylist

daveverwer:
This doesn't cover a real world error, yet, but going back to belt and braces. 🦺

### DIFF
--- a/Sources/App/Commands/Reconcile.swift
+++ b/Sources/App/Commands/Reconcile.swift
@@ -117,5 +117,6 @@ func reconcileLists(db: Database, source: [URL], target: [URL]) async throws {
 }
 
 func processPackageDenyList(packageList: [URL], denyList: [URL]) -> [URL] {
-    return Array(Set(packageList).subtracting(Set(denyList)))
+    let lowercaseDenyList = denyList.map { $0.absoluteString.lowercased() }
+    return Array(packageList.filter { lowercaseDenyList.contains($0.absoluteString.lowercased()) == false })
 }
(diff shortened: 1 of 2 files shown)


Sources/App/Commands/Reconcile.swift
```diff
@@ -117,5 +117,6 @@ func reconcileLists(db: Database, source: [URL], target: [URL]) async throws {
 }
 
 func processPackageDenyList(packageList: [URL], denyList: [URL]) -> [URL] {
-    return Array(Set(packageList).subtracting(Set(denyList)))
+    let lowercaseDenyList = denyList.map { $0.absoluteString.lowercased() }
+    return Array(packageList.filter { lowercaseDenyList.contains($0.absoluteString.lowercased()) == false })
```
finestructure, line 121: It'd be better to use case-insensitive hashing to avoid the quadratic filter:

```swift
func processPackageDenyList(packageList: [URL], denyList: [URL]) -> [URL] {
    struct CaseInsensitiveURL: Equatable, Hashable {
        var url: URL

        init(_ url: URL) {
            self.url = url
        }

        func hash(into hasher: inout Hasher) {
            hasher.combine(url.absoluteString.lowercased())
        }

        static func == (lhs: Self, rhs: Self) -> Bool {
            lhs.url.absoluteString.lowercased() == rhs.url.absoluteString.lowercased()
        }
    }

    return Array(
        Set(packageList.map(CaseInsensitiveURL.init))
            .subtracting(Set(denyList.map(CaseInsensitiveURL.init)))
    ).map(\.url)
}
```

daveverwer: Integrated this. Thanks!

finestructure (approved):

PR #2607, opened 2023-09-08
Cope with visionOS platforms being specified in the Package manifest

### LINKED ISSUE #2605: SwiftDataKit missing all version records

- https://swiftpackageindex.com/fatbobman/SwiftDataKit

I searched for this package on the index and thought it wasn't in the package list, but [it is](https://github.com/SwiftPackageIndex/PackageList/blob/60177a6d47cd78d156d94fff397346f3866fb8b3/packages.json#L1871C4-L1871C49)

Screenshot 2023-09-07 at 13 03 30@2x

Looking at the data, it's in `packages`, but the status is set to `no_valid_versions` and there are 0 version records associated with that package ID in `versions`.

Screenshot 2023-09-07 at 13 07 45@2x

I checked the package out locally and ran a `dump-package` and looked at the tags and git repository and everything looks fine. 

It'll probably recover with another analysis cycle, but it's very odd that it hasn't picked up the `0.1` tagged version since it was tagged 13 hours ago so I wanted to make sure we logged it somewhere, especially after our recent issues.

daveverwer:
Fixes #2605

The issue was [this line](https://github.com/fatbobman/SwiftDataKit/blob/4a476fc8ad2ce1ded20082e697226a2f2cebbf81/Package.swift#L13) in the `SwiftDataKit` package manifest. It’s the first time we’ve seen a `visionOS` platform explicitly declared.

I checked [SupportedPlatforms.swift](https://github.com/apple/swift-package-manager/blob/main/Sources/PackageDescription/SupportedPlatforms.swift) and we’re now up to date again with nothing missing.

When it comes to testing this, we could add a test to `ManifestTests` to verify that we never remove `visionos` or any of the other cases, but I don’t think we can add anything sensible that will catch this error for future platforms. I don’t think a test makes sense here, unless you have an idea for one @finestructure?

### DIFF
--- a/Sources/App/Models/Manifest.swift
+++ b/Sources/App/Models/Manifest.swift
@@ -45,6 +45,7 @@ struct Manifest: Decodable, Equatable {
             case macos
             case openbsd      // from 5.8
             case tvos
+            case visionos     // from 5.9
             case wasi         // from 5.3
             case watchos
             case windows      // from 5.2
--- a/Sources/App/Models/Platform.swift
+++ b/Sources/App/Models/Platform.swift
@@ -27,6 +27,7 @@ struct Platform: Codable, Equatable {
         case macos
         case openbsd
         case tvos
+        case visionos
         case wasi
         case watchos
         case windows
@@ -71,6 +72,8 @@ extension Platform: CustomStringConvertible {
                 return "OpenBSD \(version)"
             case .tvos:
                 return "tvOS \(version)"
+            case .visionos:
+                return "visionOS \(version)"
             case .wasi:
                 return "WASI \(version)"
             case .watchos:


Sherlouk:
I'll ask a potentially stupid question, why does this need to be typed and not just a string?

What about an .unknown case?

What about an associated type fallback .unhandled(String)

Don't get me wrong I don't see Apple adding a new platform for a while, but...

finestructure:
Unfortunately, SwiftPM's `Platform` isn't an enum or we could exhaustively check we handle all its cases. This is the best we can do as far as I can see.

To @Sherlouk's question: I'd rather we have the proper values in the db at the expense of one issue upfront than end up with various unstructured `unhandled` cases, which we then might have to (or at least want to) clean up afterwards.

finestructure (approved):

Sherlouk:
@finestructure, a valid point and one which rules out two of the four proposed solutions. I still don't think it rules out simply storing it as a string from the start - you're not doing any transformation here from what I can tell?

finestructure:
Not yet, no.

PR #2619, opened 2023-09-15
Increase ANALYZE_SLEEP to 10s to reduce db load

finestructure:
Not really a problem per se but let's try and keep the load a bit lower, we don't need to rush these packages through.

CleanShot 2023-09-15 at 12 46 31@2x

CleanShot 2023-09-15 at 12 54 15@2x

### DIFF
--- a/app.yml
+++ b/app.yml
@@ -122,7 +122,7 @@ services:
       - migrate
     entrypoint: ["/bin/bash"]
     command: ["-c", "--",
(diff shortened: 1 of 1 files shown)


finestructure:
I've tested this by deploying an override:

CleanShot 2023-09-15 at 14 05 40@2x

Lower db utilisation and more even CPU utilisation.

finestructure:
I'll go ahead and self-merge this trivial change.

PR #2622, opened 2023-09-17
Issue 2616 older documentation

### LINKED ISSUE #2616: Documentation showing old version

- https://swiftpackageindex.com/pointfreeco/swift-composable-architecture

I noticed that TCA's docs nicely include the 0.x and 1.x series! Hope this is intentional, as it's a nice way of keeping old docs around :)

One thing I noticed, though, is that the 0.x docs are pointing to 0.57.0, but the latest tag for 0.x is [0.59.0](https://github.com/pointfreeco/swift-composable-architecture/releases/tag/0.59.0).

finestructure:
The difficulty with re-generating documentation for older tagged versions is that we aggressively trim builds that don't belong to significant versions (default branch, release, pre-release versions).

When we trigger a build for an outdated version, the build stub (build with status 'triggered') in the builds table gets "garbage collected" almost immediately after triggering.

This is not a big problem: build reporting handles this case and does an upsert when it's reporting the build status back. If the record exists (the expected case), it gets updated. Otherwise a new record is being created.

The problem is that documentation results reporting does not do the same: it only updates, because it does not have enough information to re-create the build record. It also typically runs right after a build command, so it's the normal case to assume the build record exists.

However, for old releases, this gap is too large: the triggered build record gets deleted immediately, then the build report recreates it, and then it gets trimmed again right away before the doc report has a chance to come in.

It's unfortunately not an option for the doc report to recreate the build record (as mentioned above).

So what we have to do instead is loosen our build trimming logic.

Currently, it deletes all non-significant version builds immediately, regardless their age and all triggered or infrastructure error builds for significant version builds that are older than the `trimBuildsGracePeriod` of 4 hours.

This PR changes this to apply the `trimBuildsGracePeriod` also to non-significant version builds. I.e. instead of being deleted immediately, they are only deleted if they're older than 4 hours.

This gives the documentation process a chance to run and report back successfully before the build gets deleted.

NB: The build does not have to be around for docs to be accessible. The doc report sends doc archive info that is saved on the version. This change allows it to progress to that point.

I will roll this change out on dev from the branch and test it end-to-end with the Pointfree.co package before marking this PR ready.

Fixes #2616

### DIFF
--- a/Sources/App/Commands/TriggerBuilds.swift
+++ b/Sources/App/Commands/TriggerBuilds.swift
@@ -476,11 +476,17 @@ func trimBuilds(on database: Database) async throws -> Int {
         FROM builds b
         USING versions v
         WHERE b.version_id = v.id
+        AND b.created_at < NOW() - INTERVAL '\(raw: String(Constants.trimBuildsGracePeriod.inHours)) hours'
         AND (
-          v.latest is null
-          OR (
+          (
+            -- significant version: delete only old builds that are triggered or infrastructureError builds
+            v.latest IS NOT NULL
+            AND
             b.status IN ('\(raw: Build.Status.triggered.rawValue)', '\(raw: Build.Status.infrastructureError.rawValue)')
-            AND b.created_at < NOW() - INTERVAL '\(raw: String(Constants.trimBuildsGracePeriod.inHours)) hours'
+          )
+          OR (
+            -- non-significant version: delete all old builds
+            v.latest is null
           )
         )
         RETURNING b.id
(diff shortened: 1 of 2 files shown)


daveverwer (approved):

finestructure:
Ok, this is now working as expected on dev. After generating [docs for 0.58.0](https://staging.swiftpackageindex.com/pointfreeco/swift-composable-architecture/0.58.0/documentation):

CleanShot 2023-09-17 at 17 47 19@2x

and subsequently [0.58.1](https://staging.swiftpackageindex.com/pointfreeco/swift-composable-architecture/0.58.1/documentation), [0.58.2](https://staging.swiftpackageindex.com/pointfreeco/swift-composable-architecture/0.58.2/documentation), and [0.59.0](https://staging.swiftpackageindex.com/pointfreeco/swift-composable-architecture/0.59.0/documentation):

CleanShot 2023-09-17 at 17 55 09@2x

PR #2653, opened 2023-10-02
default 15d → 30d

finestructure:
I want to make sure we don't lose the start of this graph while we're still investigating the capacity issues:

CleanShot 2023-10-02 at 16 15 25@2x


finestructure:
ad hoc deploying on dev to test the command works

daveverwer (approved):

finestructure:
Working on dev, merging and deploying.

PR #2656, opened 2023-10-05
Remove task group

finestructure:
I deployed this version to dev yesterday morning at 10:14 CET and there hasn't been a hang since. Previously there were constant hangs (within an hour of re-deploying or restarting analysis) so this is hopefully indication that this fixes the hangs.

It's hard to judge the performance impact, because analysis hasn't been running properly on dev for quite a while. However, just comparing to prod it doesn't look too bad (maybe a factor of 2 - but dev is a slower machine).

Given how often we lose analysis for hours when I'm not around to restart it this seems like a good trade-off even if it was taking twice as long for a full cycle.

CleanShot 2023-10-05 at 08 18 50@2x

CleanShot 2023-10-05 at 08 19 52@2x

### DIFF
--- a/Sources/App/Commands/Analyze.swift
+++ b/Sources/App/Commands/Analyze.swift
@@ -183,18 +183,11 @@ extension Analyze {
             try await createCheckoutsDirectory(client: client, logger: logger, path: checkoutDir)
         }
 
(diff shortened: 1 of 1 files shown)


daveverwer (approved):
Well this looks promising!

PR #2664, opened 2023-10-12
Add readme check to score

cyndichin:
Adds points to the package score if there exists a README url

Discussed with @daveverwer

Thank you @maxxfrazer for the inspiration referenced here: https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/discussions/2591

### DIFF
--- a/Sources/App/Core/Score.swift
+++ b/Sources/App/Core/Score.swift
@@ -23,6 +23,7 @@ enum Score {
         var numberOfDependencies: Int?
         var lastActivityAt: Date?
         var hasDocumentation: Bool
+        var hasReadme: Bool
         var numberOfContributors: Int
         var hasTestTargets: Bool
     }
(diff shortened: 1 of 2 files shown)


daveverwer (approved):
To give a bit of public commentary on the `+15` score change in this PR, there are 54 packages without any README file at all, so this is effectively a 15-point lift for *every* package. That said, I don’t believe that what becomes effectively a 15-point penalty is too harsh for those 54 packages that have no README at all. Having no README file at all is a strong signal that the package is not for others to use.

15 points is as strong a signal as having documentation, which is something we put thought into, too. We see them as quite similar, with many package authors using a README file for at least documentation. Neither this nor our documentation score makes any attempt at calculating the quality of README *or* documentation.

Of course, much more can also be done in this area, such as judging the complexity/content in a README and/or documentation. That is a bigger change, and may also include a tweak to these scores if/when we implement it.

daveverwer:
Thanks so much for this @cyndichin! More great work 👍

PR #2688, opened 2023-11-01
Gate the package score breakdown to staging until we announce it

daveverwer:
Thinking on release/announcement planning, this would be better going live in production next week along with the blog post.

So close, and yet so far, @cyndichin! 🫨


finestructure (approved):

daveverwer:
Deploying as 2.92.3

PR #2699, opened 2023-11-06
Filter documentation sitemaps with an allow list

daveverwer:
I have included a selection of packages, but nothing with more than 5,000 URLs. This should take us from having combined sitemaps with almost 460,000 URLs, down to something with less than 20,000 (including Package page URLs).


finestructure (approved):

daveverwer:
Deploying as 2.92.4 to test on production as sitemaps are all disabled on staging.

PR #2700, opened 2023-11-07
Completely disable site maps on staging

daveverwer:
We previously had site maps available in all environments, but just to be double sure we don't ever serve them in a staging environment, this PR completely disables them outside production.

It was a bit risky having them before, we didn't point directly at them, but if any search engine happened to request `/sitemap.xml`, they would have seen thousands of incorrect URLs.

This also adds a new SiteMapView type to slightly separate out the queries and controller processing from the site map rendering.

### DIFF
--- a/Sources/App/Controllers/PackageController+routes.swift
+++ b/Sources/App/Controllers/PackageController+routes.swift
@@ -320,10 +320,10 @@ enum PackageController {
             [String]()
         }
 
-        return try await SiteMapView.packageSiteMap(owner: packageResult.repository.owner,
-                                                    repository: packageResult.repository.name,
-                                                    lastActivityAt: packageResult.repository.lastActivityAt,
-                                                    linkablePathUrls: urls).encodeResponse(for: req)
+        return try await SiteMapView.package(owner: packageResult.repository.owner,
+                                             repository: packageResult.repository.name,
+                                             lastActivityAt: packageResult.repository.lastActivityAt,
+                                             linkablePathUrls: urls).encodeResponse(for: req)
     }
 
     static func linkablePathUrls(client: Client, packageResult: PackageResult) async -> [String] {
--- a/Sources/App/Controllers/PackageController+routes.swift
+++ b/Sources/App/Controllers/PackageController+routes.swift
@@ -287,7 +287,7 @@ enum PackageController {
         return response
     }
 
-    static func siteMap(req: Request) async throws -> Response {
+    static func siteMap(req: Request) async throws -> SiteMap {
         guard
             let owner = req.parameters.get("owner"),
             let repository = req.parameters.get("repository")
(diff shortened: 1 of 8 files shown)


finestructure:
Sorry, it took a little longer to follow along with the changes.

It's been a while since I looked at the sitemap related stuff and that's probably why I'm a little confused regarding the three "routing types" we seem to have now:

```swift
enum PackageController {
    static func siteMap(req: Request) async throws -> SiteMap { ... }
}

enum SiteMapView {
    static func index(packages: [SiteMapController.Package]) -> SiteMapIndex { ... }
    static func staticPages() -> SiteMap { ... }
    static func package(owner: String?,
                        repository: String?,
                        lastActivityAt: Date?,
                        linkablePathUrls: [String]) async throws -> SiteMap { ... }
}

enum SiteMapController {
    static func index(req: Request) async throws -> SiteMapIndex { ... }
    static func staticPages(req: Request) async throws -> SiteMap { ... }
}
```

Should `SiteMapController` perhaps absorb `SiteMapView`? In fact `SiteMapController.statisPages()` is

```swift
    static func staticPages(req: Request) async throws -> SiteMap {
        return SiteMapView.staticPages()
    }
```

That seems odd. It looks like `SiteMapView` is the router for the individual pages and `SiteMapController` is the one for the index. `PackageController.siteMap` does part of the routing that should probably be in `SiteMapView`. Or maybe `SiteMapController` should just handle all the routing in one place with appropriate overloads:

```swift
enum SiteMapController {
    static func index(req: Request) async throws -> SiteMapIndex { ... }
    static func package(req: Request) async throws -> SiteMapIndex { ... }
    static func staticPages(req: Request) async throws -> SiteMap { ... }
}
```

Effectively merge `SiteMapView` into `SiteMapController` since it's mostly calling into `SiteMapView` anyway (like also from `SiteMapController.index`).

Does that make sense? Or am I misreading the structure and the intent here?

finestructure:
And can play around and make the change I have in mind to see if it actually makes sense and in case that makes the discussion easier.

daveverwer:
I think what you're proposing is pretty close to things are currently organised, before this PR. The controller methods were on different controllers because the package-specific site map needed a few things that were more easily available there, and since it hangs off the package route `owner/repo/sitemap.xml`, it makes sense to live on the `PackageController`.

I made the new `SiteMapView` type introduced here so that we would at least have all of the actual site map XML generation across those controller methods in one place, and with the exception of the static site map, which needs no logic at all, the separation of XML generation from site map XML generation felt significant enough to split into a new type.

We can remove that from this PR if you think it makes sense all in the various controllers. The important bit is the change from `Response` to `SiteMap` for easier testing, and the blocking of the routes in production.

finestructure (approved):

finestructure:
Let's merge this and I'll propose my changes tomorrow if they actually work out.

PR #2701, opened 2023-11-07
Renewed package collection signing certificate

daveverwer:


finestructure (approved):

daveverwer:
Do you have an easy way to run this on Linux to update the `test_sign_collection.linux.json` snapshot, @finestructure?

finestructure:
Yes, I can do that tomorrow no worries.

finestructure:
Let's see if this passes - the process ended up being a lot more complicated than it used to be. I'll explain once this is working.

finestructure:
Ok, so this worked, that's good.

Now to the problem: it appears that with Swift 5.9 we've lost the ability to compile SPI-Server on arm64/Linux. I had to go back to the last Swift 5.8 version 2.91.4 and add the new cert there to get it to generate the new signature.

Alternatively, I could have used our linux builder but I wanted to see if I can get this to run locally.

We'll want to take this to the Swift team, I don't think this compilation failure is project specific. I'll open an issue in our repo to track it and to figure out where to take the details.

It might make sense to set up a CI stage to at least build the project on arm64/Linux to make sure this doesn't break in the future.

daveverwer:
Thanks for sorting it out! Sorry it was trickier than expected 😬

daveverwer:
Deploying as 2.92.5 as our collections are currently broken 🫨

Issue #2711, opened 2023-11-13
Show funding/sponsorship information on Packge pages


daveverwer:
First suggested by @MaxDesiatov in [this comment](https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/issues/202#issuecomment-640173645), we would like to show whether packages accept funding or sponsorship, but it never got turned into a full-blown issue.

We should parse the `FUNDING.yml` file in the root of a package repository and display links to all the sites configured in that file on the Package page. More details on the format of that file are available in [the GitHub documentation](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/displaying-a-sponsor-button-in-your-repository).

The only shame here is that there is no text description of the sites in that file, so we will either have to use raw URLs or do our best with the ones that are not just custom URLs.

PR #2712, opened 2023-11-13
Fetch funding information and store in the database

daveverwer:
⚠️ SCHEMA CHANGE ⚠️

Progress towards [#2711](https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/issues/2711)

This PR implements the first stage of showing funding/sponsorship information: Collecting the data.

I wondered whether to only collect and store the raw data so we always have all the information regardless of whether we have added support for any additional platforms GitHub will support in the future. However, since the data from the YML file does not include complete URLs, we’d also need to add custom support for any new platforms in code so it feels better to keep everything typed in the database. This also allows us to query the data more easily.

This introduces a new GitHub API call, which we should be aware of, but we have plenty of capacity right now. The only alternative would be to collect this data via the builder tool, but if we do that, we should do it for several bits of metadata, including the README and maybe the license. It feels better to keep everything consistent until we get closer to our API token limits.


Sources/App/Core/Github.swift
```diff
@@ -97,6 +99,8 @@ extension Github {
         switch resource {
             case .license, .readme:
                 return URI(string: "https://api.github.com/repos/\(trunk)/\(resource.rawValue)\(queryString)")
+            case .funding:
+                return URI(string: "https://api.github.com/repos/\(trunk)/contents/FUNDING.yml")
```
finestructure, line 103: I'm pretty sure you've checked this but just in case: this isn't available via GraphQL such we could stick it in the existing call?

Sources/App/Core/Github.swift
```diff
@@ -173,6 +177,22 @@ extension Github {
...
+        guard let uri = try? Github.apiUri(for: packageUrl, resource: .funding)
+        else { return nil }
+
+        // Fetch the raw YML contents of FUNDING.yml
+        let fetchResult = try? await Github.fetch(client: client, uri: uri, headers: [
+            ("Accept", "application/vnd.github.raw")
+        ])
+        guard let yaml = fetchResult?.content else { return nil }
+
+        do {
```
finestructure, line 190:
```suggestion
        return try? YAMLDecoder().decode(Github.Funding.self, from: yaml)
```

Sources/App/Core/Github.swift
```diff
@@ -173,6 +177,22 @@ extension Github {
...
+            return try YAMLDecoder().decode(Github.Funding.self, from: yaml)
+        } catch {
+            return nil
+        }
```
finestructure, line 194:
```suggestion
        return try? YAMLDecoder().decode(Github.Funding.self, from: yaml)
```

finestructure: Or did you mean to log the error perhaps and forgot?

Tests/AppTests/IngestorTests.swift
```diff
@@ -145,7 +147,9 @@ class IngestorTests: AppTestCase {
                                    metadata: md,
                                    licenseInfo: .init(htmlUrl: "license url"),
                                    readmeInfo: .init(etag: "etag", html: "readme html", htmlUrl: "readme html url"),
+                                   fundingInfo: nil,
                                    s3Readme: .cached(s3ObjectUrl: "url", githubEtag: "etag"))
+#warning("Add funding information here")
```
finestructure, line 152: As discussed: left over

Sources/App/Core/Github.swift
Sherlouk: This information is part of the GraphQL API 😄  Might be a reason not to use it though.

```gql
query {
  repository(owner: "SwiftPackageIndex", name:"SwiftPackageIndex-Server"){
    fundingLinks {
      ... on FundingLink {
        url
        platform
      }
    }
  }
}
```

```json
{
  "data": {
    "repository": {
      "fundingLinks": [
        {
          "url": "https://github.com/SwiftPackageIndex",
          "platform": "GITHUB"
        }
      ]
    }
  }
}
```

Sherlouk: vapor/vapor - for example

```json
{
  "data": {
    "repository": {
      "fundingLinks": [
        {
          "url": "https://github.com/vapor",
          "platform": "GITHUB"
        },
        {
          "url": "https://opencollective.com/vapor",
          "platform": "OPEN_COLLECTIVE"
        }
      ]
    }
  }
}
```

daveverwer: I looked for it in the graphql api before creating this PR and missed that it was there, which is frustrating but just one of those things. 

My original version would have supported the Vapor scenario but the new version from the gql is better, and almost complete

Sources/App/Core/Github.swift
```diff
@@ -249,6 +249,53 @@ extension Github {
...
+            case issueHunt
+            case koFi
+            case lfxCrowdfunding
+            case liberaPay
+            case openCollective
+            case otechie
+            case patreon
+            case tideLift
+
+            private static let gitHubApiEncodings: [String: Platform] = [
```
daveverwer, line 267: This (and the associated `init(from:)`) feels a little weird, but I think it's better than having two different enums that have the same cases but different raw values, and it's certainly better than having the SCREAMING_SNAKE_CASE in our database JSON.

daveverwer (commented):
OK, this is ready for another review 🫨

daveverwer:
There's a small issue with some of the URLs coming back with `https://` on them and some just starting with the domain, which is weird. I'd like to see how many and if there's a pattern to it, so I'd like this PR to go in without fixing that.

finestructure:
You mean coming back from the GH API, right? I.e. there's nothing really to be done on our end for now, except see what we get and consider perhaps automatically amending the URLs at some point?

Sources/App/Core/Github.swift
finestructure: Do I understand correctly that the API can send both `COMMUNITY_BRIDGE` _and_ `communityBridge` as possible values and that's why we to do this?

Sources/App/Models/Repository.swift
```diff
@@ -131,6 +134,7 @@ final class Repository: Model, Content {
          defaultBranch: String? = nil,
          firstCommitDate: Date? = nil,
          forks: Int = 0,
+         fundingLinks: [Github.FundingLink] = [],
```
finestructure, line 137: Admittedly a bit of a nitpick but I think we should avoid "leaking" GH entities into the db model. I'm not proposing a major change here, just to lift `FundingLink` out of the `Github` namespace into a plain model inside `App/Models` - effectively alongside for instance `Release`, which is a similar entity we're getting from GraphQL.

It'll require a little type shuffle and mapping (just like for release). I'll propose the change in a PR.

(Apart from just "feeling" nicer it'll avoid referencing Github types outside GH specific ingestion and will make it easier to add other services alongside it - if we ever end up doing that 😅)

finestructure (approved):

daveverwer:
Merging to staging to test today before moving to production.

PR #2717, opened 2023-11-15
Move funding type to top level

finestructure:
As mentioned in my comment, here's the suggestion to move the type up to the top level.

### DIFF
--- a/Sources/App/Core/Github.swift
+++ b/Sources/App/Core/Github.swift
@@ -249,53 +249,6 @@ extension Github {
         var htmlUrl: String
     }
 
-    struct FundingLink: Codable, Equatable {
-
-        enum Platform: String, Codable {
-            case communityBridge
-            case customUrl
-            case gitHub
-            case issueHunt
-            case koFi
-            case lfxCrowdfunding
-            case liberaPay
-            case openCollective
-            case otechie
-            case patreon
-            case tideLift
-
-            private static let gitHubApiEncodings: [String: Platform] = [
-                "COMMUNITY_BRIDGE": .communityBridge,
-                "CUSTOM": .customUrl,
-                "GITHUB": .gitHub,
-                "ISSUEHUNT": .issueHunt,
-                "KO_FI": .koFi,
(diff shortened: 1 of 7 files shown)


Sources/App/Core/Github.swift
```diff
@@ -417,6 +370,25 @@ extension Github {
...
+                case liberaPay = "LIBERAPAY"
+                case openCollective = "OPENCOLLECTIVE"
+                case otechie = "OTECHIE"
+                case patreon = "PATREON"
+                case tideLift = "TIDELIFT"
+            }
+
+            var platform: Platform
+            var url: String
+        }
```
finestructure, line 390: This "copy" of the type is the one that carries the GH specific parts. Note that I've simplified the decode based on the assumption (from what I've seen in the GraphQL response) that the platform strings are always uppercase.

If they're not, this would need adjusting.

Sources/App/Models/FundingLink.swift
```diff
@@ -0,0 +1,73 @@
...
+        case liberaPay
+        case openCollective
+        case otechie
+        case patreon
+        case tideLift
+    }
+
+    var platform: Platform
+    var url: String
+}
```
finestructure, line 35: This is now a very simple type that's not related to GH anymore.

daveverwer:
It's funny. I had it fully implemented this way and then backed out of it as I thought you wouldn't like it. 😵‍💫

Sources/App/Core/Github.swift
daveverwer: Unfortunately the underscores in the GH responses are really inconsistent and they need hard coding as in my original version. I'll update this.

daveverwer (approved):
Merging this, and will fix up the last bits in the main PR.

Sources/App/Core/Github.swift
finestructure: Ah, my bad, I’d copied over the camel case names and used the uppercase transformation. That nuked the underscores.

Issue #2716, opened 2023-11-15
Investigate source of `Failed to decode` log messages


finestructure:
We've been getting lots of `Failed to decode` messages in our test runs recently. I suspect it's related to a package update since we've not had any related changes on our end:

```
Test Case '-[AppTests.WebpageSnapshotTests test_HomeIndexView]' started.
Failed to decode S3Readme
Failed to decode Platform for key platform
Failed to decode FundingLink
Failed to decode Array<FundingLink>
Failed to decode Array<ProductDependency> for key dependencies
Failed to decode ProductDependency
Test Case '-[AppTests.WebpageSnapshotTests test_HomeIndexView]' passed (0.333 seconds).
```

finestructure:
I believe all of these fields are `JSONB` fields, so it might be something inside Vapor/Fluent/Postgres.

finestructure:
Not entirely new it seems 2.92.5 (Nov 8):

```
Test Case '-[AppTests.WebpageSnapshotTests test_HomeIndexView]' started.
Failed to decode S3Readme
Test Case '-[AppTests.WebpageSnapshotTests test_HomeIndexView]' passed (0.326 seconds).
```

finestructure:
Seems inconsistent, 2.92.3 has some more again:

```
Test Case '-[AppTests.WebpageSnapshotTests test_HomeIndexView]' started.
Failed to decode S3Readme
Failed to decode Array<ProductDependency> for key dependencies
Failed to decode ProductDependency
Test Case '-[AppTests.WebpageSnapshotTests test_HomeIndexView]' passed (0.328 seconds).
```

finestructure:
The real root cause seems to be a `print` that we've started to hit after an update to SwiftOpenAPI:

https://github.com/dankinsoid/SwiftOpenAPI/pull/4

finestructure:
This has been fixed upstream.

PR #71, opened 2020-05-16
Rollbar deployments

finestructure:
Merge after #67

Issue #70, opened 2020-05-15
Analyser: index.lock: File exists


finestructure:
Look how/if this gets reported and if we can clear it up:

```
2020-05-15T17:39:52.883069388Z [ ERROR ] analysis error: ShellOut encountered an error
2020-05-15T17:39:52.883122027Z Status code: 128
2020-05-15T17:39:52.883138198Z Message: "fatal: Unable to create '/checkouts/github.com-facebook-facebook-ios-sdk/.git/index.lock': File exists.
2020-05-15T17:39:52.883142468Z
2020-05-15T17:39:52.883146168Z Another git process seems to be running in this repository, e.g.
2020-05-15T17:39:52.883149748Z an editor opened by 'git commit'. Please make sure all processes
2020-05-15T17:39:52.883153048Z are terminated then try again. If it still fails, a git process
2020-05-15T17:39:52.883156258Z may have crashed in this repository earlier:
2020-05-15T17:39:52.883159378Z remove the file manually to continue."
2020-05-15T17:39:52.883162458Z Output: ""
```

finestructure:
Also looks like this isn’t marked as processed and stays at the top of the analysis candidate list:

![image](https://user-images.githubusercontent.com/65520/82112946-3cb79c80-9752-11ea-8399-6077734b3f48.png)

PR #2720, opened 2023-11-16
Issue 2716 part 1

finestructure:
This eliminates the "Failed to decode S3Readme" log message, on of three tracked in #2716 .

Because this comes with a migration (migration 71), it's based on top of the funding changes (which has migration 70 but which we've reverted for now). Therefore this _must not_ be merged until we've either merged the funding changes for good or decide to swap the migrations around (make this one here migration 70 and the one in funding 71).


daveverwer:
Go ahead and make this migration #070, I have a new local branch for `fetch-funding` and I'll make it #071 in that branch. 👍

daveverwer (approved):

PR #2726, opened 2023-11-17
Dependabot commit message

finestructure:


finestructure:
Trying to trigger the dependabot.yml validator...

finestructure:
Ok, got the validation I was looking for and the format is indeed invalid like this: https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/pull/2726/checks?check_run_id=18787777355

PR #2728, opened 2023-11-18
Remove links to our Twitter profile.

daveverwer:


finestructure (approved):
Ah nice, I forgot about those. Will probably need snapshot updates, won't it?

daveverwer:
Al

> Ah nice, I forgot about those. Will probably need snapshot updates, won't it?

Already pushed :) I always push and then remember snapshots.

PR #2732, opened 2023-11-20
Updated community sponsors

daveverwer:
Thanks so much for the support, @dokun1! It means a lot.


dokun1:
Very happy to be able to display my support 🚀

finestructure (approved):

PR #2733, opened 2023-11-21
Tweaks to local development for running the app inside a container

daveverwer:
Leaving this open until we get everything working in case more tweaks are needed.

### DIFF
--- a/LOCAL_DEVELOPMENT_SETUP.md
+++ b/LOCAL_DEVELOPMENT_SETUP.md
@@ -255,3 +255,16 @@ If you use this mechanism frequently, rather than editing `.env.development` or 
 ```
  /app/Run serve --env docker --hostname 0.0.0.0 --port 8080
 ```
+
+## Debugging database access inside the container
+
+Run the following commands in order to test database access from within the docker container:
+
+```
+apt-get update
+apt-get install postgresql-client-14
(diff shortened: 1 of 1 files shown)


LOCAL_DEVELOPMENT_SETUP.md
```diff
@@ -223,7 +223,7 @@ NB: if you are on an Apple Silicon Mac you will have to [build the base image's
 Next, launch a shell inside the container:
 
 ```
-docker run --rm -it -v $PWD:/host -w /host --network="host" --entrypoint sh registry.gitlab.com/finestructure/swiftpackageindex:be565ca16725d4836efbb2517fd0285fddbe9da0
+docker run --rm -it -v $PWD:/host -w /host --network="host" --entrypoint sh registry.gitlab.com/finestructure/spi-base:0.14.1
```
finestructure, line 226: We probably want to keep the sha because it's the reference used in the context of this example. `make docker-build` will log the sha as the image reference instead of a tag.

```suggestion
docker run --rm -it -v $PWD:/host -w /host --network="host" --entrypoint sh registry.gitlab.com/finestructure/spi-base:be565ca16725d4836efbb2517fd0285fddbe9da0
```

finestructure (approved):

PR #2734, opened 2023-11-21
Add ingest by url

finestructure:
Make it easier to ingest specific packages via url fragments, just like for analysis:

```
swift run Run ingest --url Nextbike
[ INFO ] Ingesting (url: Nextbike) ... [component: ingest]
[ WARNING ] fetchResource https://api.github.com/repos/kiliankoe/Nextbike/license request failed with status 404 Not Found [component: server]
[ WARNING ] storeS3Readme failed [component: ingest]
[ WARNING ] Environment variable not set: METRICS_PUSHGATEWAY_URL [component: ingest]
Program ended with exit code: 0
```

### DIFF
--- a/Sources/App/Commands/CommandMode.swift
+++ b/Sources/App/Commands/CommandMode.swift
@@ -1,0 +1,49 @@
+// Copyright Dave Verwer, Sven A. Schmidt, and other contributors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


Sources/App/Commands/CommandMode.swift
```diff
@@ -0,0 +1,49 @@
...
+    static let defaultLimit = 1
+
+    struct Signature: CommandSignature {
+        @Option(name: "limit", short: "l")
+        var limit: Int?
+
+        @Option(name: "id", help: "package id")
+        var id: Package.Id?
+
+        @Option(name: "url")
```
daveverwer, line 28: Should we include some help here to indicate the format of the URL? Is it the fully qualified URL with the `.git` extension?

daveverwer (approved):

Sources/App/Commands/CommandMode.swift
finestructure: It's just a URL fragment and it aborts if there are multiple matches.

PR #2738, opened 2023-11-24
Remove dependencies defaults

finestructure:
⚠️ Do not deploy before the [corresponding builder change](https://gitlab.com/finestructure/swiftpackageindex-builder/-/merge_requests/237) is live ⚠️

It is currently impossible to tell if a package has no `product_dependencies` or if product dependency discovery hasn't run yet or failed, because we initialise it to `{}` by default. This removes the default and adjusts it to the same we have for `resolved_dependencies`, which is `NULL`.

However, `resolved_dependencies` also had problem: we set a default `[]` when instantiating a `Version` object, thereby giving every version zero dependencies by default instead of leaving it undecided despite there being no actual SQL level default.

In order for this second fix to be complete, we also need to update the builder, which currently sends a default `[]` if no `Package.swift` file can be found.

That is problematic, because while the lack of a `Package.swift` file can indicate no dependencies, it can also mean that package resolution failed. We can detect this though and send the correct value in that case, with some adjustments.

⚠️ Schema change ⚠️

### DIFF
--- a/Sources/App/Migrations/071/UpdateVersionResetProductDependenciesWithDefault.swift
+++ b/Sources/App/Migrations/071/UpdateVersionResetProductDependenciesWithDefault.swift
@@ -16,7 +16,7 @@ import Fluent
 import SQLKit
 
 
-struct UpdateVersionUpdateProductDependencies: AsyncMigration {
+struct UpdateVersionResetProductDependenciesWithDefault: AsyncMigration {
     func prepare(on database: Database) async throws {
         // re-create field without default
         try await database.schema("versions")
--- a/Sources/App/Migrations/071/UpdateVersionResetResolvedDependencies.swift
+++ b/Sources/App/Migrations/071/UpdateVersionResetResolvedDependencies.swift
@@ -16,25 +16,19 @@ import Fluent
 import SQLKit
 
 
-struct UpdateVersionUpdateProductDependencies: AsyncMigration {
+struct UpdateVersionResetResolvedDependencies: AsyncMigration {
     func prepare(on database: Database) async throws {
-        // re-create field without default
+        // reset field values by dropping and re-creating the field
         try await database.schema("versions")
-            .deleteField("product_dependencies")
+            .deleteField("resolved_dependencies")
             .update()
         try await database.schema("versions")
-            .field("product_dependencies",
+            .field("resolved_dependencies",
                    .array(of: .json))
             .update()
     }
 
     func revert(on database: Database) async throws {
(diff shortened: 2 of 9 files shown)


Tests/AppTests/AnalyzerTests.swift
```diff
@@ -211,7 +211,7 @@ class AnalyzerTests: AppTestCase {
 
         // validate score
         XCTAssertEqual(pkg1.score, 35)
-        XCTAssertEqual(pkg2.score, 45)
+        XCTAssertEqual(pkg2.score, 40)
```
finestructure, line 214: This show the actual bug in our live (before this change) implementation: pkg2's version has `NULL` in `resolved_dependencies` yet we awarded 5 points because we defaulted to `[]`:

CleanShot 2023-11-24 at 10 01 53@2x

This is now fixed.

daveverwer: I was just about to type, "Why the score change?" 😂

daveverwer (approved):

Sources/App/Migrations/071/UpdateVersionResetProductDependenciesWithDefault.swift
```diff
@@ -0,0 +1,40 @@
...
+struct UpdateVersionResetProductDependenciesWithDefault: AsyncMigration {
+    func prepare(on database: Database) async throws {
+        // re-create field without default
+        try await database.schema("versions")
+            .deleteField("product_dependencies")
+            .update()
+        try await database.schema("versions")
+            .field("product_dependencies",
+                   .array(of: .json))
+            .update()
```
finestructure, line 28: Instead of altering the column and keeping the values, I'm now resetting the values, because they're tainted with the previous ambiguous processing.

Sources/App/Migrations/071/UpdateVersionResetResolvedDependencies.swift
```diff
@@ -0,0 +1,34 @@
...
+struct UpdateVersionResetResolvedDependencies: AsyncMigration {
+    func prepare(on database: Database) async throws {
+        // reset field values by dropping and re-creating the field
+        try await database.schema("versions")
+            .deleteField("resolved_dependencies")
+            .update()
+        try await database.schema("versions")
+            .field("resolved_dependencies",
+                   .array(of: .json))
+            .update()
```
finestructure, line 28: Same here, we really need to reset those values, because we can't know what `{}` in the field means.

finestructure:
Builder change is live.

Issue #1979, opened 2022-08-18
DocC plugin injection taints `versions.resolved_dependencies`


finestructure:
We're inadvertently picking up the injected DocC plugin dependency when we're reporting resolved packages:

<img width="571" alt="CleanShot 2022-08-18 at 11 02 18@2x" src="https://user-images.githubusercontent.com/65520/185355937-b3a21f53-95d9-40ec-b65f-914d2d696f65.png">

We need to either capture resolved dependencies before that happens or revert the changes to the manifest(s).

finestructure:
Fixed by https://gitlab.com/finestructure/swiftpackageindex-builder/-/merge_requests/140

finestructure:
This is proving to be a bit trickier than expected.

The first stab didn't work, because reverting the `Package.swift` changes before reporting isn't enough. We'd also have to re-run package resolution, which can be expensive. So instead I changed it to capture `Package.resolved` right after the build step.

That's when I discovered, that SPM doesn't actually create a `Package.resolved` when a package has no dependencies. A better way to determine dependencies would be to use `swift package show-dependencies --format json`. However, that's a bit of a bigger change _and_ it also triggers package resolution. So it's not something to do lightly.

`DependencyResolution` is a module which lives in the server project and is used by both the server and the builder and it currently assumes to be run as a cheap operation looking for a `Package.resolved` file. We shouldn't change that.

The only "problem" with `getResolvedDependencies(...) -> [ResolvedDependency]?` is that so far we've assumed a return value of `nil` to mean "don't know" when there's no `Package.resolved` file. This is obviously distinct from "no dependencies". The difference is really whether you run this immediately after having run package resolution. If that's the case, then "no Package.resolved" really means "no dependencies".

I think the best way to deal with this is to leave `getResolvedDependencies` as is. When we run it in the server context there's no package resolution and `nil` keeps its semantic of "don't know".

In the builder, we run `getResolvedDependencies` right after build (i.e. package resolution) and therefore we can interpret `nil` as `[]`.

finestructure:
Deployed and fixed

<img width="459" alt="CleanShot 2022-08-20 at 14 46 41@2x" src="https://user-images.githubusercontent.com/65520/185746867-f4d68280-6c50-4463-bd73-b7fbc70f3664.png">

PR #2739, opened 2023-11-24
Stop updating resolved_dependencies from analysis

finestructure:
This removes `resolved_dependency` updating in analysis, as [previously outlined](https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/issues/1979#issuecomment-1221256063).

It becomes harder to reason about which process "owns" updating a field if we do it from two places. Performing this update in analysis has the advantage of being faster, because it can happen immediately instead of relying on a build to run.

However, it is also potentially incorrect if we simply take for granted (as we did) what's in a checked-in `Package.resolved` file. It need not be up to date.

Our builder will have run `swift package resolve` and report the up-to-date value.

Merge after merging #2738

### DIFF
--- a/Tests/AppTests/AnalyzerTests.swift
+++ b/Tests/AppTests/AnalyzerTests.swift
@@ -16,7 +16,6 @@ import XCTest
 
 @testable import App
 
-import DependencyResolution
 import Fluent
 import SPIManifest
 import ShellOut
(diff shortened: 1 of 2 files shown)


Tests/AppTests/AnalyzerTests.swift
```diff
@@ -210,7 +194,7 @@ class AnalyzerTests: AppTestCase {
         XCTAssertEqual(targets.map(\.name), ["t1", "t1", "t1", "t2", "t2", "t2"])
 
         // validate score
-        XCTAssertEqual(pkg1.score, 35)
+        XCTAssertEqual(pkg1.score, 30)
```
finestructure, line 197: The score now drops, because we have `NULL` in `resolved_dependencies` where we previously had one dependency. In a real world scenario (which we're not fully testing here), this score would go back up to 35 once the build has reported back.

daveverwer (approved):

Issue #2753, opened 2023-12-01
Update Postgres version to 15


finestructure:
Azure now offers 15 as the most recent supported version and we should upgrade to it.

https://learn.microsoft.com/en-us/azure/postgresql/flexible-server/concepts-major-version-upgrade

finestructure:
We've been on 16.4 for a while now.

PR #2754, opened 2023-12-01
Update postgres version, remove superfluous NSUnbufferedIO=YES

finestructure:


.github/workflows/ci.yml
```diff
@@ -28,7 +28,7 @@ jobs:
       image: registry.gitlab.com/finestructure/spi-base:0.14.1
     services:
       postgres:
-        image: postgres:13.8-alpine
+        image: postgres:13.11-alpine
```
daveverwer, line 31: I was confused at first, but if I'm with you this is just a minor version bump in 13 rather than upgrading to 15 as mentioned in #2753?

daveverwer (approved):

PR #2755, opened 2023-12-01
Bump swift-custom-dump to 1.0+, umpinning swift-parser et al

finestructure:


daveverwer (approved):

finestructure:
I wonder if this is the same underlying problem (and fix) as here: https://github.com/dankinsoid/SwiftOpenAPI/issues/5#issuecomment-1836135210

PR #2756, opened 2023-12-01
Applying version 4.4.4 and updated example

heckj:
With the combination of https://github.com/dankinsoid/VaporToOpenAPI/issues/17, and some follow on work, he released a version 4.4.4 of VaporToOpenAPI that correctly renders the openapi.json (or at least it fully validates). I grabbed the openapi.json that was generated with this PR in my local dev instance, and stashed it at https://github.com/dankinsoid/VaporToOpenAPI.git - and then in turn used that to generate source (clients and types) using the Swift OpenAPI Generator project.

### DIFF
--- a/Package.resolved
+++ b/Package.resolved
@@ -437,8 +437,8 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/dankinsoid/SwiftOpenAPI.git",
       "state" : {
-        "revision" : "2ee6c2ad3da09601a1979bd51e6f76c0517369f5",
-        "version" : "2.18.4"
+        "revision" : "aec41de7bbea8176e68ec5d4d270da453e9c0d11",
+        "version" : "2.19.3"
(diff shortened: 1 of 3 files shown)


heckj:
library reproduction case reproduced (a transitive dependency to VaporToOpenAPI - PR with a failing test case at https://github.com/dankinsoid/SwiftOpenAPI/pull/6).

finestructure:
This seems to be a common problem: https://github.com/pointfreeco/swift-parsing/discussions/327#discussioncomment-7731905

finestructure:
It's mostly a Linux thing, I think, but since that's where we deploy... 😅

finestructure:
I'll push this into the SPI repo and rebase, because with Swift 5.9.2 we should be able to unpin the dependency.

finestructure:
Closing in favour of #2789

Issue #2562, opened 2023-08-18
Investigate why apple/swift-nio-ssl-support is not being added by the nightly script


finestructure:
It's being referenced as a dependency by 77 packages yet our nightly doesn't discover it as a missing package.

Example package: https://swiftpackageindex.com/brokenhandsio/SteamPress

finestructure:
Ok, this is interesting. None of the 77 packages actually has `swift-nio-ssl-support` among its direct dependencies. That's why our validator isn't picking it up. What's happening is that apparently all of these packages are referencing an older Vapor version (3.x) which _did_ have it as a dependency.

Take for instance `SteamPress`. I has

```swift
    	.package(url: "https://github.com/vapor/vapor.git", from: "3.0.0"),
```

in its manifest and `show-dependencies` shows the graph as follows:

![CleanShot 2023-08-20 at 14 29 22@2x](https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/assets/65520/cf3b4688-d12b-4869-8c89-e746f233a2bb)

Because we use package resolution when we populate `versions.resolved_dependencies` we get the "true" graph for the versions as specified. `validator` doesn't do package resolution - it just takes each direct dependency url and crawls the tree - but not historically. So since later Vapor versions don't reference `swift-nio-ssl-support` it's not being picked up.

finestructure:
I think we can both fix this discrepancy _and_ make `validator` much more efficient. Rather than crawling 6k packages and probe each of their direct dependencies, we can just compile the full list of dependencies to test server side (like I did manually here) and simply probe the ones we don't have in the index yet, also server side. No GH requests required at all. We just need an API endpoint we can query from the validator to get the missing dependencies.

finestructure:
This is now running in our nightly and as we progress through the backlog of new packages `apple/swift-nio-ssl-support` will be added. From a local run finding all packages that will eventually be added:

```
      ...

  ✅ ADD 31 https://github.com/aidantwoods/swift-sodium.git
  ✅ ADD 32 https://github.com/alibaba/HandyJSON.git
  ✅ ADD 33 https://github.com/apple/swift-cmark.git
  ✅ ADD 34 https://github.com/apple/swift-lmdb.git
  ✅ ADD 35 https://github.com/apple/swift-nio-nghttp2-support.git
  ✅ ADD 36 https://github.com/apple/swift-nio-ssl-support.git
[...]

PR #2776, opened 2023-12-11
Dependencies endpoint

finestructure:
Add `/api/dependencies` endpoint. This is to be used as part of improving missing dependency discovery in our nightly job (#2562) - and in general to further dependency analysis by giving us an easy way to extract the full dependency tree across all packages.


finestructure:
Ad hoc deploying this to proceed with testing

daveverwer (approved):

PR #2787, opened 2023-12-15
2nd attempt at: Merge pull request #2781 from SwiftPackageIndex/dependabot/npm_and_ya…

finestructure:
…rn/dependencies-7337d43281

NPM: Bump the dependencies group with 6 updates

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,17 +21,6 @@ on:
     - cron: "0 8 * * *"
 
 jobs:
-  test-build-front-end:
-    name: Build front-end
-    runs-on: ubuntu-latest
-    container:
-      image: registry.gitlab.com/finestructure/spi-base:0.14.2
-    steps:
-      - name: Checkout code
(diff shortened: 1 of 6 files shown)


daveverwer (approved):

finestructure:
Seems to be working now: https://gitlab.com/finestructure/swiftpackageindex/-/jobs/5759980412

.github/workflows/ci.yml
```diff
@@ -21,6 +21,17 @@ on:
     - cron: "0 8 * * *"
 
 jobs:
+  test-build-front-end:
```
daveverwer, line 24: Wouldn't failures here be caught in the lint CI step? We can make it more explicit, but I think we are already covered by that step.

finestructure: Ah right, that's already over in the other file :+1:

daveverwer: For clarity, we could move the build to here and leave the lint there. It's rather badly named at the moment.

PR #2820, opened 2024-01-08
Add social post snapshot test

finestructure:


Sources/App/Core/Mastodon.swift
```diff
@@ -25,7 +25,8 @@ enum Mastodon {
         var accessToken: String
     }
 
-    static func post(client: Client, message: String) async throws {
+    // NG: _testEncodedURL is a callback that exists purely to be able to regression test the encoded value
+    static func post(client: Client, message: String, _testEncodedURL: (String) -> Void = { _ in }) async throws {
```
finestructure, line 29: Unfortunately we're leaking this test callback into the production interface but I can't come up with a better way to capture and validate the encoded url before posting. (Brining up a whole fake endpoint to post to feels a bit much.)

We also can't `#ifdef DEBUG` out the parameter itself. We'd have to duplicate the whole function but then we're maintaining two implementations, which is bad.

Tests/AppTests/SocialTests.swift
```diff
@@ -232,4 +233,26 @@ class SocialTests: AppTestCase {
...
+            summary: nil,
+            maxLength: Social.postMaxLength
+        )
+
+        // MUT
+        try? await Mastodon.post(client: app.client, message: message) { encoded in
+            assertInlineSnapshot(of: encoded, as: .lines) {
+                """
+                https://mas.to/api/v1/statuses?status=%E2%AC%86%EF%B8%8F%20owner%20just%20released%20packageName%20v2.6.4%0A%0Ahttp://localhost:8080/owner/SuperAwesomePackage%23releases
+                """
```
finestructure, line 253: I've checked that this test actually fails now when we update Vapor. Will push that in a new branch.

daveverwer (approved):

PR #2835, opened 2024-01-15
Tweak label

finestructure:
Tested locally:

CleanShot 2024-01-15 at 17 58 24@2x

### DIFF
--- a/Sources/App/Views/PackageController/PackageShow+View.swift
+++ b/Sources/App/Views/PackageController/PackageShow+View.swift
@@ -192,8 +192,7 @@ extension PackageShow {
                         listPhrase(nodes: model.fundingLinks.map { fundingLink in
                                 .a(
                                     .href(fundingLink.url),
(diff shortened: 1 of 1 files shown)


Sources/App/Views/PackageController/PackageShow+View.swift
```diff
@@ -356,20 +355,12 @@ extension PackageShow {
...
-}
-
-extension FundingLink.Platform {
-    var name: String {
-        switch self {
+    // TODO: Test this
+    var label: String {
+        switch platform {
             case .communityBridge: return "LFX Mentorship"
+            case .customUrl: return URL(string: url)?.host ?? url
```
finestructure, line 363: I believe this is the best fallback we can construct if all else fails.

daveverwer (approved):
We need to fix the URLs at the source, as it's not just the display of them that needs fixing, but the URL for the link, too. However, this is also a good change for display and is simpler than my solution.

Thanks!

PR #2836, opened 2024-01-15
Fix the common issue with funding URLs sent back via the GitHub API

daveverwer:
This one *is* ready for review.

Once merged, we will need to let this run a full ingestion cycle in production to make sure that the database is fully updated before the front-end code assumes it will find valid URLs.

### DIFF
--- a/Sources/App/Models/FundingLink.swift
+++ b/Sources/App/Models/FundingLink.swift
@@ -36,9 +36,20 @@ struct FundingLink: Codable, Equatable {
 
 
 extension FundingLink {
-    init(from node: Github.Metadata.FundingLinkNode) {
+    init?(from node: Github.Metadata.FundingLinkNode) {
         platform = .init(from: node.platform)
-        url = node.url
+
+        // Some URLs come back from the GitHub API without a scheme. In every circumstance I have
+        // observed, this is a simple lack of any scheme. We should try a simple fix of prepending
+        // `https`, but if it's still won't parse as a URL, we should discard that funding link entirely.
+        switch node.url {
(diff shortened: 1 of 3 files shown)


Sources/App/Models/FundingLink.swift
```diff
@@ -36,9 +36,20 @@ struct FundingLink: Codable, Equatable {
...
         platform = .init(from: node.platform)
-        url = node.url
+
+        // Some URLs come back from the GitHub API without a scheme. In every circumstance I have
+        // observed, this is a simple lack of any scheme. We should try a simple fix of prepending
+        // `https`, but if it's still won't parse as a URL, we should discard that funding link entirely.
+        switch node.url {
+            case let url where URL(string: url)?.scheme != nil:
+                self.url = url
+            case let url where URL(string: "https://\(url)") != nil:
```
daveverwer, line 48: This may be overkill. I thought URLs could fail initialisation, but in testing with a Playground they seem to be *unbelievably* forgiving with input data.

I'm not sure if we should leave this in. It doesn't do any harm that I can see. What do you think, @finestructure?

daveverwer: Exhibit A 😬

Screenshot 2024-01-15 at 17 14 11@2x

finestructure (approved):

Sources/App/Models/FundingLink.swift
finestructure: Yes, `URL(string:)` only really fails on an empty string 🤷‍♂️

PR #2845, opened 2024-01-17
Add back the package funding feature

### LINKED ISSUE #2711: Show funding/sponsorship information on Packge pages

daveverwer:
Fixes #2711

Reverts #2842 which reverted #2834

### DIFF
--- a/Sources/App/Views/PackageController/PackageShow+View.swift
+++ b/Sources/App/Views/PackageController/PackageShow+View.swift
@@ -15,6 +15,7 @@
 import Ink
 import Vapor
 import Plot
+import TLDExtract
 
 
 extension PackageShow {
@@ -136,6 +137,7 @@ extension PackageShow {
                 .class("details two-column"),
                 .section(
                     mainColumnMetadata(),
+                    packageSponsorship(),
                     .hr(
                         .class("minor")
                     ),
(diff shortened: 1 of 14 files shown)


finestructure (approved):

Sources/App/Views/PackageController/PackageShow+View.swift
```diff
@@ -329,3 +354,21 @@ extension PackageShow {
         }
     }
 }
+
+extension FundingLink {
+    var label: String {
+        switch platform {
+            case .communityBridge: return "LFX Mentorship"
+            case .customUrl: return URL(string: url)?.host?.removingPrefix("www.").removingPrefix("blog.") ?? url
```
daveverwer, line 362: There are a couple of cases in the existing data where this doesn't get the subdomain and while this CTA doesn't look as good with them left in, I also don't want to start getting into adding special cases for "jaywcjlove" and "spezi.sites".

finestructure (approved):

PR #2846, opened 2024-01-17
Pre process funding links

daveverwer:
Separating from #2845 as this needs to process for a full ingestion cycle before the display code goes live.

### DIFF
--- a/Sources/App/Models/FundingLink.swift
+++ b/Sources/App/Models/FundingLink.swift
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 import Foundation
+import TLDExtract
 
 
 struct FundingLink: Codable, Equatable {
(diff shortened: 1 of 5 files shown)


Sources/App/Models/FundingLink.swift
```diff
@@ -32,6 +33,7 @@ struct FundingLink: Codable, Equatable {
 
     var platform: Platform
     var url: String
+    var urlLabel: String
```
finestructure, line 36: Nit: I think we should call this `rootDomain` and make it

```swift
var rootDomain: String?
```

and then to the `?? URL(string: url)?.host ?? url` on the package page view side. That way we don't store view related data in the db.

finestructure: Or we could, and this may be better, push the label into the `customUrl` case. Let me experiment with that.

PR #2856, opened 2024-01-22
Copy backtrace binary to image

### LINKED ISSUE #2855: Investigate swift-backtrace hangs
Run

```
lldb -p PID
bt all
```

when it happens. Perhaps force it to happen on staging.

finestructure:
Fixes #2855 

- [x] ad hoc deploy and test crashing behaviour

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -43,6 +43,9 @@ WORKDIR /staging
 # Copy main executable to staging area
 RUN cp "$(swift build --package-path /build -c release --show-bin-path)/Run" ./
 
+# Copy static swift backtracer binary to staging area
+RUN cp "/usr/libexec/swift/linux/swift-backtrace-static" ./
+
 # Copy resources bundled by SPM to staging area
 RUN find -L "$(swift build --package-path /build -c release --show-bin-path)/" -regex '.*\.resources$' -exec cp -Ra {} ./ \;
 


daveverwer (approved):

finestructure:
BTs are in the logs now but hard to spot, because they bypass our service filters (their logs don't go through standard logging, which makes sense, but that means they don't show up in our standard service logs).

They're a bit hard to spot, because they don't have a format that lends itself to filtered out and alerted on. I've raise an issue to ask for support for a common prefix (suffix) to make that easier: https://github.com/apple/swift/issues/71057

In the meantime, we can look for the `Backtrace took` line at the end or perhaps `Backtracing from` for the start (the former seems more reliable, I'm not sure if the latter is universal).

finestructure:
The biggest issue is that backtraces take 45s or more, so we'll want to limit them to dev. I tried various options but none affect runtime in any noticeable way.

Dockerfile
```diff
@@ -43,6 +43,9 @@ WORKDIR /staging
 # Copy main executable to staging area
 RUN cp "$(swift build --package-path /build -c release --show-bin-path)/Run" ./
 
+# Copy static swift backtracer binary to staging area
+RUN cp "/usr/libexec/swift/linux/swift-backtrace-static" ./
```
MahdiBM, line 47: @finestructure I don't think this is enough. Down below you should also need https://github.com/vapor/template/pull/120

finestructure: We're not setting this via the Dockerfile so we can update it without having to rebuild. Our full `SWIFT_BACKTRACE` is

`enable=yes,images=all,interactive=no,swift-backtrace=./swift-backtrace-static,threads=all,output-to=stderr`

MahdiBM: You are settings this via the Dockerfile though, down below. Maybe that's unintended?

MahdiBM: to be clear, by "this" I mean "SWIFT_BACKTRACE".

finestructure: Oh, I'd missed that! In my testing this is being overridden by our dynamically set variable but I'll definitely want to remove it to avoid confusing which one is being used. Good catch, thanks!

Issue #2868, opened 2024-01-26
`.spi.yml` exceeds size limit


finestructure:
### Discussed in https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/discussions/2867

<div type='discussions-op-text'>

<sup>Originally posted by **Wei18** January 26, 2024</sup>
**Please describe the bug**

I modified [.spi.yml@af70860](https://github.com/Wei18/github-rest-api-swift-openapi/blob/main/.spi.yml) and verified on [validate-spi-manifest](https://swiftpackageindex.com/validate-spi-manifest), wait for days but not working.

The only first time added [.spi.yml@a97557c](https://github.com/Wei18/github-rest-api-swift-openapi/blob/a97557c42628b2621faf3f8e8c3fa8e1fcaa2b93/.spi.yml) worked.
(doc: https://swiftpackageindex.com/Wei18/github-rest-api-swift-openapi/main/documentation/githubrestapiissues)

Thanks

</div>

finestructure:
- [ ] raise size limit slightly
- [ ] make sure our validator fails with a proper error message if the file is too large

finestructure:
The job [times out](https://swiftpackageindex.com/Wei18/github-rest-api-swift-openapi/builds), with the build alone taking 6 minutes. It's just too many targets to try and generate docs for afterwards. We just manage two of them before it times out, and that's before uploading the docs.

@Wei18, I'm afraid there's nothing we can do here. You'll have to get the build time down to a couple of minutes to even have a chance generating docs for that many targets, I believe.

![CleanShot 2024-01-26 at 17 28 09@2x](https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/assets/65520/c712413c-ea04-480e-828c-3801ff069624)

finestructure:
OK, I've re-run the build job and this time it went faster. The build only took 2m20s and we manage to generate docs for around 25 targets before it times out (at target `GitHubRestAPIPackages`).

It might work with 10 doc targets but I'd actually try and keep it below 5 or so to make the results more reliable.

Wei18:
Okay, I will try other methods to reduce the number of targets.

Thanks.

PR #2870, opened 2024-01-26
Issue 2868

finestructure:
Server part of #2868

CleanShot 2024-01-26 at 11 08 17@2x

### DIFF
--- a/Tests/AppTests/__Snapshots__/WebpageSnapshotTests/test_ValidateSPIManifest_show.1.html
+++ b/Tests/AppTests/__Snapshots__/WebpageSnapshotTests/test_ValidateSPIManifest_show.1.html
@@ -77,7 +77,7 @@ are indexed. Validate your manifests on this page."/>
               <p>Enter the contents of a 
                 <code>.spi.yml</code> file for validation:
               </p>
-              <textarea name="manifest" autofocus rows="15" maxlength="1000">version: 1
+              <textarea name="manifest" autofocus rows="15" maxlength="1600">version: 1
 builder:
   configs:
     - documentation_targets: [Target1, Target2]</textarea>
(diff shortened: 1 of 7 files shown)


Tests/AppTests/__Snapshots__/WebpageSnapshotTests/test_ValidateSPIManifest_show.1.html
```diff
@@ -77,7 +77,7 @@ <h2 class="trimmed">Validate a Swift Package Index manifest</h2>
               <p>Enter the contents of a 
                 <code>.spi.yml</code> file for validation:
               </p>
-              <textarea name="manifest" autofocus rows="15" maxlength="1000">version: 1
+              <textarea name="manifest" autofocus rows="15" maxlength="1600">version: 1
```
daveverwer, line 80: 1600 or 1500? Is this so there's a chance for the error to show?

We should have the error message server-side in case anything breaks or is being bypassed on the client-side, but I think if we only allow 1,500 bytes, we should limit this to 1,500 characters.

daveverwer (approved):
Just the one note about the max size in the text field.

Tests/AppTests/__Snapshots__/WebpageSnapshotTests/test_ValidateSPIManifest_show.1.html
daveverwer: Can we also drive this from the constant?

finestructure: That's intentional so we can actually detect an input that's too large. `maxlength` clips the input string, doesn't it? If we clip it here, we won't give `Manifest.load` a chance to detect that it's over size.

PR #2893, opened 2024-02-12
Dynamic build timeouts

finestructure:
This allows us to configure a dynamic build timeout, which is a modification of our existing flat build timeout of 10m based on whether a build needs to generate documentation or not.

I went back and forth where to handle the dynamic build timeout and ended up pushing it up into the server. It is easier to do here. In the builder it would require a whole new subcommand to determine doc generation from the SPI manifest file, because the timeout is required in `build.sh`.

This change is additive and can be deployed without any related builder changes. It sends a new `TIMEOUT` parameter which older builders will simply ignore.


Sources/App/Commands/TriggerBuilds.swift
```diff
@@ -448,16 +452,27 @@ func missingPairs(existing: [BuildPair]) -> Set<BuildPair> {
...
+        .field(Version.self, \.$id)
+        .field(Version.self, \.$packageName)
+        .field(Version.self, \.$reference)
+        .field(Version.self, \.$spiManifest)
+        .all()
+    let builds = try await Build.query(on: database)
+        .filter(\.$version.$id ~~ versions.compactMap(\.id))
+        .field(Build.self, \.$platform)
+        .field(Build.self, \.$swiftVersion)
+        .field(Build.self, \.$version.$id)
```
finestructure, line 466: This is a slightly unrelated change: we were pulling _every_ field in `Version` and `Build` to create a trigger when we only needed a handful. This trims the list considerably.

finestructure: Note that we were running two queries before as well. `.with(\.$builds)` runs an additional query just like we do here and fetches all the siblings of the parent relation. This query style doesn't support `.field` modifiers so we have to run the queries and merge the results manually.

daveverwer:
Is there a way on the build machine to determine the timeout for a specific build?

One task I had for today was to implement the "killing" of builds that go over the timeout in cases where things go wrong. I could implement something over a "maximum timeout" or something, but is there any way for me to figure out that timeout without being that unsubtle?

finestructure:
That's the `TIMEOUT` variable we're sending or what else would you be looking for?

daveverwer (approved):

PR #2906, opened 2024-02-20
Newer 5.10 toolchain

finestructure:


finestructure:
No faster than the existing one, not worth updating for now.

PR #2922, opened 2024-02-29
Remove Yarn as our dependency manager for JavaScript

daveverwer:
We have been stuck on a very old version of Yarn for too long and the upgrade path for it is not clear at all. We should move to plain old `npm`, which has improved significantly since the start of this project when we picked Yarn.

I also did a few more cleanup tasks on that side of the project while I was doing it.


daveverwer:
I'm going to give this a test on Staging before moving it to production

daveverwer:
All looks good. Deploying as 2.99.5

Issue #2937, opened 2024-03-12
Caching of README files is potentially not working


daveverwer:
Needs investigation, but something isn't right.

We have code that appears to check the ETAG before fetching/storing it a README file to our cache, but at least locally on my machine, it’s not working:

```
[ INFO ] Ingesting (limit: 10) ... [component: ingest]
[ INFO ] Candidate count: 10 [component: ingest]
[ INFO ] Copying readme to s3://spi-dev-readmes/sdggiesbrecht/swift-driver/readme.html ... [component: server]
[ INFO ] Copying readme to s3://spi-dev-readmes/shareup/json-apple/readme.html ... [component: server]
[ INFO ] Copying readme to s3://spi-dev-readmes/ctreffs/swiftsimctl/readme.html ... [component: server]
[ INFO ] Copying readme to s3://spi-dev-readmes/swhitty/swiftdraw/readme.html ... [component: server]
[ INFO ] Copying readme to s3://spi-dev-readmes/serverdriven/screendataui-ios/readme.html ... [component: server]
[ INFO ] Copying readme to s3://spi-dev-readmes/adamayoung/police-data-kit/readme.html ... [component: server]
[ INFO ] Copying readme to s3://spi-dev-readmes/mcrich23/mcrich23-toolkit/readme.html ... [component: server]
[ INFO ] Copying readme to s3://spi-dev-readmes/ptsochantaris/can-proceed/readme.html ... [component: server]
[ INFO ] Copying readme to s3://spi-dev-readmes/purelayout/purelayout/readme.html ... [component: server]
```

Both with `--limit` and `--id`, it fetches and caches **all** README files, every time.

I deployed a quick bit of logging to staging to see if it was happening there, and it's much less clear that there's a bug:

```
2024-03-12 11:13:40 | [ INFO ] Copying readme to s3://spi-dev-readmes/william-weng/wwtoast/readme.html ... [component: server]
2024-03-12 11:13:41 | [ INFO ] Copying readme to s3://spi-dev-readmes/goldenplan/tvoskeyboard/readme.html ... [component: server]
2024-03-12 11:13:41 | [ INFO ] Copying readme to s3://spi-dev-readmes/naftaly/threadcrumb/readme.html ... [component: server]
2024-03-12 11:13:41 | [ INFO ] Copying readme to s3://spi-dev-readmes/mtj0928/slidekit/readme.html ... [component: server]
[...]

PR #2939, opened 2024-03-13
Cache protected images in GitHub readme files

### LINKED ISSUE #2790: Broken README image
We do some wrangling of GitHub image URLs to make them work in SPI READMEs but there's obviously something going on here that's breaking this one:

Screenshot 2023-12-15 at 12 59 37@2x

and on GitHub:

Screenshot 2023-12-15 at 13 00 39@2x

In the GitHub page, the URL is:

```
https://private-user-images.githubusercontent.com/60271542/286382420-cf47949d-e99b-42f8-bd74-b308252edf59.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTEiLCJleHAiOjE3MDI2NDU1ODYsIm5iZiI6MTcwMjY0NTI4NiwicGF0aCI6Ii82MDI3MTU0Mi8yODYzODI0MjAtY2Y0Nzk0OWQtZTk5Yi00MmY4LWJkNzQtYjMwODI1MmVkZjU5LmdpZj9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFJV05KWUFYNENTVkVINTNBJTJGMjAyMzEyMTUlMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjMxMjE1VDEzMDEyNlomWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPWQ0ZTk2N2UyYjE0ZjY0OTNjNTlkNjgxZWI0ODQzZjY5MTk2ZmY2NDczNTYxNDVkZGQyZjVjMDQwZWYxMTA4MTMmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0JmFjdG9yX2lkPTAma2V5X2lkPTAmcmVwb19pZD0wIn0.CQn8CPtaCbrZtPYS1KpNCgnX506Xkv4V5J1hfqOWCb8
```

and in our page, the URL is:

```
https://private-user-images.githubusercontent.com/60271542/286382420-cf47949d-e99b-42f8-bd74-b308252edf59.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTEiLCJleHAiOjE3MDI2NDQxMDcsIm5iZiI6MTcwMjY0MzgwNywicGF0aCI6Ii82MDI3MTU0Mi8yODYzODI0MjAtY2Y0Nzk0OWQtZTk5Yi00MmY4LWJkNzQtYjMwODI1MmVkZjU5LmdpZj9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFJV05KWUFYNENTVkVINTNBJTJGMjAyMzEyMTUlMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjMxMjE1VDEyMzY0N1omWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPWQ0NzQ5YTZkMGM0YmJiOTE2NTJlM2YxOTljZDY0YjVjN2Q0MDE3NTdmYTcxMWRlNjRjNmYwMTM2YjQzYTYzZDYmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0JmFjdG9yX2lkPTAma2V5X2lkPTAmcmVwb19pZD0wIn0.2O283Gh1A5PqRJYi65hoPOuoywkibbC3pBxTA_TN8m0
```

The `jwt` is wrong in the second one. I thought that we may have just been out of sync with our README cache, but the repository hasn't been been touched in 5 days. Odd.

daveverwer:
Fixes #2790 

This PR scans any readme files fetched from the GitHub API that also have images that are hosted on `private-user-images.githubusercontent.com` and also have a `jwt`parameter.

The images are collected and uploaded to S3 in the same directory as the readme file is stored. They retain their filename in that directory.

Images are only uploaded if/when the `etag` of the readme itself changes, which is the behaviour it already had (also see #2937, although this should not make that worse if it even is a problem!)


Sources/App/Migrations/076/UpdateRepositoryResetReadmes.swift
```diff
@@ -0,0 +1,33 @@
...
+
+struct UpdateRepositoryResetReadmes: AsyncMigration {
+    func prepare(on database: Database) async throws {
+        guard let db = database as? SQLDatabase else {
+            fatalError("Database must be an SQLDatabase ('as? SQLDatabase' must succeed)")
+        }
+
+        try await db.raw("""
+        UPDATE repositories
+        SET s3_readme = jsonb_set(s3_readme, '{cached,githubEtag}', '""', FALSE);
```
finestructure, line 27: This is to reset the field, isn't it? Why aren't we just setting it to `NULL`? That way we don't "bake in" any format and can be sure it's readable.

Sources/S3Store/S3Store.swift
```diff
@@ -27,6 +27,12 @@ public struct S3Store {
     }
     
     public func save(payload: String, to key: Key) async throws {
+        guard let data = payload.data(using: .utf8)
+        else { throw Error.genericError("Could not convert readme string to data") }
+        try await save(payload: data, to: key)
```
finestructure, line 32:
```suggestion
        try await save(payload: Data(payload.utf8), to: key)
```

Tests/AppTests/GithubTests.swift
```diff
@@ -15,10 +15,12 @@
 @testable import App
 
 import Vapor
+import SwiftSoup
+import S3Store
 import XCTest
 
 
-class GithubTests: AppTestCase {
+class GithubTests: SnapshotTestCase {
```
finestructure, line 23:
```suggestion
class GithubTests: AppTestCase {
```

Class doesn't seem to have gained any snapshot tests?

Tests/AppTests/__Snapshots__/GithubTests/test_fetchReadme_withPrivateGitHubImages.1.txt
```diff
@@ -0,0 +1,6 @@
+<html>
+ <head></head>
+ <body> 
+  <img src="https://awsReadmeBucket.s3.us-east-2.amazonaws.com/foo/bar/with-jwt.jpg" data-original-src="https://private-user-images.githubusercontent.com/with-jwt.jpg?jwt=some-jwt" /> 
+ </body>
+</html>
```
finestructure, line 6: It looks like this was a snapshot for a test in `GithubTests` in an earlier version but isn't anymore? There's no `GithubTests.test_fetchReadme_withPrivateGitHubImages`.

finestructure (changes requested):

Sources/App/Migrations/076/UpdateRepositoryResetReadmes.swift
daveverwer: It's to reset just the `githubEtag` part of the JSON field. If we reset the field fully, we'll lose all README files for the duration of a full pass of ingestion as the URL to the S3 cached copy of the README is in the other part of this JSON field. 

I'd welcome verification of the SQL statement, though. I have tried it and it does what I expect it to. The `FALSE` at the end means that if there is no value for etag then it won't create it, which is also for safety around fields that do not yet have a value (for new packages yet to be ingested).

It seemed like the correct way to approach it to me.

finestructure: Oh, I see! That makes sense.

In that case we should put that in a test though. We have a couple that do similar things, for example `test_pending_to_triggered_migration` and `test_DeleteArmBuilds_migration`.

Sources/App/Core/Extensions/S3Store+ext.swift
```diff
@@ -41,15 +41,41 @@ extension S3Store {
...
+        if let imageUrl {
+            guard let url = URL(string: imageUrl)
+            else { throw S3Store.Error.genericError("Invalid imageUrl \(imageUrl)") }
+            let filename = url.lastPathComponent
+            let path = "\(owner)/\(repository)/\(filename)".lowercased()
+            return .init(bucket: bucket, path: path)
+        } else {
+            let path = "\(owner)/\(repository)/readme.html".lowercased()
+            return .init(bucket: bucket, path: path)
+        }
```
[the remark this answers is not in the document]
daveverwer: I don't know if I agree here. I like that the API is the same on the consumer side for both parts of this since they are so closely related. I'd certainly not call it weird

Tests/AppTests/__Snapshots__/GithubTests/test_fetchReadme_withPrivateGitHubImages.1.txt
daveverwer: Well spotted. Thanks

Sources/App/Migrations/076/UpdateRepositoryResetReadmes.swift
daveverwer: Added `test_migration076_updateRepositoryResetReadmes`

finestructure (approved):

PR #2940, opened 2024-03-14
Receive buildErrors in build reports

finestructure:
Receive build errors in build reports. Comparing builder PR: https://gitlab.com/finestructure/swiftpackageindex-builder/-/merge_requests/268

I've put the migration out of the way so it doesn't conflict with the #2790 PR (but these would need updating before we merge this PR here)

### DIFF
--- a/Sources/App/Migrations/080/UpdateBuildAddBuildErrors.swift
+++ b/Sources/App/Migrations/080/UpdateBuildAddBuildErrors.swift
@@ -1,0 +1,31 @@
+// Copyright Dave Verwer, Sven A. Schmidt, and other contributors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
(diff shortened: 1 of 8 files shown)


Sources/App/Migrations/080/UpdateBuildAddBuildErrors.swift
```diff
@@ -0,0 +1,31 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import Fluent
+
+
+#warning("move to proper folder before merging")
```
finestructure, line 18: This is in a folder `080` only so it doesn't conflict with the migration in branch `fix-2970`. Will move to proper sequence before merging, depending on which one ends up going first.

Sources/App/configure.swift
```diff
@@ -334,6 +334,11 @@ public func configure(_ app: Application) throws -> String {
     app.commands.use(TriggerBuildsCommand(), as: "trigger-builds")
     app.commands.use(ReAnalyzeVersions.Command(), as: "re-analyze-versions")
 
+#warning("move to proper place before merging")
+    do {  // Migration 0xx - Add `build_errors` to `builds`
+        app.migrations.add(UpdateBuildAddBuildErrors())
+    }
+
```
finestructure, line 341: This is out of order only so it doesn't conflict with the migration in branch `fix-2970`. Will move to proper sequence before merging, depending on which one ends up going first.

daveverwer (approved):

Sources/App/configure.swift
finestructure: Updated

PR #2948, opened 2024-03-18
Removed deprecated method introduced with #2939

daveverwer:


Tests/AppTests/GithubTests.swift
```diff
@@ -347,11 +338,10 @@ class GithubTests: AppTestCase {
...
         Current.githubToken = { "secr3t" }
-        let pkg = Package(url: "https://github.com/foo/bar")
         let client = MockClient { _, resp in resp.status = .notFound }
 
         // MUT
-        let res = await Github.fetchLicense(client: client, packageUrl: pkg.url)
+        let res = await Github.fetchLicense(client: client, owner: "foo", repository: "bar")
 
         // validate
         XCTAssertEqual(res, nil)
```
finestructure, line 347: We have a warning on line 456 (see below):

> Initialization of immutable value 'images' was never used; consider replacing with assignment to '_' or removing it

It's a bit weird, it doesn't show up consistently in Xcode for me for some reason, so it's easy to miss.

One way of fixing it would be to just check on `images`:

```swift
XCTAssertEqual(images, [])
```

daveverwer: Odd that it had not shown up in Xcode for me at all!

Fixed

PR #2955, opened 2024-03-18
Ready for Swift 6 scaffolding

daveverwer:
I thought I'd do this piece by piece rather than in one huge PR. This adds the page, a basic snapshot, and an announcement on the home page (which we may want to use a different design for, but this gives us an easy link while testing).

Everything gated to dev/staging right now, so it's safe to merge.

### DIFF
--- a/Sources/App/Views/PublicPage.swift
+++ b/Sources/App/Views/PublicPage.swift
@@ -314,7 +314,7 @@ class PublicPage {
 
         return .p(
             .class("announcement"),
-            .text("How many packages are compatible with Swift 6? Find out Check out which packages are  "),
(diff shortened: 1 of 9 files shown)


Sources/App/Views/PublicPage.swift
```diff
@@ -309,12 +309,18 @@ class PublicPage {
...
-//        .p(
-//            .class("announcement"),
-//            .text("No current announcement.")
-//        )
+        guard Current.environment() == .development
+        else { return .empty }
+
+        return .p(
+            .class("announcement"),
+            .text("How many packages are compatible with Swift 6? Find out Check out which packages are  "),
```
finestructure, line 317: `Find out Check out` :)

finestructure (approved):

Issue #2797, opened 2023-12-19
Move all documentation hosting to stable URLs


daveverwer:
We currently host documentation on a variety of URLs:

* `[owner]/[repo]/main/documentation/package`
* `[owner]/[repo]/0.1.0/documentation/package`
* `[owner]/[repo]/0.1.0-pre1/documentation/package`

and every time there is a new release, instead of redirecting to `/0.1.0/documentation/package` we now start redirecting to `/0.2.0/documentation/package`. We always set the canonical URLs to the new version and update sitemaps to point at the new version, but this is causing a *huge* amount of churn in what we are asking Google to index and is contributing to our ongoing search index issues.

We need to host documentation like this:

### `[owner]/[repo]/documentation/package`

This is the canonical URL for a package's documentation, and is _not_ a redirect. This should have a canonical URL meta tag and header pointing to itself and be marked for indexing.

### `[owner]/[repo]/[reference]/documentation/package`

This is a canonical path to a reference specific documentation set, but should not be marked as canonical and should be excluded from Google indexing with a `noindex` tag and header. These pages should all point 

## Notes:

* We should host the most recent documentation on the canonical URL *and* the reference specific URL.

## Steps

- [ ] Host the canonical documentation for a package on the `[owner]/[repo]/documentation/package` URL
- [ ] Set the canonical URL _all_ documentation pages to be the `[owner]/[repo]/documentation/package` URL
- [ ] Set `noindex` via a meta tag and HTTP header on every documentation page apart from the canonical page
- [ ] Update the sitemap to only ever include the canonical documentation URLs, never reference specific URLs
- [ ] Change the link on the package page to be the package's canonical documentation URL
- [ ] Add a permalink icon to the documentation header that copies a link to a reference specific URL for the page's content

daveverwer:
Progress is in the `stable-url-doc-hosting` branch.

finestructure:
As discussed, I've had a look into how many packages opt-in to generate docs but don't have any releases and the number is quite small: 14.

However, there's a different set of packages that is significantly larger that would currently be affected if we only ask Google to index packages with release docs, and that's packages that generate docs but have not had a release since they opted-into doc generation. They _also_ don't have release docs. There are 95 of those, which is ~15% of all packages requesting docs (14 of 629).

[...]

PR #2961, opened 2024-03-22
Rewrite doc base paths

finestructure:
This introduces support for `{owner}/{repo}/~/documenation` style doc urls that are base-path rewritten from our existing doc sets. It does not attempt to make any further changes to our URLs with respect to the remaining steps laid out in #2797. It's a pre-requisite to make these changes.

This is a rebased version of the (now named) `no-direct-old` branch, which gates all the changes to dev only.

I think I know how to refactor and clean this up but it's going to be a large change and I don't think we want to roll this into this same PR.

We should thoroughly review this change here as is, while it closely resembles the existing routes and test it on staging. Once we're happy we should roll it out completely (i.e. more the dev-gating).

Only once _that's_ done we can work in the remaining changes for stable hosting and then we can look into refactoring with a `DocRoute` type (as described in the `FIXME` comment).


finestructure:
I've run some additional manual tests against the ad deployed branch. Process recorded in the tracking issue #2797 so it doesn't get closed with the PR: https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/issues/2797#issuecomment-2022375073

daveverwer (approved):

Issue #2969, opened 2024-03-26
"Ready for Swift 6" front-end work


daveverwer:
Issue to keep track of the remaining tasks for the "Ready for Swift 6" page:

- [x] Dark mode support
- [x] Mobile support (static images rather than SVGs)
- [x] Fix snapshot tests so we snapshot a chart, if it's not too much work

daveverwer:
This is all done!

PR #2970, opened 2024-03-26
Add placeholder charts to the Ready for Swift 6 page

daveverwer:
Progress towards #2969

The primary goal of merging this now is to have a basis to discuss wording, charts, links, etc...  One important part of this is that I have separated the chart specification from the data files so we are less likely to break the chart as we add new stats every time we build.

There are some notes on the two charts below, and then a screenshot:

### First chart: Compatible packages over time

* X-axis intentionally starts from approximate date of our first set of tests and goes through almost to the end of the year so we can see that there is more to come, especially while there are only a few data points.
* Y-axis will grow automatically over time as the maximum number increases, we could also set this to be fixed as we will know our total compatibility testing set ahead of time.
* Hovering over a point in the line shows the exact date and number of packages for that data point.

### Second chart: Total errors over time

* I'm not sure of the value of this chart, but it was one we said we'd do so here it is!
* No hovering here, the values are baked into the bars

–

Screenshot 2024-03-26 at 19 36 32@2x


finestructure:
Looks great!

I think the second chart is important to show overall progress. It could very well be that the "compatible package count" doesn't change much between versions because only few packages managed to eliminate _all_ errors. This would mean the first graph would remain rather flat. However, the second graph would still show that progress nicely.

Regarding that second graph's style: Is there a reason this is a bar chart with x/y flipped WRT graph 1? The x values are the same and this could just be the same x/y graph as the first one with the error count on the y axis. I think that would make more sense and also show the progression in tandem.

finestructure:
(They could even be the same graph with left/right y axis but I'm not sure if that'd be better overall.)

Sources/App/Views/ReadyForSwift6/ReadyForSwift6Show+View.swift
```diff
@@ -34,14 +35,88 @@ extension ReadyForSwift6Show {
...
+    }
+}
+
+private extension Data {
+    func compactJson() -> String? {
+        guard let json = try? JSONSerialization.jsonObject(with: self),
+              let compactedJsonData = try? JSONSerialization.data(withJSONObject: json),
+              let compactJson = String(data: compactedJsonData, encoding: .utf8)
+        else { return nil }
+        return compactJson
```
finestructure, line 120: What's this doing exactly? Looks to me like it's taking JSON loaded from disk and essentially passing it through to a String? Couldn't we just pass the String from the file?

finestructure (approved):

PR #2992, opened 2024-04-12
Documentation routing refactor

### LINKED ISSUE #2983: Excessive querying when using ~ documentation URLs
We have a problem with our URL rewriting in terms of resource usage. I noticed this while reworking the doc routing into a more unified mechanism.

We rewrite to the following HTML in order to support ~ URLs:

```html
...
    <link rel="icon" href="/swiftpackageindex/semanticversion/~/favicon.ico"/>
    <link rel="mask-icon" href="/swiftpackageindex/semanticversion/~/favicon.svg" color="#333333"/>
    <title>SemanticVersion Documentation – Swift Package Index</title>
    <script>
    var baseUrl = "/swiftpackageindex/semanticversion/~/"
    </script>
    <link href="/swiftpackageindex/semanticversion/~/css/chunk-c0335d80.10a2f091.css" rel="prefetch"/>
    <link href="/swiftpackageindex/semanticversion/~/css/documentation-topic.1d1eec04.css" rel="prefetch"/>
    <link href="/swiftpackageindex/semanticversion/~/css/documentation-topic~topic.b6287bcf.css" rel="prefetch"/>
    <link href="/swiftpackageindex/semanticversion/~/css/documentation-topic~topic~tutorials-overview.d6f5411c.css" rel="prefetch"/>
...
```

The problem with this is that all these URLs are "under qualified". We need to make DB lookups in order to find out which S3 resource to load (here, the reference is unspecified). In case of our SemanticVersion docs for example, this leads to a total of 19 (!) db lookups - one for each resource referenced in the HTML file.

I don't think we want to be leaning on this in this form. We need to explore if we can at least rewrite the static resources to fully qualified paths that don't require lookups.

Using

```html
    <link rel="icon" href="/swiftpackageindex/semanticversion/0.4.0/favicon.ico"/>
```

probably won't work when the `baseUrl` is set to `/swiftpackageindex/semanticversion/~/` (to be tested) but

```html
    <link rel="icon" href="https://swiftpackageindex.com/swiftpackageindex/semanticversion/0.4.0/favicon.ico"/>
```

should.

finestructure:
This should be it, finally.

To do:

- [x] verify if this fixes #2983


finestructure:
This does not yet fix #2983 on its own

daveverwer (approved):

PR #2999, opened 2024-04-16
Issue 2983 excessive doc queries

### LINKED ISSUE #2983: Excessive querying when using ~ documentation URLs

finestructure:
Fixes #2983 

Merge after #2992

### DIFF
--- a/Sources/App/Core/CurrentReferenceCache.swift
+++ b/Sources/App/Core/CurrentReferenceCache.swift
@@ -5,6 +5,8 @@ typealias CurrentReferenceCache = ExpiringCache<String, String>
 
 
 extension CurrentReferenceCache {
+    static let live = CurrentReferenceCache(duration: .minutes(5))
+
     subscript(owner owner: String, repository repository: String) -> String? {
         get {
             let key = "\(owner)/\(repository)".lowercased()
--- a/Sources/App/Core/CurrentReferenceCache.swift
+++ b/Sources/App/Core/CurrentReferenceCache.swift
@@ -1,0 +1,18 @@
+import Cache
+
+
+typealias CurrentReferenceCache = ExpiringCache<String, String>
+
+
+extension CurrentReferenceCache {
+    subscript(owner owner: String, repository repository: String) -> String? {
+        get {
+            let key = "\(owner)/\(repository)".lowercased()
+            return self[key]
+        }
+        set {
+            let key = "\(owner)/\(repository)".lowercased()
+            self[key] = newValue
+        }
+    }
+}
--- a/Sources/App/routes+documentation.swift
+++ b/Sources/App/routes+documentation.swift
@@ -159,25 +159,17 @@ private extension Request {
         if fragment.requiresArchive && archive == nil { throw Abort(.badRequest) }
         let pathElements = parameters.pathElements(for: fragment, archive: archive)
 
+#warning("add tests to check when we're hitting the db and when we don't")
         let docVersion = try await { () -> DocRoute.DocVersion in
             if reference == String.current {
-                if fragment.requiresArchive {
-                    guard let params = try await DocumentationTarget.query(on: db, owner: owner, repository: repository)?.internal
-                    else { throw Abort(.notFound) }
-                    if archive?.lowercased() != params.archive.lowercased() {
-                        throw Abort(.notFound)
-                    }
-                    Current.currentReferenceCache()?[owner: owner, repository: repository] = params.reference
-                    return .current(referencing: params.reference)
-                } else {
-                    if let ref = Current.currentReferenceCache()?[owner: owner, repository: repository] {
-                        return .current(referencing: ref)
-                    } else {
-                        guard let params = try await DocumentationTarget.query(on: db, owner: owner, repository: repository)?.internal
-                        else { throw Abort(.notFound) }
-                        return .current(referencing: params.reference)
-                    }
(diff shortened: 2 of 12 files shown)


Sources/App/Core/CurrentReferenceCache.swift
```diff
@@ -0,0 +1,20 @@
+import Cache
+
+
+typealias CurrentReferenceCache = ExpiringCache<String, String>
+
+
+extension CurrentReferenceCache {
+    static let live = CurrentReferenceCache(duration: .minutes(5))
```
finestructure, line 8: I don't love this but I don't see another way for us to deal with this than introduce a short-lived in-memory cache of owner/repo → reference mappings.

Sources/App/routes+documentation.swift
```diff
@@ -161,11 +161,13 @@ private extension Request {
...
-                else { throw Abort(.notFound) }
-                if fragment.requiresArchive && archive?.lowercased() != params.archive.lowercased() {
+                if let ref = Current.currentReferenceCache()?[owner: owner, repository: repository] {
+                    return .current(referencing: ref)
+                }
+                guard let params = try await DocumentationTarget.query(on: db, owner: owner, repository: repository)?.internal else {
                     throw Abort(.notFound)
                 }
+                Current.currentReferenceCache()?[owner: owner, repository: repository] = params.reference
                 return .current(referencing: params.reference)
```
finestructure, line 171: Once we've looked up the current reference for a given owner/repo we keep it around in the cache for 5 minutes. That way we don't need to fiddle with any further rewrites (we can keep the resources in the HTML as ~ refs - might as well, since the JS makes calls to ~ resources anyway) and can just rely on this cache.

Sources/App/Core/CurrentReferenceCache.swift
daveverwer: I haven't checked into how these additional queries work yet, but are they only on the first JavaScript page load?

11 extra queries on one documentation page that then facilitates any further browsing without requests might be preferable to introducing this cache? If it's on every JavaScript navigation event, I agree this will be necessary, but if it's just the first load, we should weigh this trade off up.

Sources/App/routes+documentation.swift
finestructure: It's 19 db requests on initial documentation page load. I believe subsequent loads do not hit the db but even so we really don't want all those unnecessary requests hitting the db.

Sources/App/Core/CurrentReferenceCache.swift
finestructure: It's 19 db requests on initial documentation page load. I believe subsequent loads do not hit the db but even so we really don't want all those unnecessary requests hitting the db.

daveverwer: But doesn't this way mean that there will be a 5 minute period where documentation will be broken after analysis runs and finds a new release? That's a big tradeoff.

daveverwer: Ah, I remember now, we already have something that serves up old documentation until the builds finish, so this would be no worse than that.

finestructure: I believe the build delay would take care of that, plus at worst we'd be mixing css/js that's probably the same between revisions. We could also purge the cache for the specific repo when we discover the new version.

finestructure: (re old docs vs just resource - we wouldn't _have_ to use the cache for the docs but I thought we might as well)

daveverwer (approved):

PR #3005, opened 2024-04-19
Cleanup of Package.swift formatting

gwynne:
Makes `Package.swift`'s code formatting consistent. Uses `.target()` and `.package()` consistently throughout for specifying target dependencies. Factors out the `SwiftConcurrency` setting to avoid repeating the string for every target.


daveverwer:
Thanks @gwynne!

PR #3006, opened 2024-04-19
Convert project to the modern Vapor entrypoint

gwynne:
Swaps out the old issue-prone `Dispatch`-based async entrypoint in favor of the newer async entrypoint now available upstream, matching the logic used by the latest version of the Vapor project template. Removes the separate `Run` target (redundant since Swift 5.7), making sure to keep the name of the final executable the same.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -22,6 +22,7 @@ let package = Package(
         .macOS(.v13)
     ],
     products: [
+        .executable(name: "Run", targets: ["App"]),
         .library(name: "Authentication", targets: ["Authentication"]),
         .library(name: "S3Store", targets: ["S3Store"]),
     ],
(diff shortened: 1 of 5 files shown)


daveverwer (approved):

finestructure:
For some reason Xcode isn't happy with that target change. I can build via the command line, `swift build`, but Xcode refused to build this. There are lots of `Cannot find type 'App' in scope` errors:

CleanShot 2024-05-08 at 12 10 15@2x

I've tried the usual dance of clean, nuke DD. I'm guessing this is related to the selected scheme but even autocreating schemes doesn't bring up an App scheme or something. Here's what I have:

CleanShot 2024-05-08 at 12 16 03@2x

I'm using the Run scheme day to day. I've set up lots of arguments and env vars, so ideally I'd be able to re-use or migrate that if it needs changing.

finestructure:
Making this draft until we figure out how to get this to compile in Xcode. May need to revert the target change or something?

finestructure:
Problem persists with Xcode 15.4 RC for me unfortunately.

finestructure (approved):

PR #3039, opened 2024-05-04
Fix crash

### LINKED ISSUE #3021: Investigate `DocRoute` related crashes
### How to test
- deploy to prod (currently no other way to reproduce)
- run [doc-test](https://gitlab.com/finestructure/swiftpackageindex/-/pipeline_schedules)
  - typically has at least one 521 failure, perhaps run a couple of times if not
  - three runs without errors nor crashes is strong evidence the problem is fixed
- observe server restart rate

- 2.102.4 is OK
- 2.102.5 causes crashes, seems to be triggered by running doc-test (or just regular doc browsing)
- doesn't look like the caching is the problem, crashes and doc-test failures persist when cache is disabled
- Cannot reproduce locally when running server via Xcode and running doc-test rester command against the local server (with db snapshot from April 24)
- 2.102.9 is OK (3 doc-tests without error, no crashes)
  - this release has the changes of 2.102.5 rolled back

CleanShot 2024-04-28 at 17 38 17 2@2x

### Other things to try
- run the actual linux image
- try to figure out why we're not seeing stack traces

finestructure:
Fixes #3021 

To do:

- [x] test on Linux
- [x] test more routes (non-documentation fragment routes)
- [x] try it with a Swift 6 preview toolchain to see if that fixes the original problem, too
  - crashes the compiler (on `main`, on `doc-route-2`, and on `fix-doc-route-crash`), only tried toolchain `org.swift.600202404221a`

### DIFF
--- a/Sources/App/routes+documentation.swift
+++ b/Sources/App/routes+documentation.swift
@@ -108,6 +108,10 @@ private extension Parameters {
                 // See https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/issues/2168
                 // and https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/issues/2172
                 // for details.
+                // ⚠️ DO NOT CHANGE THE LINE AFTER THIS COMMENT WITHOUT REVIEWING ISSUE
+                // https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/issues/3021
+                // AND THE FIX
+                // https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/pull/3039
                 return [archive].compacted().map { $0.lowercased() } + catchall.map { $0.lowercased() }
             case .css, .faviconIco, .faviconSvg, .images, .img, .index, .js, .linkablePaths, .themeSettings:
                 return catchall
--- a/Sources/App/routes+documentation.swift
+++ b/Sources/App/routes+documentation.swift
@@ -108,7 +108,7 @@ private extension Parameters {
                 // See https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/issues/2168
                 // and https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/issues/2172
                 // for details.
-                return ([archive].compacted() + catchall).map { $0.lowercased() }
+                return [archive].compacted().map { $0.lowercased() } + catchall.map { $0.lowercased() }
             case .css, .faviconIco, .faviconSvg, .images, .img, .index, .js, .linkablePaths, .themeSettings:
                 return catchall
         }


Sources/App/routes+documentation.swift
```diff
@@ -108,7 +108,7 @@ private extension Parameters {
                 // See https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/issues/2168
                 // and https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/issues/2172
                 // for details.
-                return ([archive].compacted() + catchall).map { $0.lowercased() }
+                return [archive].compacted().map { $0.lowercased() } + catchall.map { $0.lowercased() }
```
finestructure, line 111: I have no idea why this fixes the crash in release mode but it does (on macOS at least, Linux TBD).

This looks like a compiler bug to me.

There's something weird about calling `getCatchall()` on `Parameters` and using the result. It seems like unless the resulting `[String]` is being modified (i.e. copied) somehow this will crash in the `LockStorage.deinit` when it goes out of scope.

finestructure:
I've tested all 134 doc links in `doc-test` manually by opening them in Safari with the server running in release mode in an arm64 Linux container and they all load fine without issues. This tests all different fragment paths.

When doing the same on the `doc-route-2` branch (i.e. without the fix in this PR), this crashes almost immediately, as previously observed.

So this weird change _does_ fix our issue, although I'm worried how obscure this fix is and why it even works.

I'll add comment as to why this change is necessary, because it's easy to be "cleaned up" again in the future.

Sources/App/routes+documentation.swift
gwynne: Offhand I'd guess it's something to do with the use of the `+` operator, which seems to give the compiler conniptions even at the best of times. `compacted()` doesn't return an `Array`, whereas `map()` does, so by splitting the map you've changed the type solver's problem from `CompactedSequence<String> + Array<String>` to `Array<String> + Array<String>`. I'm willing to bet `([archive].compactMap { $0 } + catchall).map { $0.lowercased }` would also have worked. But then again, I would have probably just written `[archive, catchall].compactMap { $0?.lowercased }` in the first place 😰.

finestructure: Interesting! I always thought of `.compacted() == .compactMap { $0 }`. Didn't even occur to me to look at its actual type. What's weird though is that it compiles and only crashes in release mode, isn't it? Seems like a compiler bug if that's the underlying problem, doesn't it?

I'll poke around a bit to see if I can isolate that out of the project.

Your replacement `[archive, catchall].compactMap { $0?.lowercased }` doesn't work unfortunately, because `archive` is `String?` and `catchall` is `[String]`.

weissi: I'd put money on this not being a compiler bug. Almost certainly there's a concurrency bug and this is not a fix, it is merely hiding the underlying (presumed) concurrency bug. This is quite expected because now you're allocating a new, fresh Array faster than before. So if something mutates (only possible because of a concurrency bug) the old array at the same time as you're reading it, you'll crash. And now you're using it for a shorter amount of time so you shortened the race condition and therefore reduced the probability of reproduction.

Just to be clear: If it were me, I would revert this and find the underlying bug. It will still crash now but you'll have a harder time to reproduce which is a bad thing.

weissi: I'd have a really good look at:

- is `Parameters` maybe a mutable `class`? If yes:
  - is `catchall` coming from a `var` on `Paramters`? If yes, is it appropriately protected by a lock/actor in all cases?
  - is `archive` coming from a `var` on `Parameter`? If yes, it is appropriately protected by a lock/actor in all cases?
- if `Parameters` is a `struct`, what `class` is holding onto it and is the access to the parameters appropriately protected by a lock/actor in all cases?

gwynne: https://github.com/vapor/routing-kit/blob/main/Sources/RoutingKit/Parameters.swift#L91-L97

It's a struct and the method is mutating with no locking involved.

gwynne: However, Sven is acting on a local copy of it at the time of calling the method, so I don't think the mutation is the issue.

finestructure: Thanks to @gwynne alerting me to the fact that I'm accidentally using the transient dependency swift-algorithms here which defines `compacted`. I thought this was a standard library function. I'll replace that entirely now with `compactMap`.

gwynne: (N.B.: There is no reason that the copy's original would be getting mutated at the time of this logic either, as far as I can tell.)

finestructure: Yeah, `Parameters` and its `Catchall` was the first thing I looked at to make sure I'm not messing with shared state.

I agree with @weissi that it feels like I'm papering over something but I just don't know where else to look. I have

- no TSan, Asan diagnostics
- inconclusive backtraces
- the PR doesn't introduce any new concurrency constructs

on the flip side I have

- a very clear "direct cause" for the crash
- a stable reproducer - i.e. it doesn't sometimes work/fail like concurrency bugs often do

I simply don't know what else to do about this and I really need to wrap this change up. I'm all out of ideas and need to get this merged ASAP. It's quite a blocker in terms of other work 😞

weissi: Or it could be a closure capturing a `var something: Parameters`. Like this would be illegal

```
var parameters = Parameters(...)
DispatchQueue.global().async {
    print(parameters)
}
pararameters.somethingMutating()
```

finestructure: The same problem actually persists with a newer version of RoutingKit where this function isn't `mutating` anymore. I'll make sure to update the test project to be on that version!

finestructure: Ok, I've retested this in our Server and it _does_ indeed crash with RoutingKit 4.9.1

https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/blob/8a0849b5d78cfd807e6c4f130f16c9799f37ee86/Sources/App/routes%2Bdocumentation.swift#L40

```
~/P/S/spi-server on release-crash took 4s
❯ git rev-parse @
8a0849b5d78cfd807e6c4f130f16c9799f37ee86
~/P/S/spi-server on release-crash
❯ swift run -c release Run serve --hostname 0.0.0.0 --port 8080 --env development
Building for production...
[1/1] Write swift-version-6044DDE57671499D.txt
Build complete! (0.45s)
[ NOTICE ] Server starting on http://0.0.0.0:8080 [component: server]
[ INFO ] GET /crash/1234567890 [component: server, request-id: E613BE6E-BF16-4B0A-886E-4F5237D126A3]
fish: Job 1, 'swift run -c release Run serve …' terminated by signal SIGSEGV (Address boundary error)
```

finestructure: So my reproducer is imperfect in the sense that it didn't crash in all the same circumstances as the full project but hopefully it's still illustrative of the overall problem.

weissi: > So my reproducer is imperfect in the sense that it doesn't crash in all the same circumstances as the full project but hopefully it's still illustrative of the overall problem.

This is super great! I think this should be enough

PR #3047, opened 2024-05-08
Give `DocumentationTarget` references some type safety

daveverwer:
Just a little refactor. I see you did something similar in `DocRoute` @finestructure .

Ideally, we could get rid of the constant string and keep ie encapsulated. This does not do that, but it moves us a little closer.

### DIFF
--- a/Sources/App/Core/DocumentationTarget.swift
+++ b/Sources/App/Core/DocumentationTarget.swift
@@ -20,8 +20,8 @@ enum DocumentationTarget: Equatable, Codable {
     case `internal`(reference: Reference, archive: String)
 
     enum Reference: Equatable, Codable {
-        case canonical
-        case named(ref: String)
+        case current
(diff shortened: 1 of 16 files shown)


Sources/App/Core/DocumentationTarget.swift
```diff
@@ -17,7 +17,12 @@ import Fluent
 
 enum DocumentationTarget: Equatable, Codable {
     case external(url: String)
-    case `internal`(reference: String, archive: String)
+    case `internal`(reference: Reference, archive: String)
+
+    enum Reference: Equatable, Codable {
+        case current
+        case reference(String)
+    }
```
finestructure, line 25: This is the exact same type as `DocRoute.DocVersion`. I'll have a look if they can be merged, it feels weird to have this declared twice.

finestructure (approved):

PR #3049, opened 2024-05-09
Re use doc version

finestructure:
This replaces `DocumentationTarget.Reference` with `DocVersion`.

### DIFF
--- a/Sources/App/routes+documentation.swift
+++ b/Sources/App/routes+documentation.swift
@@ -160,7 +160,7 @@ extension Request {
         if fragment.requiresArchive && archive == nil { throw Abort(.badRequest) }
         let pathElements = parameters.pathElements(for: fragment, archive: archive)
 
-        let docVersion = try await { () -> DocRoute.DocVersion in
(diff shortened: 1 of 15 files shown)


Sources/App/routes+documentation.swift
```diff
@@ -169,8 +169,8 @@ extension Request {
                 guard let params = try await DocumentationTarget.query(on: db, owner: owner, repository: repository)?.internal
                 else { throw Abort(.notFound) }
 
-                Current.currentReferenceCache()?[owner: owner, repository: repository] = "\(params.reference)"
-                return .current(referencing: "\(params.reference)")
+                Current.currentReferenceCache()?[owner: owner, repository: repository] = "\(params.docVersion)"
+                return .current(referencing: "\(params.docVersion)")
```
finestructure, line 173: We could now also cache `DocVersion`s instead of `String`s but I didn't want to pack that in here.

finestructure: Abandoning this, not worth the trouble.

daveverwer (approved):

PR #3051, opened 2024-05-09
Update chart data

finestructure:


heckj:
is this the start of "real" data from the runs you've been making?

daveverwer (approved):

PR #3078, opened 2024-05-20
Switchover to "current" documentation paths

### LINKED ISSUE #2797: Move all documentation hosting to stable URLs

daveverwer:
Fixes #2797

⚠️ Not ready to merge ⚠️

I want several PRs to go live at the same time so I'll consolidate into thif branch with further PRs.

### DIFF
--- a/Sources/App/Controllers/PackageController+PackageResult.swift
+++ b/Sources/App/Controllers/PackageController+PackageResult.swift
@@ -90,4 +90,16 @@ extension PackageController.PackageResult {
           preReleaseVersion?.model
         ].canonicalDocumentationTarget()
     }
+
+    func currentDocumentationTarget() -> DocumentationTarget? {
+        guard let target = canonicalDocumentationTarget()
+        else { return nil }
+
+        switch target {
+            case .external:
+                return target
+            case .internal(_, let archive):
+                return .internal(docVersion: .current(referencing: nil), archive: archive)
+        }
+    }
 }
--- a/Sources/App/Controllers/API/API+PackageController+GetRoute+Model.swift
+++ b/Sources/App/Controllers/API/API+PackageController+GetRoute+Model.swift
@@ -45,7 +45,7 @@ extension API.PackageController.GetRoute {
         var isArchived: Bool
         var hasBinaryTargets: Bool
         var homepageUrl: String?
-        var documentationTarget: DocumentationTarget? = nil
+        var currentDocumentationTarget: DocumentationTarget? = nil
         var weightedKeywords: [WeightedKeyword]
         var releaseReferences: [App.Version.Kind: App.Reference]
         var fundingLinks: [FundingLink]
@@ -75,7 +75,7 @@ extension API.PackageController.GetRoute {
                       isArchived: Bool,
                       hasBinaryTargets: Bool = false,
                       homepageUrl: String? = nil,
-                      documentationTarget: DocumentationTarget? = nil,
+                      currentDocumentationTarget: DocumentationTarget? = nil,
                       weightedKeywords: [WeightedKeyword] = [],
                       defaultBranchReference: App.Reference,
                       releaseReference: App.Reference?,
@@ -109,7 +109,7 @@ extension API.PackageController.GetRoute {
             self.isArchived = isArchived
             self.hasBinaryTargets = hasBinaryTargets
             self.homepageUrl = homepageUrl
-            self.documentationTarget = documentationTarget
+            self.currentDocumentationTarget = currentDocumentationTarget
             self.weightedKeywords = weightedKeywords
             self.releaseReferences = {
                 var refs = [App.Version.Kind.defaultBranch: defaultBranchReference]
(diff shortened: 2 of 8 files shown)


daveverwer:
@finestructure even though this PR is not complete. What is here is ready for review and other changes will come from other PRs merged into this.

Sources/App/Controllers/PackageController+PackageResult.swift
```diff
@@ -90,4 +90,16 @@ extension PackageController.PackageResult {
...
+
+    func currentDocumentationTarget() -> DocumentationTarget? {
+        guard let target = canonicalDocumentationTarget()
+        else { return nil }
+
+        switch target {
+            case .external:
+                return target
+            case .internal(_, let archive):
+                return .internal(docVersion: .current(referencing: nil), archive: archive)
```
finestructure, line 102: Or:

```suggestion
            case let .internal(reference, archive):
                return .internal(docVersion: .current(referencing: reference), archive: archive)
```

It'll need some thinking through which one of these is correct and if it matters.

daveverwer: I'm not sure I'm with you. The purpose of this method is to ensure that it always returns a "current" reference rather than anything else. If you want the real reference, `canonicalDocumentationTarget` defined just above is the method to use.

daveverwer: The question this raises is maybe whether `canonicalDocumentationTarget` should be renamed. It's going to be easy to misuse that when you want `currentDocumentationTarget`.

finestructure: Yes, it will still return a current reference. What needs thinking through is whether we should / need to use the underlying reference (which we have!) and pass it along. It's potentially an important parameter used in rewriting a doc archive to a current reference.

For example, if we load a doc archive that's for ref `1.2.3` we can only turn it into a `~` referenced archive if we know that it's a `.1.2.3` ref archive originally.

daveverwer: But if we pass the reference in, then when we bring the reference back out (via DocVersion line 24) it won't be a "current" reference, will it? Unless I'm missing something and that `reference` property isn't used how I think it's used.

finestructure: That's when you get its underlying `reference`, which is what I think we need to consider preserving.

In terms of the actual URL, that's driven by `pathEncoded`:

```swift
    var pathEncoded: String {
        switch self {
            case .current: String.current
            case .reference( let reference): reference.pathEncoded
        }
    }
```

finestructure: (or `description`)

```swift
extension DocVersion: CustomStringConvertible {
    var description: String {
        switch self {
            case .current:
                return String.current
            case .reference(let string):
                return string
        }
    }
}
```

daveverwer: I see. Thanks.

I don't think we want to keep the information around as it could potentially cause confusion (as it did 😬) and I don't think we'll ever need to reference the original reference.

finestructure: I can confirm that setting `nil` here is fine (as is the other variant where was pass on the reference). I think I have a way to make this less confusing but I'm still trying to figure out if this is a net gain. I'll do this separately and won't mix this into this PR.

finestructure (approved):

daveverwer:
To be re-opened at a later time

PR #3083, opened 2024-05-24
Show s6 readiness

finestructure:
I've made it so we always show the node with the S6 readiness status, not just when it's "ready" so people can always inspect the details via the tooltip.

Before taking this live on prod we should probably link to a details page rather than stick it in the tooltip but this is a quick way to get the info across.

CleanShot 2024-05-24 at 10 13 16@2x

CleanShot 2024-05-24 at 10 13 45@2x

CleanShot 2024-05-24 at 10 06 16@2x

### DIFF
--- a/Sources/App/Controllers/API/API+PackageController+BuildInfo.swift
+++ b/Sources/App/Controllers/API/API+PackageController+BuildInfo.swift
@@ -21,9 +21,11 @@ extension API.PackageController {
         typealias NamedBuildResults = GetRoute.Model.NamedBuildResults
         typealias PlatformResults = GetRoute.Model.PlatformResults
         typealias SwiftVersionResults = GetRoute.Model.SwiftVersionResults
+        typealias Swift6Readiness = GetRoute.Model.Swift6Readiness
 
         var platform: ModelBuildInfo<PlatformResults>?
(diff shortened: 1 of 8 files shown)


daveverwer (commented):
This looks great!

I’ll make a few front-end tweaks early next week. I think we can get away with the tooltip with a couple of tweaks for this first iteration.

One thing I'm not sure about is whether we should make this data visible for all cases with this first version. I completely agree we'll need all cases eventually, and we may even end up experimenting with colour in the warning if/when this becomes expected. However, it feels like a rather onerous requirement to apply to all packages on day one.

I know this breaks all the rules about data that only gets shown in certain circumstances, but since we’re so early on this feature it might take some people by surprise and result in some pushback from package authors.

finestructure:
The full display is for dev only so early adopters can inspect their package data until we have a better place to link to the details.

finestructure:
If we could merge this in the meantime that'd be helpful to the SSWG folks to inspect their packages.

daveverwer (approved):

daveverwer:
> If we could merge this in the meantime that'd be helpful to the SSWG folks to inspect their packages.

Ah I hadn't looked at the code yet so didn't realise it was still staging only.

PR #3095, opened 2024-05-27
Two small tweaks into the "Current" documentation PR before some bigger changes

daveverwer:
One better comment and a small refactor.


daveverwer:
Hmm, I thought it would show a diff just between the two branches but it seems to be showing a full diff to main.

daveverwer:
Something is horribly wrong with the git timeline here, fixing.

finestructure (approved):

PR #3096, opened 2024-05-27
Added a "more information" link to the data race safety metadata item

daveverwer:
Not final wording, and we will need the URL of the blog post or documentation we want to link to before this goes live, but what about this as a starting point?

Screenshot 2024-05-27 at 16 21 51@2x

Which links to:

Screenshot 2024-05-27 at 16 23 21@2x

### DIFF
--- a/Resources/Markdown/docs/builds.md
+++ b/Resources/Markdown/docs/builds.md
@@ -92,6 +92,6 @@ You can [find more details in this issue](https://github.com/SwiftPackageIndex/S
 
 The Swift 6.0 compiler can test whether code is safe from data races at compile time when strict concurrency checks are enabled. The data race safety information we publish on package pages comes from metadata output by the compiler during our build process.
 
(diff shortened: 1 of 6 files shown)


Resources/Markdown/docs/builds.md
```diff
@@ -86,3 +87,11 @@ In cases like this, Xcode cannot resolve package dependencies because of how it
 ```
 
 You can [find more details in this issue](https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/issues/1532).
+
+<h3 id="data-race-safety">What is data race safety and how is it tested?</h3>
+
+The Swift 6.0 compiler can test whether code is safe from data races at compile time when strict concurrency checks are enabled. The data race safety information we publish on package pages comes from metadata output by the compiler during our build process.
```
ktoso, line 93:
```suggestion
The Swift 6.0 compiler can check whether code is safe from data races at compile time when strict concurrency checks are enabled. The data race safety information we publish on package pages comes from metadata output by the compiler during our build process.
```

probably best to stick to "check" in terms of compiler and leave "test" wording to tests.

ktoso (approved):
Thanks, this is very good -- it'll help clarify and educate about the feature and what to do about it if your package isn't clear of errors :)

PR #3141, opened 2024-06-17
Enable data race safety label on package pages

finestructure:
We want this to go live with the blog post, don't we?


daveverwer:
We do, but that's several days away.

We can enable this as soon as the run is complete, though. 👍

daveverwer (approved):

finestructure:
Several days? It should be done in the morning!

daveverwer:
> Several days? It should be done in the morning!

I meant the blog post. That will take several days of review etc…

PR #3144, opened 2024-06-18
Add totals to the Ready for Swift 6 charts

daveverwer:
Thanks to @mattmassicotte for the feedback that led to this!

Adds new "total" marks that show the total number of packages when hovered.

Screenshot 2024-06-18 at 09 01 14@2x

They are associated with the various plots, so when you remove a plot the total also gets hidden.

Screenshot 2024-06-18 at 09 03 57@2x

Unchecking "Show totals" hides all totals.

Screenshot 2024-06-18 at 09 04 48@2x

This only affects the first chart, the total number of errors needs no totals.


finestructure (approved):

finestructure:
I was baffled for a moment why the total is 3396 and not ~7599 until I realised that we're constraining the list to recently updated packages. That's probably worth a note somewhere 🤔

daveverwer:
> I was baffled for a moment why the total is 3396 and not ~7500 until I realised that we're constraining the list to recently updated packages. That's probably worth a note somewhere 🤔

We have this:
Screenshot 2024-06-18 at 09 56 54@2x

mattmassicotte:
Great stuff, thank you so much!

mattmassicotte:
I have a potentially annoying question: why not just have a graph that goes down?

"number of packages with zero data race errors"
"number of packges with data race errors"

I'd also possibly change this to:

"number of packges with Swift 6 errors"

But that's maybe being too pendant 😉

daveverwer:
Has cheesy stock photography taught you nothing @mattmassicotte? Charts are best when they go up and to the right!

image

mattmassicotte:
My understanding is this is only true in the nothern hemisphere...

PR #3152, opened 2024-06-19
Avoid redirect to migration guide

finestructure:
I'm getting a "page has moved" interstitial that can stay up as long as a second, so we may want to link to the redirect target instead.


daveverwer (approved):
I mentioned this in the original PR, but this is the new URL for the permanent location of the guide. I thought the change was imminent, but if we merge this we'll need to change it back when the guide moves.

PR #3156, opened 2024-06-21
Adopting Swift 6 language mode

gwynne:
This work is not complete in its present form. The main app itself builds in Swift 6 language mode with these changes, but due to problems with transitive language mode affecting dependencies, the tests do not yet build.


finestructure:
I think we're really close now, or we might even be there with what _we_ can do given the remaining warnings and expected upstream compiler changes.

I've added the output of a `swift build --build-tests` run to a gist for reference:

https://gist.github.com/finestructure/37cecd4c1b8b2c38c2850ff61358bb56

Total: 187 warnings
- 145 converting non-sendable function value to '@Sendable (from AppEnvironment)
- 42 code after 'throw' will never be executed (from throw XCTSkip, introduced to skip failing tests)

Xcode 16b2 counts this as a total of 41 warnings:
- 34 converting non-sendable function value to '@Sendable
- 7 code after 'throw' will never be executed

It's weird that Xcode counts differently and unfortunate that `swift build` doesn't give a count summary but that's what we've got.

@gwynne , I believe the `converting non-sendable function value to '@Sendable` warning is one that should go away with a compiler change/update, isn't it? I believe there was also a simple fix that we don't want to deploy because it'll be redundant with the compiler fix?

finestructure:
As to the tests failures, which I'm looking into, there are currently

- 78 failures across 26 tests
- 7 tests we're skipping due
  - a crash with NSInternalInconsistencyException in SnapshotTesting/AssertSnapshot.swift:317
  - a failure due to https://github.com/vapor/fluent-kit/issues/612

The Fluent issue probably affects at least some if not all of the remaining 26 test failures, TBD.

finestructure:
I've updated the gist after the last commit, we're now down to 0 concurrency warnings in our code 🎉

finestructure:
All tests are now passing with Xcode 16b3. It's possible now to use Xcode 16b3 as the daily driver for development. (We already build and run with Swift 6 in the base image for deployment, so 15.4 was technically lagging behind.)

finestructure (approved):

PR #3159, opened 2024-06-24
Issue 3133 build details

### LINKED ISSUE #3133: Show revision on build details page for default branch builds
While we do log the tested revision in the build logs themselves, it's hard to find there and moreover it can be pushed off the top in case of noisy builds (looking at you here, xcodebuild).

We should show the git sha alongside the build command to make it easier to verify which revision was actually built.

finestructure:
Fixes #3133

### DIFF
--- a/Sources/App/Views/PackageController/Builds/BuildShow+View.swift
+++ b/Sources/App/Views/PackageController/Builds/BuildShow+View.swift
@@ -12,6 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+import Foundation
+
 import Plot
 
 
@@ -84,7 +86,10 @@ enum BuildShow {
                             .text(" using "),
                             .strong(.text($0)),
                             .text(" at "),
-                            .strong(.text(model.reference))
+                            .strong(.text(model.reference)),
+                            .text(" (\(model.commit.prefix(6)))"),
+                            .text(" on "),
+                            .strong(.text(DateFormatter.displayFormatter.string(from: model.buildDate) + " UTC"))
                         )
                     },
                     .text(".")
@@ -132,5 +137,17 @@ private extension Build.Status {
             case .failed: return "red"
             case .infrastructureError, .triggered, .timeout: return ""
         }
+    }
+}
+
+
+private extension DateFormatter {
+    static var displayFormatter: DateFormatter {
+        let formatter = DateFormatter()
+        formatter.dateFormat = "d MMM yyyy HH:mm:ss"
+        formatter.locale = Locale(identifier: "en_US_POSIX")
+        formatter.timeZone = TimeZone(secondsFromGMT: 0)
+
+        return formatter
     }
 }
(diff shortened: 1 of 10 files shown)


Sources/App/Views/PackageController/Builds/BuildShow+View.swift
```diff
@@ -87,6 +89,13 @@ enum BuildShow {
                             .strong(.text(model.reference))
                         )
                     },
+                    .unwrap(model.buildInfo.buildDetails) {
+                        .group(
+                            .text(" (\($0.commitHash.prefix(6)))"),
+                            .text(" on "),
+                            .strong(.text(DateFormatter.displayFormatter.string(from: $0.buildDate) + " UTC"))
```
daveverwer, line 96: We do it in a separate PR if you like, but we should also add build duration here. People have no way of telling whether they are near our timeout at the moment.

Sources/App/Views/PackageController/Builds/BuildShow+View.swift
```diff
@@ -134,3 +143,15 @@ private extension Build.Status {
         }
     }
 }
+
+
+private extension DateFormatter {
+    static var displayFormatter: DateFormatter {
```
daveverwer, line 149: I've been trying to consolidate the date formatters we use in the front end into `Date+ext.swift` so that we standardise a bit on how we display dates. It's no problem that this is a new format as we rarely need the specific time of something, but I wonder if we could move this there so we have a standard "with time" formatter?

daveverwer (approved):

Sources/App/Views/PackageController/Builds/BuildShow+View.swift
finestructure: I've kept this private because I don't think we want this date formatter to be used in general. It's printing the date in UTC and we should ideally have this represented in the client's locale. I'm not sure how to best do this so UTC is the best I could come up with to get this out for now.

Sources/App/Views/PackageController/Builds/BuildShow+View.swift
finestructure: We should add that but more importantly I'd like to display the Swift 6 error count per build here as well. At that point that shouldn't just be a run-on sentence but rather something like a list

```
Build date:
Build duration:
Swift 6 errors:
```

Sources/App/Views/PackageController/Builds/BuildShow+View.swift
daveverwer: OK, lets add that as a comment then. I'll make a PR.

Sources/App/Views/PackageController/Builds/BuildShow+View.swift
daveverwer: 👍

If you want to open a PR that brings the Swift 6 error data through I can format it into the page in a way that makes sense.

PR #3183, opened 2024-07-03
Remove Runner ID from build information

daveverwer:
Runner ID is no longer very useful as it just either indicates Orka or Linux 1/2, so lets remove it.


finestructure (approved):

finestructure:
Hold on a sec

Sources/App/Commands/Alerting.swift
```diff
@@ -83,7 +83,6 @@ extension Alerting {
             self.updatedAt = build.updatedAt!
             self.builderVersion = build.builderVersion
             self.platform = build.platform
-            self.runnerId = build.runnerId
```
finestructure, line 86: We don't want to remove this from Alerting. The linux info can be helpful (and even the orka one could be if there's a stuck runner).

finestructure (changes requested):
Need to keep the Alerting runnerId

Sources/App/Commands/Alerting.swift
daveverwer: OK, so we need to leave the field and just remove it user-facing?

finestructure: Yes, exactly.

Sources/App/Commands/Alerting.swift
```diff
@@ -123,7 +122,6 @@ extension Alerting {
             .field(\.$updatedAt)
             .field(\.$builderVersion)
             .field(\.$platform)
-            .field(\.$runnerId)
```
finestructure, line 126: Keep this as well.

finestructure: I.e. no changes in `Alerting.swift`, the rest should all be front-end only.

Sources/App/Migrations/078/RemoveBuildsRunnerId.swift
```diff
@@ -0,0 +1,29 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import Fluent
+
+struct RemoveBuildsRunnerId: AsyncMigration {
+    func prepare(on database: Database) async throws {
+        try await database.schema("builds")
+            .deleteField("runner_id")
```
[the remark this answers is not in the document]
finestructure: We definitely still want to report build ids.

Sources/App/Models/Build.swift
```diff
@@ -71,9 +71,6 @@ final class Build: @unchecked Sendable, Model, Content {
     @Field(key: "platform")
     var platform: Platform
 
-    @Field(key: "runner_id")
-    var runnerId: String?
-
```
finestructure, line 76: Need to keep this

Sources/App/Models/Build.swift
```diff
@@ -101,7 +98,6 @@ final class Build: @unchecked Sendable, Model, Content {
         self.jobUrl = jobUrl
         self.logUrl = logUrl
         self.platform = platform
-        self.runnerId = runnerId
```
finestructure, line 104: and this

Sources/App/Commands/Alerting.swift
finestructure: No, there's more that we need to keep. Need time to look through.

finestructure:
TBH, I think it'd be easier and safer to redo this and just limit the change to the build monitor page.

finestructure (approved):

PR #3198, opened 2024-07-09
More Apple repos were moved last night

finestructure:


finestructure:
Self-merging this to fix the doc alerts

PR #3220, opened 2024-07-15
More swift 6 fixes

finestructure:
I noticed we were still showing build errors in the latest run even though it was building with the [Swift 6 fixes merged](https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/issues/3156) in:

CleanShot 2024-07-15 at 14 28 10@2x

This addresses a lot of them in a first pass. These are fixes that @gwynne had already applied and which I accidentally rolled back, because they didn't trigger warnings in my runs.

That was because for some reason we'd actually disabled the concurrency check in the package manifest. It's not clear to me why/how it still showed _some_ Swift 6 warnings. That's why that omission wasn't obvious.

With these fixes, Xcode 16b3 shows exactly 34 warnings:

  1. 1 dependency 'swift-case-paths' is not used by any target (temporary)
  2. 32 Converting non-sendable function value
  3. 1 Static property 'json' is not concurrency-safe (in SwiftOpenAPI dependency)

Item 1. is not Swift 6 related, 3. should go away once the compiler stops applying Swift 6 warnings to dependencies (assuming that's the reason - either way, it's beyond our direct control).

That leaves 32 concurrency warnings in our code base. These are all of a type that is likely a false positive (has been reported) and should go away in a future compiler release. We _can_ fix this in the meantime by expanding all function parameters into closures.

Since there are quite a few and it's going to make things more verbose we may want to hold off on that. (We've done a bit of this in the previous Swift 6 PR but it was in the dependency code that's going to to away soon.)

Note that the count of 32 warnings is higher than the 5 warnings reported by the preview processing, because a) the 32 warnings also include test targets and b) the stats reporting we use in the preview builds has some aggregation mechanism that deduplicates warnings, I believe.

I've reproduced the number of 5 warnings running the same build command as we do in the preview build, so the reporting is correct. It's just not aligning 1:1 with what Xcode shows.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -1,4 +1,4 @@
-// swift-tools-version:6.0
+// swift-tools-version:5.10
 
 // Copyright Dave Verwer, Sven A. Schmidt, and other contributors.
 //
@@ -109,7 +109,9 @@ let package = Package(
                     ],
                     swiftSettings: swiftSettings)
     ],
-    swiftLanguageVersions: [.v6]
+    swiftLanguageVersions: [.v5]
 )
 
-var swiftSettings: [SwiftSetting] { [] }
+var swiftSettings: [SwiftSetting] { [
+    .enableExperimentalFeature("StrictConcurrency"),
+] }
--- a/Package.swift
+++ b/Package.swift
@@ -1,4 +1,4 @@
-// swift-tools-version:5.10
+// swift-tools-version:6.0
 
 // Copyright Dave Verwer, Sven A. Schmidt, and other contributors.
 //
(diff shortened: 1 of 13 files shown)


Package.swift
```diff
@@ -113,5 +113,6 @@ let package = Package(
 )
 
 var swiftSettings: [SwiftSetting] { [
-//    .enableExperimentalFeature("StrictConcurrency"),
+    .enableExperimentalFeature("StrictConcurrency"),
+    .enableUpcomingFeature("StrictConcurrency")
```
finestructure, line 117: I'm enabling both, because otherwise `swift build` and Xcode build disagree about warnings. I seem like they each take one but not the other parameter? I didn't dig into this further since this just works and avoid confusing.

I also tried dropping the settings entirely and setting language mode 6 but that would mean we actually have to fix all the issues (which we can't, because one is in a dependency).

daveverwer (approved):

PR #3228, opened 2024-07-18
Added a front end debug panel for inspecting useful page metadata

daveverwer:
In preparation for final testing before merging https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/pull/3078, I wanted a way to check canonical URLs against the current browser window URL quickly and efficiently, without needing to locate the canonical in the web inspector manually each time. 

While I was building this console, I thought it’d be a good place for some of the other data we currently hide in HTML comments, like package ID and score.

This console is hidden from end-users completely and must be switched on manually from the JavaScript console:

```javascript
localStorage.setItem('spiDebug', 'true');
```

Once shown, it can be temporarily hidden with the “Hide” button, if it’s blocking something at the bottom of the page, or completely disabled with the “Disable” button. After disabling it would need switching back on with the console command above.

This PR also introduces shared styling and JavaScript between the main site and the Vue DocC renderer through new `shared.css` and `shared.js` files.

Screenshot 2024-07-18 at 10 31 11@2x


daveverwer:
Dark mode:
Screenshot 2024-07-18 at 10 34 07@2x

finestructure (approved):

PR #3240, opened 2024-07-22
Moved Plot fork into our organisation

daveverwer:
I am still tracking `main` here rather than tagging versions as I don't really want to overwrite any future version tags that John might make, and no one else should be using this fork, so it's not super important that we won't make releases.


finestructure:
FYI tags are separate per repo. We can tag our own versions here.

finestructure (approved):

daveverwer:
> FYI tags are separate per repo. We can tag our own versions here.

Of course, but I didn't want to tag a 0.15.0, which would conflict with a 0.15.0 that John tags in the future. I do plan to merge in any changes from upstream (if they happen).

PR #3250, opened 2024-07-24
Improve build details

finestructure:
This moves the tested reference info closer to the package name and adds Swift 6 data race info (when available).

CleanShot 2024-07-24 at 11 56 18@2x

Current live version:

CleanShot 2024-07-24 at 11 35 28@2x


daveverwer (approved):
Did this not trigger a snapshot change? If not, we should probably snapshot that build detail page.

daveverwer:
In your screenshot, this looked great:

Screenshot 2024-07-24 at 11 14 06@2x

But in the snapshot the spacing looks off:

Screenshot 2024-07-24 at 11 14 52@2x

Hold off on merging this. I'm running it locally to see what's up.

daveverwer:
Looks great locally. Merging.

PR #3278, opened 2024-08-08
Bump to 16

finestructure:
Both `dev` and `prod` are on version 16.3 now.

In order to switch over locally, run

```
make db-down db-up
```

⚠️ This will delete local data ⚠️

Check the postgres version via the following SQL query:

```sql
SELECT version();
```


daveverwer (approved):

finestructure:
Putting this back in draft, because I'm seeing a huge performance regression when running the tests locally with the `postgres:16.3-alpine` container: 360 seconds vs 90 seconds.

I'm guessing it's related to the temp fs flag `--tmpfs /pgdata:rw,noexec,nosuid,size=1024m` we're using. Needs investigation.

finestructure:
No, it's Postgres itself. Here are the timing just running the `AnalyzerTests`:

```
❯ make db-down-test db-up-test && sleep 3 && time swift test --disable-automatic-resolution --filter AppTests.AnalyzerTests
```

for different versions of Postgres (plus with/without tempfs in two cases):

- pg13 + tempfs: 11s
- pg16 + tempfs: 27s
- pg13: 15s
- pg14: 37s
- pg15: 37s
- pg16: 34s

Next guess it's related to the table truncation we're doing in db test setup.

PR #3298, opened 2024-08-16
Don't bump packages.updated_at when receiving build reports

### LINKED ISSUE #3290: Analysis/Ingestion taking a very long time for some packages
I pushed a change to Leftpad's main branch at around midday yesterday and as of 6pm today, 30 hours later, the main branch change still has not been picked up by analysis and the README change still has not been picked up by ingestion.

Screenshot 2024-08-13 at 18 41 45@2x

On staging, both changes were picked up in reasonable time.

### LINKED ISSUE #3292: Incorrect package data for grpc-swift

- https://swiftpackageindex.com/grpc/grpc-swift/

The last changes to the default branch (`main`) are showing as 21 days ago. It has had almost daily changes since then.

This is also reflected in the build status: the last builds for the default branch were on 24th July.

The Swift 6.0 beta builds are also showing as "Pending" and have done over the last few days when I've checked.

**If the problem relates to the compatibility matrix, does it apply to specific builds?**

N/a

**Screenshots**

N/a

**Additional context**

- W.r.t. Swift 6, the package does have a version specific manifest: `Package@swift-6.swift`.
- One change we made soon after the last reported main builds was to change from `swiftLanguageVersion` to `swiftLanguageMode` (https://github.com/grpc/grpc-swift/commit/0aad7094c3f3c973b7dee92c05b8fa64a5c29177) in the Swift 6 manifest. It may or may not be related (it looks like your Swift 6 builder image is currently new enough to handle `swiftLanguageMode`)

finestructure:
Fixes #3290
Fixes #3292

### DIFF
--- a/Sources/App/Models/Package.swift
+++ b/Sources/App/Models/Package.swift
@@ -258,8 +258,7 @@ extension Package {
                 AND b.status = 'ok'
                 GROUP BY b.platform
                 HAVING count(*) > 0
-            ),
-            updated_at = NOW()
+            )
             WHERE p.id = \#(bind: packageId)
             """#
         ).run()
--- a/Tests/AppTests/ApiTests.swift
+++ b/Tests/AppTests/ApiTests.swift
@@ -408,6 +408,50 @@ class ApiTests: AppTestCase {
             })
     }
 
+    func test_post_buildReport_package_updatedAt() async throws {
+        // Ensure we don't change packages.updatedAt when receiving a build report.
+        // https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/issues/3290#issuecomment-2293101104
+        // setup
+        Current.builderToken = { "secr3t" }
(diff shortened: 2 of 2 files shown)


daveverwer (approved):

Sources/App/Models/Package.swift
```diff
@@ -258,8 +258,7 @@ extension Package {
                 AND b.status = 'ok'
                 GROUP BY b.platform
                 HAVING count(*) > 0
-            ),
-            updated_at = NOW()
+            )
```
finestructure, line 261: We're technically updating the package record but what we're really doing is aggregating up build data. So it's not the package per se that's changing.

A cleaner solution would perhaps be to move this aggregate field `platform_compatibility` to a separate table since we're using `packages` as a convenient place to stick it.

Maybe `packages_aggregates`. It could hold `platform_compatibility` and `score_details`. I'll add an issue to track.

PR #3305, opened 2024-08-19
Add support for a HOMEPAGE_INTERSTITIAL env variable to allow us to put up a status page

finestructure:
I've updated our dev db to Postgres 16 yesterday and it takes the db offline for about 8 minutes.

I was planning to do the same next Sunday morning but rather than have the server show an obscure error page we could put up a simple, configurable interstitial on the home page via an env variable `HOMEPAGE_INTERSTITIAL`. Its content can/should be Markdown.

This could also be useful in case we ever hit db issues, because this page can be served when the db is down.

CleanShot 2024-08-19 at 12 10 02@2x


daveverwer (approved):
I might give this a separate template from the Markdown page just so I can give it a bit of space, but I can do that as a separate PR.

finestructure:
That'd be great, it's a bit tight the way it is 👌

PR #3310, opened 2024-08-19
change alert text color based on light/dark mode

rahafjrw:
Fix alert text to be white if in dark mode and black if in light mode
Before:
before-darkmode
After:
after-darkmode


daveverwer (approved):

daveverwer:
I just rebased this change on top of today's changes. Let's hope that fixes it.

daveverwer:
The rebase did it 👍 Thanks so much for this @rahafjrw!

PR #3313, opened 2024-08-20
Adds styling of the maintenance page

daveverwer:
As part of this, I split the test into two. One tests the request and that the maintenance page is displayed and the other is a regular snapshot test of the new MaintenanceIndex view. I did this so the snapshot HTML is not compressed so that we get easier diffs.

Screenshot 2024-08-20 at 16 08 00@2x


Sources/App/Views/Maintenance/Maintenance+Model.swift
```diff
@@ -0,0 +1,31 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import Foundation
+import Ink
+
+enum MaintenanceIndex {
```
finestructure, line 18: Nit: I called this "interstitial" because I can see us using it also in case of unscheduled downtime, say if we had a db issue and wanted to put up a notice.

Not sure if "interstitial" is the best term but at the very least we should make the env variable align with what we end up calling it.

daveverwer: The only context I've seen interstitial used in related to this is for adverts. I think unscheduled downtime could also be named maintenance, and it's probably a slightly more common term than interstitial. I'll change it to match.

finestructure: Or maybe placeholder. I'm fine with maintenance as well, my main point is that I'd then want to rename the env var and the Current property accordingly.

daveverwer:
Changed across the board to `maintenanceMessage` and `MAINTENANCE_MESSAGE`.

daveverwer:
Oh and I also changed the icon. Even a server maintenance isn't anything we're "warning" anyone about, so the ⚠️ symbol was less appropriate than an information icon.

Screenshot 2024-08-21 at 12 39 49@2x

finestructure (approved):

PR #3316, opened 2024-08-21
Snapshots with Xcode 16 beta 6

daveverwer:
I got two snapshot changes with beta6. Hopefully they flip the same way with the Linux image!


daveverwer:
```
    ✔ test_case_insensitive_owner_matching (0.291 seconds)
Error:     test_generate_for_owner, failed - Snapshot does not match reference.
    ✔ test_generate_for_owner_noResults (0.251 seconds)
Error:     test_generate_from_urls, failed - Snapshot does not match reference.
```

I suspected this might happen 😬

finestructure:
Yes, this is what I dealt with in https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/issues/3307

We'll need to update Xcode and the docker base image at the same time.

PR #3326, opened 2024-08-23
More elf to async await

finestructure:
We're so close now, might as well take it all the way...

- [x] ad hoc deploy and test metrics output (this is a nontrivial conversion to a/a)


daveverwer (approved):

finestructure:
(Still a couple more to come. Breaking these down a bit to create deploy points.)

finestructure:
Metrics are reporting correctly:

CleanShot 2024-08-23 at 14 57 26@2x

PR #3349, opened 2024-08-31
Show correct code snippet for products type of plugin

### LINKED ISSUE #3297: Deal with non-library products better in "Use this Package"
As reported via Discord, “Use this Package” should provide the correct code when selecting a plugin product from the drop-down menu.

Screenshot 2024-08-16 at 09 49 44@2x

The code for this should be:

```swift
.plugin(name: "swift-openapi-generator", package: "swift-openapi-generator")
```

We might also want to filter out `.executable` products at the same time, as I don’t believe there’s anything useful we can do here. Or maybe we want to include them, but instead of code we could include some text saying how to use a command line tool.

supersonicbyte:
Fixes #3297

This PR addresse this issue: https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/issues/3297

Wrong code snippet shown for non-library products. Additionally, executable products are filtered out and will be addressed in a different PR.

### DIFF
--- a/Sources/App/Controllers/API/API+PackageController+GetRoute+Model.swift
+++ b/Sources/App/Controllers/API/API+PackageController+GetRoute+Model.swift
@@ -235,6 +235,17 @@ extension API.PackageController.GetRoute.Model {
             case executable
             case plugin
             
+            var stringValue: String {
+                switch self {
+                case .library:
+                    return "library"
+                case .executable:
+                    return "executable"
+                case .plugin:
+                    return "plugin"
+                }
+            }
+            
             init?(_ productType: App.ProductType) {
                 switch productType {
                     case .executable:
--- a/Sources/App/Controllers/API/API+PackageController+GetRoute+Model.swift
+++ b/Sources/App/Controllers/API/API+PackageController+GetRoute+Model.swift
@@ -230,21 +230,10 @@ extension API.PackageController.GetRoute.Model {
             self.type = type
         }
 
-        enum ProductType: Codable, Equatable {
+        enum ProductType: String, Codable, Equatable {
             case library
             case executable
             case plugin
-            
-            var stringValue: String {
-                switch self {
-                case .library:
-                    return "library"
-                case .executable:
-                    return "executable"
-                case .plugin:
-                    return "plugin"
-                }
-            }
             
             init?(_ productType: App.ProductType) {
                 switch productType {
--- a/Sources/App/Views/PackageController/GetRoute.Model+ext.swift
+++ b/Sources/App/Views/PackageController/GetRoute.Model+ext.swift
@@ -415,10 +415,12 @@ extension API.PackageController.GetRoute.Model {
                     .data(named: "action", value: "input->use-this-package-panel#updateProductSnippet"),
                     .attribute(named: "name", value: "products"),
                     .id("products"),
-                    .forEach(products, { product in
+                    // Filter out products of type `executable` until we add support for them.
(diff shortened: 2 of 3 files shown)


Sources/App/Views/PackageController/GetRoute.Model+ext.swift
```diff
@@ -415,10 +415,12 @@ extension API.PackageController.GetRoute.Model {
                     .data(named: "action", value: "input->use-this-package-panel#updateProductSnippet"),
                     .attribute(named: "name", value: "products"),
                     .id("products"),
-                    .forEach(products, { product in
+                    // Filter out products of type `executable` until we add support for them.
+                    .forEach(products.filter({ $0.type != .executable }), { product in
                             .option(
                                 .data(named: "package", value: package),
                                 .data(named: "product", value: product.name),
+                                .data(named: "type", value: product.type.stringValue),
```
finestructure, line 423:
```suggestion
                                .data(named: "type", value: product.type.rawValue),
```

Sources/App/Controllers/API/API+PackageController+GetRoute+Model.swift
```diff
@@ -235,6 +235,17 @@ extension API.PackageController.GetRoute.Model {
...
+                switch self {
+                case .library:
+                    return "library"
+                case .executable:
+                    return "executable"
+                case .plugin:
+                    return "plugin"
+                }
+            }
+
```
finestructure, line 248:
```suggestion
```

We can remove this in favour of using `rawValue`.

finestructure (commented):
Thanks for the PR, @supersonicbyte!

I've left a comment regarding a small thing about using `rawValue`.

Another thing I noticed when running your changes is that I'm getting a blank snippet:

Screenshot 2024-09-01 at 16 46 55

Screenshot 2024-09-01 at 16 47 20

Does this work for you or is this a problem on my end?

Sources/App/Controllers/API/API+PackageController+GetRoute+Model.swift
supersonicbyte: Hey, I suppose you added a conformation to `String` on the `ProductType`?

supersonicbyte:
> Thanks for the PR, @supersonicbyte!
> 
> I've left a comment regarding a small thing about using `rawValue`.
> 
> Another thing I noticed when running your changes is that I'm getting a blank snippet:
> 
> Screenshot 2024-09-01 at 16 46 55 Screenshot 2024-09-01 at 16 47 20
> Does this work for you or is this a problem on my end?

Hey, thanks for the feedback! Actually, I tried your suggestion and when I change the code to use `rawValue` the `type` on the controller cannot be found and that's why you get a blank. Running the code without the `rawValue` works. I will add your suggestion and fix it to work properly.

supersonicbyte:
@finestructure Ok, updated! Please take a look when you find the time.

daveverwer (approved):
This is great, thanks so much @supersonicbyte! Thank you for your contribution. 

I had some thoughts on how we handle executables, but as discussed yesterday, that would be a separate PR. I'll also make a new issue for it, so this can be considered to fix and close #3297.

daveverwer:
@cla-bot recheck please

daveverwer:
@cla-bot check

finestructure:
I've opened a superseding PR #3351 to run the CI, update the snapshots, and fix a linting issue.

This one here should automatically be closed once #3351 is merged.

Issue #551, opened 2020-07-22
Show if a package is a fork


daveverwer:
The SwiftPM Library recorded if a repository was a fork of another repository and displayed that and I think the SPI should do the same.

Several parts to this:

- [ ] Record a `parent_url` field in the `repositories` table and populate it from the GitHub API call.
- [ ] Display whether a repository is a fork on the Package page and give a link back to the original repository.

elfenlaid:
Here's what I found:

- Unused `forked_from_id: Repository?` in `repositories` 
It seems like an overshot for a simple link to the original repository 🤔
Can it be replaced with the `parent_url` field?

- Unused `isFork` flag in GraphQL request `Github.Repository`
Probably should be also replaced with
   ```graphql
   parent {
       url
   }
   ```

daveverwer:
Yea there's a bit of history to this feature. The original SwiftPM Library actually supported displaying when a repo was a fork of another repo and linked people out to the original repository as well. It's a useful indicator, even if the fork is the one being more actively maintained.

So, this field existed as we were originally planning to replicate all existing functionality for launch, but this feature didn't quite make it and then other things took priority.

I'd love to have it as a feature though, it's a great indicator for that top section. What I'd ideally like is to be able to put something like this in:

```
This package is a fork of [example/OriginalPackage]
```

Where [example/OriginalPackage] is a link to either:

* The package page on the SPI for the original repository, if we have it.
* The GitHub page for the forked repository, if we don't have the original in our package list.

Sherlouk:
Worth noting (though doesn't stop this ticket going through) that the package list nightly scripts currently ignore all dependencies which are forks: https://github.com/SwiftPackageIndex/PackageList-Validator/blob/100a768c6c5d2840417576a85bc4d2bf8c9bab5a/Sources/ValidatorCore/Commands/CheckDependencies.swift#L187

Once we add support for it here, we may want to consider removing that restriction.

finestructure:
I think that's still sensible - we don't necessarily want to greedily pull in all forks in the nightly. It's different if someone submits a fork for including, that's more deliberate than the crawler.

elfenlaid:
> Where [example/OriginalPackage] is a link to either:
> 
>     * The package page on the SPI for the original repository, if we have it.
> 
>     * The GitHub page for the forked repository, if we don't have the original in our package list.

[...]

PR #3355, opened 2024-09-02
Add Forked From

supersonicbyte:
This PR tries to resolve the following issue: https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/issues/551

The `forkedFrom` attribute on the `Repository` will enable us to determine if the package is forked from another repo, and we can reflect it on the UI.

### DIFF
--- a/Sources/App/Commands/Ingest.swift
+++ b/Sources/App/Commands/Ingest.swift
@@ -191,7 +191,7 @@ func updateRepository(on database: Database,
     do {
         if let parentUrl = metadata.repository?.normalizedParentUrl {
             if let packageId = try await Package.query(on: database)
-                .filter(\.$url == parentUrl)
+                .filter(\.$url, .custom("ilike"), parentUrl)
                 .first()?.id {
                 fork = .parentId(packageId)
             } else {
--- a/Sources/App/Commands/Ingest.swift
+++ b/Sources/App/Commands/Ingest.swift
@@ -124,13 +124,30 @@ func ingest(client: Client,
                         Current.logger().warning("storeS3Readme failed")
                         s3Readme = .error("\(error)")
                     }
+                    
+                    var fork: Fork?
+                    do {
+                        if let parentUrl = metadata.repository?.normalizedParentUrl {
+                            if let packageId = try await Package.query(on: database)
+                                .filter(\.$url == parentUrl)
+                                .first()?.id {
+                                fork = .parentId(packageId)
+                            } else {
+                                fork = .parentURL(parentUrl)
+                            }
+                        }
+                    } catch {
+                        Current.logger().warning("updating forked from failed")
+                    }
 
                     try await updateRepository(on: database,
                                                for: repo,
                                                metadata: metadata,
                                                licenseInfo: license,
                                                readmeInfo: readme,
-                                               s3Readme: s3Readme)
+                                               s3Readme: s3Readme,
+                                               forkedFrom: fork)
+                        
                     return pkg
                 }
 
@@ -177,7 +194,8 @@ func updateRepository(on database: Database,
                       metadata: Github.Metadata,
                       licenseInfo: Github.License?,
                       readmeInfo: Github.Readme?,
-                      s3Readme: S3Readme?) async throws {
+                      s3Readme: S3Readme?,
+                      forkedFrom: Fork? = nil) async throws {
     guard let repoMetadata = metadata.repository else {
         if repository.$package.value == nil {
             try await repository.$package.load(on: database)
@@ -208,6 +226,7 @@ func updateRepository(on database: Database,
     repository.releases = repoMetadata.releases.nodes.map(Release.init(from:))
     repository.stars = repoMetadata.stargazerCount
     repository.summary = repoMetadata.description
+    repository.forkedFrom = forkedFrom
 
     try await repository.save(on: database)
 }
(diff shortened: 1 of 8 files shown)


Sources/App/Commands/Ingest.swift
```diff
@@ -124,13 +124,30 @@ func ingest(client: Client,
...
+                                .filter(\.$url == parentUrl)
+                                .first()?.id {
+                                fork = .parentId(packageId)
+                            } else {
+                                fork = .parentURL(parentUrl)
+                            }
+                        }
+                    } catch {
+                        Current.logger().warning("updating forked from failed")
+                    }
```
daveverwer, line 141: I'll take a proper look at this soon, but I noticed this as a first small issue. I wonder if it might be better to do this inside `updateRepository`. It would save us passing one more variable into that function as we already pass in `metadata`, which drives this logic.

I can give it a proper review tomorrow, but I thought I'd mention this first as it may change the structure of the PR a little bit.

supersonicbyte: Thanks for the feedback. It definitely makes sense to move this to the `updateRepository`. Will do that.

daveverwer:
Just one note here missing from the PR description in case @finestructure takes a look and is missing the context. @supersonicbyte and I chatted and agreed this should be done in two PRs. One to populate the data and one to show it once populated. This is the first of the PRs.

Sources/App/Commands/Ingest.swift
```diff
@@ -185,6 +186,21 @@ func updateRepository(on database: Database,
         throw AppError.genericError(repository.package.id,
                                     "repository metadata is nil for package \(repository.name ?? "unknown")")
     }
+    
+    var fork: Fork?
+    do {
+        if let parentUrl = metadata.repository?.normalizedParentUrl {
+            if let packageId = try await Package.query(on: database)
+                .filter(\.$url == parentUrl)
```
finestructure, line 194: We'll want to make sure to do a case-insensitive compare here:

```suggestion
                .filter(\.$url, .custom("ilike"), parentUrl)
```

finestructure (commented):
Thanks for the PR, @supersonicbyte, it'll be nice to have this show up on the page!

Sources/App/Commands/Ingest.swift
supersonicbyte: Good catch! Thanks.

daveverwer (approved):
Apologies for the slow review on this. I was convinced we had some code already that normalised a package URL, but I had a good look for it yesterday and couldn't find it, so it's possible it got removed.

I waited until this morning to merge this so I could keep an eye on it on staging. I'll merge it now and watch its progress over the next few hours.

Thanks so much for this contribution, @supersonicbyte! This is great. Let's chat about how to represent it in the UI next 👍 Thank you!

finestructure:
> Apologies for the slow review on this. I was convinced we had some code already that normalised a package URL, but I had a good look for it yesterday and couldn't find it, so it's possible it got removed.

We do over in the PackageList repo. I'd half written a reply to copy its `URL` based implementation and re-use it that way but didn't want to make this more complicated than necessary for @supersonicbyte . I've made a note to see if it makes sense and follow-up with a PR if it does.

finestructure:
Sorry, slight tweak to my proposed `getFork` - we shouldn't pass in all of `Metadata` but instead use

```swift
func getFork(on database: Database, parent: Github.Metadata.Parent?) async throws -> Fork? {
    guard let parentUrl = parent?.normalizedURL else { return nil }

    if let packageId = try await Package.query(on: database)
        .filter(\.$url, .custom("ilike"), parentUrl)
        .first()?.id {
        return .parentId(packageId)
    } else {
        return .parentURL(parentUrl)
    }
}
```

with normalisation on the `Parent` instead of `Github.Metadata.Repository`:

```swift
private extension Github.Metadata.Parent {
    // Returns a normalized version of the URL. Adding a `.git` if not present.
    var normalizedURL: String? {
        guard let url else { return nil }
        guard !url.hasSuffix(".git") else { return url }
        let normalizedUrl = url + ".git"
        return normalizedUrl
    }
}
```

The test is also a bit nicer this way:

```swift
    func test_getFork() async throws {
        try await Package(id: .id0, url: "https://github.com/foo/parent.git".url, processingStage: .analysis).save(on: app.db)
        try await Package(url: "https://github.com/bar/forked.git", processingStage: .analysis).save(on: app.db)

        do {  // test lookup when package is in the index
            let fork = try await getFork(on: app.db, parent: .init(url: "https://github.com/foo/parent.git"))
            XCTAssertEqual(fork, .parentId(.id0))
        }

        do {  // test lookup when package is in the index but with different case in URL
            let fork = try await getFork(on: app.db, parent: .init(url: "https://github.com/Foo/Parent.git"))
            XCTAssertEqual(fork, .parentId(.id0))
        }

        do {  // test lookup when package is not in the index
            let fork = try await getFork(on: app.db, parent: .init(url: "https://github.com/some/other.git"))
            XCTAssertEqual(fork, .parentURL("https://github.com/some/other.git"))
        }

        do {  // test lookup when parent url is nil
            let fork = try await getFork(on: app.db, parent: nil)
            XCTAssertEqual(fork, nil)
        }
    }
```

finestructure:
I believe this has been superseded by #3363 ?

PR #3359, opened 2024-09-03
Fix crashes in Xcode 15

finestructure:


Tests/AppTests/PackageCollectionControllerTests.swift
```diff
@@ -58,10 +58,11 @@ class PackageCollectionControllerTests: AppTestCase {
...
         // MUT
+        let encoder = self.encoder
         try await app.test(
             .GET,
             "foo/collection.json",
-            afterResponse: { res async throws in
+            afterResponse: { @MainActor res async throws in
                 // validation
                 XCTAssertEqual(res.status, .ok)
                 let json = try res.content.decode(PackageCollection.self)
```
finestructure, line 68: The snapshot test here (and the others below) still fail due to the ordering issue. We _could_ skip them for `compiler(<6)` but it's safer to just let it be for now. If it becomes a problem in day-to-day running we can revisit that.

daveverwer (approved):

PR #3363, opened 2024-09-05
Add `forked_from` as a pointer either to an existing package or as a reference to a URL

daveverwer:
Supercedes #3355 and implements the first part of #551. UI changes still pending.

@supersonicbyte I was seeing test failures in `test_decode_Metadata_null` and `test_fetchMetadata` which I have fixed by making `Parent` optional in the GitHub metadata. I don’t believe that `parent` is a mandatory field supplied by GitHub, so this feels like a reasonable fix to me but could you verify please?

Thanks!

### DIFF
--- a/Sources/App/Commands/Ingest.swift
+++ b/Sources/App/Commands/Ingest.swift
@@ -191,7 +191,7 @@ func updateRepository(on database: Database,
     do {
         if let parentUrl = metadata.repository?.normalizedParentUrl {
             if let packageId = try await Package.query(on: database)
-                .filter(\.$url == parentUrl)
+                .filter(\.$url, .custom("ilike"), parentUrl)
                 .first()?.id {
                 fork = .parentId(packageId)
             } else {
--- a/Sources/App/Commands/Ingest.swift
+++ b/Sources/App/Commands/Ingest.swift
@@ -124,13 +124,30 @@ func ingest(client: Client,
                         Current.logger().warning("storeS3Readme failed")
                         s3Readme = .error("\(error)")
                     }
+                    
+                    var fork: Fork?
+                    do {
+                        if let parentUrl = metadata.repository?.normalizedParentUrl {
+                            if let packageId = try await Package.query(on: database)
+                                .filter(\.$url == parentUrl)
+                                .first()?.id {
+                                fork = .parentId(packageId)
+                            } else {
+                                fork = .parentURL(parentUrl)
+                            }
+                        }
+                    } catch {
+                        Current.logger().warning("updating forked from failed")
+                    }
 
                     try await updateRepository(on: database,
                                                for: repo,
                                                metadata: metadata,
                                                licenseInfo: license,
                                                readmeInfo: readme,
-                                               s3Readme: s3Readme)
+                                               s3Readme: s3Readme,
+                                               forkedFrom: fork)
+                        
                     return pkg
                 }
 
@@ -177,7 +194,8 @@ func updateRepository(on database: Database,
                       metadata: Github.Metadata,
                       licenseInfo: Github.License?,
                       readmeInfo: Github.Readme?,
-                      s3Readme: S3Readme?) async throws {
+                      s3Readme: S3Readme?,
+                      forkedFrom: Fork? = nil) async throws {
     guard let repoMetadata = metadata.repository else {
         if repository.$package.value == nil {
             try await repository.$package.load(on: database)
@@ -208,6 +226,7 @@ func updateRepository(on database: Database,
     repository.releases = repoMetadata.releases.nodes.map(Release.init(from:))
     repository.stars = repoMetadata.stargazerCount
     repository.summary = repoMetadata.description
+    repository.forkedFrom = forkedFrom
 
     try await repository.save(on: database)
 }
@@ -223,4 +242,14 @@ private extension Github.Metadata {
 private extension Github.Metadata.Repository {
     var repositoryOwner: String? { owner.login }
     var repositoryName: String? { name }
+}
+
+private extension Github.Metadata.Repository {
+    // Returns a normalized version of the URL. Adding a `.git` if not present.
+    var normalizedParentUrl: String? {
+        guard let url = parent.url else { return nil }
+        guard !url.hasSuffix(".git") else { return url }
+        let normalizedUrl = url + ".git"
+        return normalizedUrl
+    }
 }
--- a/Sources/App/Commands/Ingest.swift
+++ b/Sources/App/Commands/Ingest.swift
@@ -125,28 +125,12 @@ func ingest(client: Client,
                         s3Readme = .error("\(error)")
                     }
                     
-                    var fork: Fork?
-                    do {
-                        if let parentUrl = metadata.repository?.normalizedParentUrl {
-                            if let packageId = try await Package.query(on: database)
-                                .filter(\.$url == parentUrl)
-                                .first()?.id {
-                                fork = .parentId(packageId)
-                            } else {
-                                fork = .parentURL(parentUrl)
-                            }
-                        }
-                    } catch {
-                        Current.logger().warning("updating forked from failed")
-                    }
-
                     try await updateRepository(on: database,
                                                for: repo,
                                                metadata: metadata,
                                                licenseInfo: license,
                                                readmeInfo: readme,
-                                               s3Readme: s3Readme,
-                                               forkedFrom: fork)
+                                               s3Readme: s3Readme)
                         
                     return pkg
                 }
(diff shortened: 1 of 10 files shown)


finestructure:
Yes, `parent` is optional in the GraphQL schema:

CleanShot 2024-09-05 at 09 03 31@2x

Note that GraphQL uses sort of a reverse syntax to Swift:

```
optionalField: String          -> var optionalField: String?
nonOptionalField: String!      -> var nonOptionalField: String
```

Sources/App/Models/Repository.swift
```diff
@@ -122,6 +119,9 @@ final class Repository: @unchecked Sendable, Model, Content {
 
     @Field(key: "summary")
     var summary: String?
+    
+    @Field(key: "forked_from")
+    var forkedFrom: Fork?
```
finestructure, line 124: Nit: we're _trying_ (😅) to keep the data fields sorted (although this isn't true generally, because it's something we started half-way through without going back and fixing it everywhere). I'll patch this up real quick.

supersonicbyte (approved):
LGTM! Thanks for the review and fixes. @finestructure could you link the logic for normalizing the repo urls from PackageList? I couldn't find it.

Tests/AppTests/IngestorTests.swift
```diff
@@ -594,4 +597,92 @@ class IngestorTests: AppTestCase {
...
+            XCTFail("Failed to get forked package id")
+            return
+        }
+        
+        let repo = try await Repository
+            .query(on: app.db)
+            .filter(\Repository.$package.$id == forkedId).first()
+        
+        XCTAssertNil(repo?.forkedFrom)
+    }
```
finestructure, line 687: By moving the main functionality to `getFork`, these tests become much easier to set up because we don't need to test all of `ingest`. Instead we just test `getFork` in isolation:

```swift
    func test_getFork() async throws {
        try await Package(id: .id0, url: "https://github.com/foo/parent.git".url, processingStage: .analysis).save(on: app.db)
        try await Package(url: "https://github.com/bar/forked.git", processingStage: .analysis).save(on: app.db)

        var metadata = Github.Metadata.mock

        do {  // test lookup when package is in the index
            metadata.repository?.parent = .init(url: "https://github.com/foo/parent.git")
            let fork = try await getFork(on: app.db, metadata: metadata)
            XCTAssertEqual(fork, .parentId(.id0))
        }

        do {  // test lookup when package is in the index but with different case in URL
            metadata.repository?.parent = .init(url: "https://github.com/Foo/Parent.git")
            let fork = try await getFork(on: app.db, metadata: metadata)
            XCTAssertEqual(fork, .parentId(.id0))
        }

        do {  // test lookup when package is not in the index
            metadata.repository?.parent = .init(url: "https://github.com/some/other.git")
            let fork = try await getFork(on: app.db, metadata: metadata)
            XCTAssertEqual(fork, .parentURL("https://github.com/some/other.git"))
        }

        do {  // test lookup when parent url is nil
            metadata.repository?.parent = .init(url: nil)
            let fork = try await getFork(on: app.db, metadata: metadata)
            XCTAssertEqual(fork, nil)
        }
    }
```

and then we simply extend a single test where we ensure `fork` is written to the db in `updateRepository`. For example, I'd modify `test_updateRepository_update` and add fork (call on or around line 152):

```swift
        try await updateRepository(on: app.db,
                                   for: repo,
                                   metadata: md,
                                   licenseInfo: .init(htmlUrl: "license url"),
                                   readmeInfo: .init(etag: "etag",
                                                     html: "readme html",
                                                     htmlUrl: "readme html url",
                                                     imagesToCache: []),
                                   s3Readme: .cached(s3ObjectUrl: "url", githubEtag: "etag"),
                                   fork: .parentURL("https://github.com/foo/bar.git"))
```

and validate it (around line 168):

```swift
            XCTAssertEqual(repo.forkedFrom, .parentURL("https://github.com/foo/bar.git"))
```

finestructure (changes requested):

finestructure:
> LGTM! Thanks for the review and fixes. @finestructure could you link the logic for normalizing the repo urls from PackageList? I couldn't find it.

It's here: https://github.com/SwiftPackageIndex/PackageList/blob/08b5021396a618f2c596dc6f48c336c82901d966/.github/add_package.swift#L24

This is essentially what determines the normalisation in the `packages.url` field when we add packages.

supersonicbyte:
> > LGTM! Thanks for the review and fixes. @finestructure could you link the logic for normalizing the repo urls from PackageList? I couldn't find it.
> 
> It's here: https://github.com/SwiftPackageIndex/PackageList/blob/08b5021396a618f2c596dc6f48c336c82901d966/.github/add_package.swift#L24
> 
> This is essentially what determines the normalisation in the `packages.url` field when we add packages.

Thanks! That could be definitely reused.

supersonicbyte:
Hey @finestructure I incorporated your suggestions, please take a look when u find the time! I also took the URL approach from `PackageList`, it seems nicer than the way I did it.

Sources/App/Commands/Ingest.swift
```diff
@@ -124,13 +124,22 @@ func ingest(client: Client,
...
                         s3Readme = .error("\(error)")
                     }
+                    
+                    let fork: Fork?
+                    do {
+                        fork = try await getFork(on: database, metadata: metadata)
+                    } catch {
+                        fork = nil
+                        Current.logger().warning("updating forked from failed")
+                    }
```
finestructure, line 134:
```suggestion
                    let fork = try? await getFork(on: database, metadata: metadata)
```

Let's just straight up assign it. `getFork` is technically throwing due to query but we could even write `try? await Package.query` there, to be honest.

In fact, I think

```swift
func getFork(on database: Database, metadata: Github.Metadata) async -> Fork? {
```

is better, because we're mixing using `nil` as an "error" value and `throws` at the same time - my bad!

Sources/App/Commands/Ingest.swift
```diff
@@ -208,10 +218,25 @@ func updateRepository(on database: Database,
     repository.releases = repoMetadata.releases.nodes.map(Release.init(from:))
     repository.stars = repoMetadata.stargazerCount
     repository.summary = repoMetadata.description
+    repository.forkedFrom = fork
 
     try await repository.save(on: database)
 }
 
+func getFork(on database: Database, metadata: Github.Metadata) async throws -> Fork? {
```
finestructure, line 226:
```suggestion
func getFork(on database: Database, metadata: Github.Metadata) async -> Fork? {
```

Sources/App/Commands/Ingest.swift
```diff
@@ -208,10 +218,25 @@ func updateRepository(on database: Database,
...
+    guard let url = metadata.repository?.parent?.url,
+          let parentUrl = URL(string: url)?.normalizedParent?.absoluteString else {
+        return nil
+    }
+
+    if let packageId = try await Package.query(on: database)
```
finestructure, line 232:
```suggestion
    if let packageId = try? await Package.query(on: database)
```

Sources/App/Commands/Ingest.swift
```diff
@@ -224,3 +249,12 @@ private extension Github.Metadata.Repository {
     var repositoryOwner: String? { owner.login }
     var repositoryName: String? { name }
 }
+
+private extension URL {
+    var normalizedParent: Self? {
```
finestructure, line 254:
```suggestion
    var normalized: Self? {
```

as this is for URLs generally.

And I think we should drop the `private` and put it in a new file `URL+ext.swift` in `Core/Extensions`.

Sources/App/Commands/Ingest.swift
finestructure: Also note that I'd left another little tweak to `getFork` in a follow up comment on the other PR https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/pull/3355#issuecomment-2330991491

So in total it should be

```swift
func getFork(on database: Database, parent: Github.Metadata.Parent?) async -> Fork? {
    guard let parentUrl = parent?.normalizedURL else { return nil }

    if let packageId = try? await Package.query(on: database)
        .filter(\.$url, .custom("ilike"), parentUrl)
        .first()?.id {
        return .parentId(packageId)
    } else {
        return .parentURL(parentUrl)
    }
}
```

Sorry for the back and forth!

finestructure (changes requested):
Looks good! Only a couple of small things left that are essentially changes to my previously proposed changes.

supersonicbyte:
@finestructure thanks for the feedback! Here we go again, could you please review again? Thanks!

Sources/App/Commands/Ingest.swift
```diff
@@ -124,13 +124,16 @@ func ingest(client: Client,
                         Current.logger().warning("storeS3Readme failed")
                         s3Readme = .error("\(error)")
                     }
+                    
+                    let fork: Fork? = try? await getFork(on: database, parent: metadata.repository?.parent)
```
finestructure, line 128: There's a warning here, I'll fix it up real quick.

finestructure (approved):

PR #3388, opened 2024-09-15
Merge ForkedFromResult and ForkedFromInfo

finestructure:
As discussed, merges ForkedFromResult into ForkedFrom Info.


Sources/App/Controllers/API/API+PackageController+GetRoute+Model.swift
```diff
@@ -370,32 +355,11 @@ extension API.PackageController.GetRoute.Model {
...
-        }
-
-        var ownerURL: String? {
-            switch self {
-            case .fromSPI(_, let owner, _, _, _):
-                return SiteURL.author(.value(owner)).relativeURL()
-            case .fromGitHub:
-                return nil
-            }
-        }
```
finestructure, line 398: I've moved these two view related properties into an extension in the file under `Views/PackageController/GetRoute.Model+ext.swift`. That way, view related things don't distract from the basic definition and appear "closer" to where they're actually used.

Sources/App/Controllers/API/API+PackageController+GetRoute+Model.swift
```diff
@@ -197,7 +182,7 @@ extension API.PackageController.GetRoute {
                 preReleaseReference: result.preReleaseVersion?.reference,
                 fundingLinks: result.repository.fundingLinks,
                 swift6Readiness: swift6Readiness,
-                forkedFromResult: forkedFromResult
+                forkedFromInfo: forkedFromInfo
```
finestructure, line 185: The model `init` now doesn't do any logic with `ForkedFromInfo` - it just assigns it. That makes the tests simpler, see below.

Sources/App/Controllers/API/API+PackageController+GetRoute.swift
```diff
@@ -49,6 +45,7 @@ extension API.PackageController {
             async let buildInfo = API.PackageController.BuildInfo.query(on: database,
                                                                         owner: owner,
                                                                         repository: repository)
+            async let forkedFromInfo = forkedFromInfo(on: database, fork: packageResult.repository.forkedFrom)
```
finestructure, line 48: We're potentially running another query, using `async let` makes this concurrent.

Sources/App/Controllers/API/API+PackageController+GetRoute+Model.swift
```diff
@@ -370,32 +355,11 @@ extension API.PackageController.GetRoute.Model {
...
-            packageName: String,
-            originalOwner: String,
-            originalOwnerName: String,
-            originalRepo: String,
-            originalPackageName: String
-        )
+        case fromSPI(originalOwner: String,
+                     originalOwnerName: String,
+                     originalRepo: String,
+                     originalPackageName: String)
```
finestructure, line 361: Note that the `packageName: String` field was actually unused.

Tests/AppTests/API+PackageController+GetRoute+ModelTests.swift
```diff
@@ -119,84 +116,29 @@ class API_PackageController_GetRoute_ModelTests: SnapshotTestCase {
...
-                                                                       swiftVersionBuildInfo: nil,
-                                                                       platformBuildInfo: nil,
-                                                                       weightedKeywords: [],
-                                                                       swift6Readiness: nil,
-                                                                       forkedFromResult: forkedFromResult
-                                                                      ))
-
-        // validate
-        XCTAssertEqual(model.forkedFromInfo, API.PackageController.GetRoute.Model.ForkedFromInfo.fromGitHub(url: "https://github.com/example/repo.git"))
-    }
```
finestructure, line 152: This test is now pretty much superfluous - we're passing through the simple `ForkedFromInfo` enum and there's no point testing the GH case.

Tests/AppTests/API+PackageController+GetRoute+ModelTests.swift
```diff
@@ -119,84 +116,29 @@ class API_PackageController_GetRoute_ModelTests: SnapshotTestCase {
...
-            originalOwner: "original",
-            originalOwnerName: "OriginalOwner",
-            originalRepo: "original",
-            originalPackageName: "OriginalPkg"
-        )
-        XCTAssertEqual(model.forkedFromInfo, info)
+        XCTAssertEqual(forkedFrom, .fromSPI(originalOwner: "original",
+                                            originalOwnerName: "OriginalOwner",
+                                            originalRepo: "original",
+                                            originalPackageName: "OriginalPkg"))
```
finestructure, line 141: The part that we _do_ want to test is the `ForkedFromInfo.query` and so I've converted `test_init_forked_from_spi` into a test for that. It also gets simpler, because we don't need as much set up just to test that query.

supersonicbyte (approved):

PR #3396, opened 2024-09-18
Disable workflow concurrency on branch basis

supersonicbyte:
Disabling concurrency for workflows on a branch can be helpful to reduce redundant jobs executions. For example:
• I made a PR and it triggers the CI workflows
• I notice that I have a typo in the tests and I that it will cause the tests to fail
• I commit the fix
• Workflows for both commits will run concurrently

This PR disables workflow concurrency on branch basis and the previous scenario would be:
• I made a PR and it triggers the CI workflows
• I notice that I have a typo in the tests and I that it will cause the tests to fail
• I commit the fix
• The workflow of the most recent commit will cancel any in progress or pending workflows from the previous commit

Results of this PR is that we don't waste build time on non-latest commits.

More about workflow concurrency can be found [here](https://docs.github.com/en/enterprise-cloud@latest/actions/writing-workflows/choosing-what-your-workflow-does/control-the-concurrency-of-workflows-and-jobs#example-using-concurrency-and-the-default-behavior).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,6 +20,10 @@ on:
   schedule:
     - cron: "0 8 * * *"
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
   test-linux:
     name: Test
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -18,6 +18,10 @@ on:
     workflow_dispatch:
     push:
 
+concurrency:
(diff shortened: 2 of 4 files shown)


supersonicbyte:
I made some commits to test does it work, and it seems it works just fine.[ Here](https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/actions/runs/10924617672).

daveverwer (approved):
I do this all the time when I push a branch before running snapshots! 😬

I'd love to see this not waste quite so many resources. 👍 from me.

finestructure (approved):

Issue #3384, opened 2024-09-11
SPI Manage Packages Portal and Authentication


rahafjrw:
## Goal

This project provides a base for more features, reporting, and control for package authors and maintainers.

## Scope

To keep the scope of each part of this project under control, this issue is not concerned with adding any additional author-related features. This issue only creates the pages needed for authentication and a single authenticated page that will have no content other than links to logging out and to account deletion.

The entire authentication feature will be gated to the staging server even when this issue is complete and only go live to production once additional features are implemented.

The scope for the this feature going forward is much more significant than described here. See the Future Enhancements section for more detail.

## Authentication system

This authentication system includes:

* A “Portal” link from the site navigation bar that goes either to the login page if the user is not authenticated or to the package portal page if they are.
* Login page hosted at `/login/` containing email and password fields and a link to the password reset page.
* Password reset page hosted at `/reset-password/` with an email field and a link to the login page.
* Logout page hosted at `/logout/`.
* Delete account page hosted at `/portal/delete-account/` functionality that removes all user data and PII for a user.
* A package portal page hosted at `/portal/` that contains links to the logout and delete account pages.

## Implementation

We will use [AWS Cognito](https://aws.amazon.com/cognito/) to implement the authentication portion of this feature. It takes care of password policies, email address confirmation, password reset emails, and storage of user email addresses. It can also be extended to also take care of [fraud detection](https://aws.amazon.com/blogs/machine-learning/prevent-fake-account-sign-ups-in-real-time-with-ai-using-amazon-fraud-detector/) later, if we experience problems with that.

The user will not leave the swiftpackageindex.com domain during authentication. All communication with Cognito is through their API and the [Soto Cognito Authentication](https://swiftpackageindex.com/vapor-community/soto-cognito-authentication) package.

Password policy and confirmation/reset emails can be configured inside the Cognito control panel and we can send all email from swiftpackageindex.com. End users should not be aware that Cognito is being used unless they read the privacy policy.

We can also customise the information that Cognito stores and will start with it storing nothing but email address and (encrypted) password.

## Protection of PII

Any and all user information or personally identifiable information (PII) will be stored in a separate schema in the current database.

## Privacy and cookies

Before this feature goes live to the public, we will also update our privacy policy to include information about the data we store, where we store it, cookies we save, and any other privacy related changes.

Any cookies will be from swiftpackageindex.com, not from a third-party site.

## Future Enhancements

[...]

PR #3397, opened 2024-09-18
Roll back to Soto 6.x until the Soto Cognito libraries support 7.x

daveverwer:
Putting this in as a PR now, ahead of the work that @rahafjrw is doing on #3384. Mainly because there may be some discussion about rolling back the Soto change which I’d like to get out of the way before the first of the auth PRs land.

This rolls Soto back to 6.x due to the dependency in https://swiftpackageindex.com/soto-project/soto-cognito-authentication-kit 4.x. It undoes the change made in https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/pull/3207/files#diff-5d2ca8674645a1e92a14b68d5cc308c427c2d58120392a0df74d6a4f5fc97440.

There is a beta version of soto-cognito-authentication-kit in development, but it is not yet released. I chatted with @adam-fowler this afternoon and apparently the hold up on that beta is the dependency on the beta version of jwt-kit which is currently in RC.

It’s unlikely all of these will shake out before the authentication feature is finalised, but they also should not hold us on 6.x for very long since all of the dependencies have a RC status at the moment.

### DIFF
--- a/Package.resolved
+++ b/Package.resolved
@@ -1,5 +1,5 @@
 {
-  "originHash" : "be2441293ba73b4860ddd36ee48561798ff51a3b9497b64983ee1c24a12af89d",
+  "originHash" : "6ae7335a682b3be27994b6ef856d2a166455a5b43a582f9c17da2f694763ca42",
   "pins" : [
     {
       "identity" : "async-http-client",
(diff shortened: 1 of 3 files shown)


adam-fowler:
If you ever feel the need there is a v5 beta of soto-cognito-authentication out now. https://github.com/vapor-community/soto-cognito-authentication/releases/tag/5.0.0-beta.1. But it'd need you to upgrade to the jwt-kit v5 release candidate

finestructure (approved):

PR #3399, opened 2024-09-20
Explore using swift dependencies

finestructure:
As discussed, we'll want to start switching over, provided we find a solution for the problem of adding `@Dependency` properties to `Codable` types.

I've [opened a discussion](https://github.com/pointfreeco/swift-dependencies/discussions/283) on the swift-dependencies repo to see if there's better solution that the three that I can think of.


finestructure:
There was a better solution and also a lurking problem due to escaping closure but both are fixed now, thanks to the help by the Pointfree folks.

daveverwer (approved):

PR #3417, opened 2024-10-02
Documentation Changes

njblain1:
Added content to README file which includes suggestions for components that could be added. Added comments for code readability in Tests/AppTests/AlertingTests.swift, Tests/AppTests/LicenseTests.swift, and Tests/AppTests/PackageContributorsTests.swift.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,8 +1,10 @@
 ![The Swift Package Index](.readme-images/swift-package-index.png)
 
+## Introduction
+
 **The Swift Package Index helps you make better decisions about the dependencies you use in your apps**.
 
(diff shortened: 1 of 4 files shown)


daveverwer:
Hi @njblain1 and apologies for the delayed response to this. 

We decided that the README changes don't really fit what we want to add there. Also, while comments are always appreciated, we feel like these comments are more focused on the "what" rather than the "why", which is where comments become more valuable.

If you'd like to discuss future contributions, please open a discussion before a PR so we can chat through the changes. Thanks so much for your efforts, but we'll close this for now.

Thanks

Issue #2957, opened 2024-03-20
Custom package collections


finestructure:
Custom package collections are a way to create a collection of packages that is searchable on SPI and adds a badge to the package page of participating packages.

The canonical example are [SSWG incubated packages](https://www.swift.org/sswg/incubation-process.html), which would be grouped by the incubation stage and show the appropriate badge on the package page, e.g. "SSWG Graduated".

I presented our design to the SSWG recently and it was accepted.

![CleanShot 2024-03-20 at 09 50 03@2x](https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/assets/65520/08f9ea68-5e8f-40ca-aaeb-100d05d16ae6)

https://www.icloud.com/freeform/06f2y-34zSpXMHgo0heqXBPtQ#Custom_package_collections

Implementing this is part of our contribution to the SSWG's 2024 goals.

finestructure:
This [has been shipped](https://swiftpackageindex.com/blog/introducing-custom-package-collections)

PR #3429, opened 2024-10-13
Custom package collections

finestructure:
This adds scheme and reconciliation support for custom package collections (#2957) and reads the SSWG package's sandbox, incubating, and graduated package lists (from a temporary location).

Does not include any visual representation or use of the data in search yet.

### DIFF
--- a/Sources/App/Core/Dependencies/PackageListRepositoryClient.swift
+++ b/Sources/App/Core/Dependencies/PackageListRepositoryClient.swift
@@ -25,6 +25,26 @@ struct PackageListRepositoryClient {
 }
 
 
+extension PackageListRepositoryClient: DependencyKey {
+    static var liveValue: PackageListRepositoryClient {
+        .init(
+            fetchCustomCollection: { client, url in
+                try await client
+                    .get(URI(string: url.absoluteString))
+                    .content
+                    .decode([URL].self, using: JSONDecoder())
+            },
+            fetchCustomCollections: { client in
+                try await client
+                    .get(Constants.customCollectionsUri)
+                    .content
+                    .decode([CustomCollection.DTO].self, using: JSONDecoder())
+            }
+        )
+    }
+}
+
+
 extension PackageListRepositoryClient: Sendable, TestDependencyKey {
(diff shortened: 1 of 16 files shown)


Sources/App/Core/Dependencies/PackageListRepositoryClient.swift
```diff
@@ -0,0 +1,59 @@
...
+import Dependencies
+import DependenciesMacros
+import Vapor
+
+
+@DependencyClient
+struct PackageListRepositoryClient {
+    var fetchCustomCollection: @Sendable (_ client: Client, _ url: URL) async throws -> [URL]
+    var fetchCustomCollections: @Sendable (_ client: Client) async throws -> [CustomCollection.DTO]
+    // TODO: move other package list dependencies here
```
finestructure, line 24: I will do this in a follow-up PR right after this one but I wanted to keep the change set focused.

finestructure: Also, this illustrates how we would set up grouped dependencies (like Github, for instance) with the new `Dependencies` mechanism.

Sources/App/Models/Package.swift
```diff
@@ -271,6 +271,10 @@ extension QueryBuilder where Model == Package {
     func filter(by url: URL) -> Self {
         filter(\.$url == url.absoluteString)
     }
+
+    func filter(by urls: some Collection<URL>) -> Self {
+        filter(\.$url ~~ urls.map(\.absoluteString))
```
finestructure, line 276: There is a `SQLKit` based way to make this a case-insensitive filter but it's a bit more complicated. I'll leave that to a follow-up.

finestructure: When doing that, we should also canonicalise the urls from the custom collection. Since they are not added via our normal package list mechanism and validated (canonicalised) in the process, we can't assume they're all as well-formed as they are in the main package list.

finestructure:
Ad hoc deploying the branch for testing.

daveverwer (approved):

PR #3440, opened 2024-10-15
Make Package.filter(by: [URL]) case-insensitive and normalise the input URLs

finestructure:


daveverwer:
Is there any context to this? This seems like quite a core change.

finestructure:
It's the deferred lifting of a limiting I mentioned here in the first custom collections PR: https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/pull/3429#discussion_r1798292545

We shouldn't require maintainers of custom collection lists to have exact case-sensitive matches of the core package list we match against.

daveverwer (approved):

PR #3467, opened 2024-10-31
Add a PROCESSING_BUILD_BACKLOG flag and update the "missing compat info" message

finestructure:
Changes "missing compatibility info" message to

Screenshot 2024-10-31 at 08 16 00

or

Screenshot 2024-10-31 at 08 21 16

depending in a new `PROCESSING_BUILD_BACKLOG` flag.

I used `.br()`, because `.p` has a huge vertical gap here.

Also, I noticed that the link in the existing message text does not carry a link colour.


daveverwer:
I can fix up the styling issues here with the `p` etc… I'll add the commits directly in this PR

daveverwer:
I ended up going with two completely different bits of wording as I think people might get blind to the actual words without a bit of a layout change.

When there's no flag set:

Screenshot 2024-10-31 at 12 34 34@2x

When the flag is set:

Screenshot 2024-10-31 at 12 37 36@2x

daveverwer (approved):
Ready to go if you're happy with what I did @finestructure .

PR #3493, opened 2024-11-13
Add overflow-wrap: break-word to code

finnvoor:
Currently long `code` goes off the end of the page and results in a horizontal scroll bar, as seen here: https://swiftpackageindex.com/finnvoor/SwiftUIFX

Arc - SwiftUIFX – Swift Package Index@2x

This change wraps the code instead, which I think is nicer. Not sure if this is the best place to change it though.

Arc - SwiftUIFX – Swift Package Index@2x

### DIFF
--- a/FrontEnd/styles/base.scss
+++ b/FrontEnd/styles/base.scss
@@ -83,6 +83,7 @@ pre {
 
 code {
     font-size: 14px;
+    overflow-wrap: break-word;
 }
 
 blockquote {


daveverwer:
@cla-bot check

daveverwer (approved):
This is a good change and I just double-checked that it does not affect `pre > code` blocks like code listings, which we definitely do not want to wrap.

There's an argument to scope this additional CSS so it only affects readme files, but I think it's unlikely that we'd ever want the overflowing functionality. We can leave it here as long as it passes all snapshot tests.

Talking of tests, I'll push your branch to the repository now to get the CI running.

daveverwer:
Thank you so much for your contribution @finnvoor. We really appreciate it!

PR #3474, opened 2024-11-04
Adding Mermaid Chart Support

leogdion:

PR #3499, opened 2024-11-14
Add support for Mermaid charts to README files

daveverwer:
Supercedes #3474

Well, what a rabbit hole this turned out to be!

Thank you so much @leogdion for this contribution, and please don’t take the fact that I completely rewrote it as any slight on your code.

Once I realised what you were doing with removing and then re-populating the `pre` element with the Mermaid source, it didn’t seem like the best approach. I changed things around and now we use `render` on `mermaid` to grab SVG data directly and insert _both_ a light and dark mode chart into a new container.

This has several advantages:

1. No need to run JavaScript when switching between light and dark mode. We can do with CSS just like we do for the rest of the site.
2. We get to style the div as we like. You mentioned on our call this afternoon that you wanted to try and centre the chart which I have now done. It also removes it from the unnecessary `pre` styling with a grey background. I think it looks better this way, and it gives more options for future styling.

The disadvantage is that we now need to render every chart twice. I can’t see that this would be a performance problem, but it’s something to watch out for.

Thank you again. This is a great step forward for README files on SPI!

### DIFF
--- a/FrontEnd/scripts/controllers/readme_controller.js
+++ b/FrontEnd/scripts/controllers/readme_controller.js
@@ -16,60 +16,44 @@ import { Controller } from '@hotwired/stimulus'
 import mermaid from 'mermaid'
 
 export class ReadmeController extends Controller {
-    onThemeChange(darkMode) {
-        // If `darkMode` isn't passed, get the current value
-        if (darkMode == undefined) {
-            darkMode = window.matchMedia('(prefers-color-scheme: dark)').matches
-            console.log('actually dark mode is', darkMode)
-        }
-
-        // Get all mermaid diagrams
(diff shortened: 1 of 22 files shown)


leogdion:
Great job! @daveverwer 
Is there a way I can see it in action?

daveverwer:
> Is there a way I can see it in action?

The easiest way would be to run it locally, but here are some screenshots:

Screenshot 2024-11-15 at 09 06 55@2x

and in dark mode

Screenshot 2024-11-15 at 09 06 29@2x

finestructure (approved):

daveverwer:
@leogdion It's also now live: https://swiftpackageindex.com/brightdigit/Sublimation

PR #3509, opened 2024-11-19
Custom collections styling

daveverwer:
Here are two options for displaying custom collection membership for packages in the SSWG collections:

Screenshot 2024-11-19 at 14 53 04@2x

I’m using the `badge` property of the metadata to create the lozenge and then the name after that. Multiple collections are in a `listPhrase`. For the screenshot, I have manually adjusted the collection name to not include the letters “SSWG”. We’d need to adjust the PackageList JSON to make that permanent.

I thought I’d prefer the separate badge when I set out to design it, but it doesn’t look like a link and it didn’t fit well where I first wanted it to go, under the package title.

I think the list item is the way to go.

Here's an alternative on the lozenge:

Screenshot 2024-11-19 at 15 13 13@2x

Let me know which one the workgroup prefers.


finestructure:
> For the screenshot, I have manually adjusted the collection name to not include the letters “SSWG”. We’d need to adjust the PackageList JSON to make that permanent.

This looks great!

The only issue I see is that if we drop the "SSWG" from the name, we end up with "Graduated package collection" on the collection page:

Screenshot 2024-11-20 at 10 15 48

We can fix that by simply prefixing it with the `badge` there. Or should we replicate the lozenge?

finestructure (approved):

finestructure:
FYI, I'll rebase this branch because the blog post one is based on this branch.

daveverwer:
Was there any feedback from the SSWG on this design? The metadata one would be my preference, did they agree?

finestructure:
They were happy with either, we mainly talked about the process to get the json files hosted tbh (this was on Nov 20).

daveverwer:
I'll wrap this up today then.

daveverwer:
This is ready for a final review and can go live whenever.

Issue #2639, opened 2023-09-25
Package mattpolzin/Sampleable stuck in analysis loop


finestructure:
```
│   [ INFO ] fetch failed: Shell command failed:                                                                                                     
│   command: "git checkout master --quiet"                                                                                                           
│   path:    "/checkouts/github.com-mattpolzin-sampleable"                                                                                           
│   message: "ShellOut encountered an error                                                                                                          
│   Status code: 1                                                                                                                                   
│   Message: "error: pathspec 'master' did not match any file(s) known to git"                                                                       
│   Output: """ [component: analyze]                                                                                                                 
│   [ INFO ] removing directory [component: analyze]                                                                                                 
│   [ INFO ] cloning https://github.com/mattpolzin/Sampleable.git to /checkouts/github.com-mattpolzin-sampleable [component: analyze]                
│   [ CRITICAL ] updatePackages: unusually high error rate: 1/5 = 20.0% [component: analyze]                                                         
│   [ WARNING ] Shell command failed:                                                                                                                
│   command: "git log -n1 --format=tformat:"%H-%ct" master"                                                                                          
│   path:    "/checkouts/github.com-mattpolzin-sampleable"                                                                                           
│   message: "ShellOut encountered an error                                                                                                          
│   Status code: 128                                                                                                                                 
│   Message: "fatal: ambiguous argument 'master': unknown revision or path not in the working tree.                                                  
│   Use '--' to separate paths from revisions, like this:                                                                                            
```

This repeats on every analysis cycle. Since we've recently reduced the analysis batch size, we're running the risk of blocking analysis should more packages get stuck like this: right now, we're effectively only processing 4 packages per batch.

The problem in this case appears to be that the default branch has changed from `master` to `main`. But since we're stuck in analysis failure, we don't advance to ingestion where we'd _detect_ the default branch change.

finestructure:
The real problem is that we let an `AppError.shellCommandFailed` bubble up out of an analysis function. This leads to a problem in `recordError` where we're switching on the errors:

```swift
    switch error {
        case let .analysisError(id, _):
            try await setStatus(id: id, status: .analysisFailed)
        case .envVariableNotSet, .shellCommandFailed:
            break
        case let .genericError(id, _):
            try await setStatus(id: id, status: .ingestionFailed)
```

Since we don't have a package id (`shellCommandFailed` doesn't have one), we can't update the record and therefore the package isn't moved forward in the pipeline.

PR #3549, opened 2024-12-12
Rework error handling

finestructure:
Rationale: https://noteplan.co/n/575ADC0D-8D6C-4ADF-BB25-972717C42C41

To do

- [x] Wait for upstream merge and release of https://github.com/pointfreeco/swift-concurrency-extras/pull/52, then remove dependency pin on fork
- [x] Write up a rationale for this change (based on https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/issues/2639 and the extended `test_ingest_continue_on_error`)
- [x] Let it run on staging for a few days
- [x] Check if we're seeing any weird error handling issues
  - [x] is the error formatting ok?
  - [x] do we see new errors popping up?
  - [x] are some errors not popping up anymore?
  - [x] cause some errors on `dev` to confirm things work as expected

This is best reviewed with the "ignore whitespace" setting to hide all the indentation noise from new namespaces.


Sources/App/Commands/Common.swift
```diff
@@ -17,111 +17,125 @@ import PostgresKit
...
 
 
-/// Update packages (in the `[Result<Joined<Package, Repository>, Error>]` array).
-///
-/// - Parameters:
-///   - client: `Client` object
-///   - database: `Database` object
-///   - results: `Joined<Package, Repository>` results to update
-///   - stage: Processing stage
-func updatePackages(client: Client,
```
finestructure, line 27: `updatePackages` (note the plural) was only called from analysis. Moving this into the `Analyze` namespace (which I'll update to `Analysis` at some point in the future) makes this explicit.

Tests/AppTests/IngestorTests.swift
```diff
@@ -83,12 +83,17 @@ class IngestorTests: AppTestCase {
...
-        XCTAssertEqual(repo.licenseUrl, "license")
+        do {
+            // validate the second package's license is updated
+            let repo = try await Repository.query(on: app.db)
+                .filter(\.$name == "2")
+                .first()
+                .unwrap()
+            XCTAssertEqual(repo.licenseUrl, "license")
+            for pkg in try await Package.query(on: app.db).all() {
+                XCTAssertEqual(pkg.processingStage, .ingestion, "\(pkg.url) must be in ingestion")
```
finestructure, line 94: This is a crucial bit of additional test coverage where we ensure the processing stage is advanced when there are errors.

Tests/AppTests/IngestorTests.swift
```diff
@@ -354,12 +364,12 @@ class IngestorTests: AppTestCase {
 
     func test_ingest_unique_owner_name_violation() async throws {
         // Test error behaviour when two packages resolving to the same owner/name are ingested:
-        //   - don't update package
         //   - don't create repository records
```
finestructure, line 367: This is an important change to how we're dealing with this error scenario. Previously, we did two things:

- avoid creating a duplicate repository record
- avoid updating the related package

We could preserve this behaviour but it actually simplifies things quite a bit to drop the second part of this. And it's the theme of this change: make sure processing is advanced through a stages despite errors.

In order to ensure this change doesn't break things in the following stage, this test now explicitly runs analysis as part of its validation to ensure it doesn't raise any errors (see below).

Tests/AppTests/IngestorTests.swift
```diff
@@ -390,31 +399,51 @@ class IngestorTests: AppTestCase {
...
+                if cmd.description.hasSuffix("package dump-package") {
+                    return .packageDump(name: "foo")
+                }
+                return ""
+            }
+
+            try await Analyze.analyze(client: app.client, database: db, mode: .id(.id0))
+            try await Analyze.analyze(client: app.client, database: db, mode: .id(.id1))
+            try await XCTAssertEqualAsync(try await Package.find(.id0, on: db)?.processingStage, .analysis)
+            try await XCTAssertEqualAsync(try await Package.find(.id1, on: db)?.processingStage, .analysis)
```
finestructure, line 446: This is where we're ensuring analysis can deal with this scenario where one package doesn't have a corresponding repository record.

Sources/App/Core/Dependencies/EnvironmentClient.swift
```diff
@@ -44,6 +44,16 @@ struct EnvironmentClient {
...
+
+    enum FailureMode: String {
+        case fetchMetadataFailed
+        case findOrCreateRepositoryFailed
+        case invalidURL
+        case noRepositoryMetadata
+        case repositorySaveFailed
+        case repositorySaveUniqueViolation
+    }
+    var shouldFail: @Sendable (_ failureMode: FailureMode) -> Bool = { _ in false }
```
finestructure, line 56: I've added this error triggering mechanism, which allows us to configure an environment variable `FAILURE_MODE` as for example

```json
{
    "fetchMetadataFailed": 0.1,
    "findOrCreateRepositoryFailed": 0,
    "invalidURL": 0,
    "noRepositoryMetadata": 0,
    "repositorySaveFailed": 0,
    "repositorySaveUniqueViolation": 0
}
```

which will trigger a `fetchMetadataFailed` error at a rate of 10%.

I'm using this to ensure on `dev` that there's no knock-on effect from errors in subsequent stages. It's also made it easier to ensure locally that the error logs are ok.

We can revert this after testing.

finestructure:
Ok, this is all working as expected:

Screenshot 2024-12-15 at 13 52 19

daveverwer (approved):

finestructure:
Now that we've got 1.3.1 of swift-concurrency-extras I've rebased all the package manifest changes out of the PR.

PR #3565, opened 2024-12-24
Added a link to the issue we're using to track data race safety reporting inconsistencies to the builds FAQ

daveverwer:
Screenshot 2024-12-24 at 10 58 17@2x


finestructure:
Good idea!

finestructure (approved):

Resources/Markdown/docs/builds.md
```diff
@@ -92,6 +92,8 @@ You can [find more details in this issue](https://github.com/SwiftPackageIndex/S
...
+It’s not possible to distinguish between a package that genuinely has no data race safety errors and one where the errors are being suppressed in code. The number we display on the page is the number of data race errors that the compiler reports.
 
 Note that this does not affect package compatibility, as shown in the compatibility matrix. A package can be fully compatible with Swift 6.x without enabling strict concurrency checks, provided it is not running in Swift 6 language mode. For more information on opting into Swift 6 language mode, [read the Swift 6 migration guide for more information](https://www.swift.org/migration/documentation/swift-6-concurrency-migration-guide/completechecking).
+
+**Note:** If you see inconsistencies between the number of data race safety errors reported by the Swift Package Index and the number you see in local testing, we are [tracking a potential issue with false negatives reported by the Swift compiler](https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/issues/3233#issuecomment-2560966121). Please feel free to note packages that may be affected in that issue thread.
```
finestructure, line 99: I kinda wonder if I false positives is the better term here. It depends on which way you're looking at it but I suppose most people would think of the "positive case" as the one where there are errors. When I named the issue I was thinking in terms of positive/negative results and flipped the meaning essentially.

PR #3568, opened 2024-12-29
Issue 3469 dependency transition 12

finestructure:
This removes the obsolete `httpClient` from `AppEnvironment` and moves `mastodonPost` from `EnvironmentClient` to `HTTPClient`.

I've confirmed that posting (and encoding) works as expected via our test account:

Screenshot 2024-12-29 at 18 36 48


daveverwer (approved):

Sources/App/Core/Mastodon.swift
```diff
@@ -18,34 +18,45 @@ import Vapor
...
+    private static let apiPath = "/api/v1/statuses"
     static let postMaxLength = 490  // 500, leaving some buffer for unicode accounting oddities
 
     struct Credentials {
         var accessToken: String
     }
 
-    // NB: _testEncodedURL is a callback that exists purely to be able to regression test the encoded value
-    static func post(client: Client, message: String, _testEncodedURL: (String) -> Void = { _ in }) async throws {
+    static func apiURL(with message: String) throws -> String {
```
finestructure, line 29: This change allowed us to get rid of this ugly hack with the trailing closure that was purely used for testing.

PR #3572, opened 2024-12-30
Issue 3469 dependency transition 13

finestructure:
This is the first step to move the Github related dependencies into a `GithubClient` `DependencyClient`.

This step moves `fetchLicense` and also drops the `client` parameter from its function call. This trickles through to a few functions in `Github` which now have `client`-free overloads. The outdated versions are marked with deprecation warnings. These will all go away once we've fully transition alle three Github dependency functions over to `GithubClient`.

### DIFF
--- a/Sources/App/Commands/Ingestion.swift
+++ b/Sources/App/Commands/Ingestion.swift
@@ -258,12 +258,13 @@ enum Ingestion {
 
     static func fetchMetadata(client: Client, package: Package, owner: String, repository: String) async throws(Github.Error) -> (Github.Metadata, Github.License?, Github.Readme?) {
         @Dependency(\.environment) var environment
+        @Dependency(\.github) var github
         if environment.shouldFail(failureMode: .fetchMetadataFailed) {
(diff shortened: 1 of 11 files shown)


daveverwer (approved):

finestructure:
I've marked this WIP for now as I've encountered a potentially blocking issue for using typed throws with swift-dependencies. If that can't be resolved, we may want to hold off moving Github functions out of AppEnvironment so we don't carry around deprecation warnings and method overloads for a long time.

finestructure:
I've opened a discussion about the issue here: https://github.com/pointfreeco/swift-dependencies/discussions/324

There are a couple of possible workarounds:

- isolate the typed throw closures to a single `@DependencyClient` (possibly requiring one for each - not ideal
- do this without the `@DependencyClient` macro (haven't tried it yet but that _should_ work)

finestructure:
Dropping the `@DependencyClient` macro worked for the typed throws dependency as a workaround.

PR #3575, opened 2025-01-01
Issue 3469 dependency transition 14

finestructure:
This moves `fetchMetadata` to `GithubClient`.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -19,7 +19,7 @@ import PackageDescription
 let package = Package(
     name: "SPI-Server",
     platforms: [
-        .macOS(.v13)
+        .macOS(.v15)
     ],
     products: [
         .executable(name: "Run", targets: ["Run"]),
--- a/Tests/AppTests/GithubTests.swift
+++ b/Tests/AppTests/GithubTests.swift
@@ -199,44 +199,27 @@ class GithubTests: AppTestCase {
         }
     }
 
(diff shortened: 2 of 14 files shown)


Package.swift
```diff
@@ -19,7 +19,7 @@ import PackageDescription
 let package = Package(
     name: "SPI-Server",
     platforms: [
-        .macOS(.v13)
+        .macOS(.v15)
```
finestructure, line 22: macOS 15 is required for typed throw runtime support due to our use of

```
        @Dependency(\.github.fetchMetadata) var fetchMetadata
```

below.

Tests/AppTests/GithubTests.swift
```diff
@@ -118,143 +114,133 @@ class GithubTests: AppTestCase {
...
+                                                   repository: "alamofire")
+                XCTFail("expected error to be thrown")
+            } catch {
+                guard case Github.Error.requestFailed(.badRequest) = error else {
+                    XCTFail("unexpected error: \(error.localizedDescription)")
+                    return
+                }
             }
         }
     }
```
finestructure, line 200: This test is now superfluous. It tested the `parseOwnerName` part of the `static func fetchMetadata(client: Client, packageUrl: String) async throws → Metadata` overload that was actually not used and which I removed.

daveverwer (approved):

PR #3576, opened 2025-01-01
Log cf-ray headers from CloudFlare, if available

heckj:
This adds a replacement for the default route logging middleware from Vapor that adds in the cf-ray id, if any, from cloudflare.

This expands the logging data a bit, but gets cf-ray data exposed in the logs from the server. I'm not 100% confident how the cf-ray header is spelled and passed through, so I set this up as a low-cost experiment. My idea would be to(if this is acceptable) running this in prod for a week or two, do a log dump, and use analytics on the request logs to identify a bit more of normal-use pattern to be able to tune a rate-limiter-like cache to observe, and hopefully ameliorate, excessive bot-load, like what happened on Dec 30th. 

Right now it's dumping out `--nil--` in the logger metadata. More than happy to tune this to be *in* the log message instead of the metadata, or just _not_ log it if the header doesn't come through. I do expect to see this in a server fronted by CloudFlare, but I don't have an easy experimental case to know for sure.


Sources/App/Core/CFRayRouteLoggingMiddleware.swift
```diff
@@ -0,0 +1,17 @@
...
+// Replica of the Vapor RouteLoggingMiddleware that's tweaked to explicitly expose the cf-ray header in the logger metadata for the request.
+public final class CFRayRouteLoggingMiddleware: Middleware {
+    public let logLevel: Logger.Level
+    
+    public init(logLevel: Logger.Level = .info) {
+        self.logLevel = logLevel
+    }
+    
+    public func respond(to request: Request, chainingTo next: Responder) -> EventLoopFuture<Response> {
+        let cfray = request.headers.first(name: "cf-ray") ?? "--nil--"
```
finestructure, line 12: Is there a reason you want to track nils in the logs? Or should we rather just skip them?

```suggestion
        guard let cfray = request.headers.first(name: "cf-ray") else {
            return next.respond(to: request)
        }
```

finestructure: Also, I think it'll be useful to be able to switch this on and off with a simple env variable toggle that only requires a variable change + redeployment without a full new release build.

I'll open a PR on top of yours with suggested changes!

heckj:
yeah, the `--nil--` was only because I wanted to get a explicit reinforcement if I got the spelling of the header correct, and highlight where the output would be while doing the dev work - since none of that is behind CF

finestructure:
Let's get this running on dev and we'll see how it goes.

heckj:
@daveverwer @finestructure - I think it might be useful to have this in place, although now I'm wondering if I shouldn't augment it to expose more detail. My thought it to have the details available from the SPI logs to do post-attack analysis that encompass where the attacks came from, and perhaps their user-agent reporting as well.

I'm not sure what is available through the CloudFlare analysis tooling, so if it's easier to view/analyze from these elements there, then I'll close this out - no use having duplicated content. I'm afraid the analysis side of mitigating the pressure of random crawlers on the site is a bit hidden to me, so this is kind of stab in the dark.

(just rebased this on the latest updates to main)

If this is obviously redundant to what's available, or potentially useful but insufficient to give a view to do useful analysis, saw the word either way. I can close this out as not needed, or bolster it up.

finestructure (approved):

PR #3579, opened 2025-01-02
Add redis cache

finestructure:
Adds a redis service to `app.yml`.

### DIFF
--- a/app.yml
+++ b/app.yml
@@ -74,6 +74,7 @@ services:
     <<: *shared
     depends_on:
       - migrate
+      - redis
     ports:
       - '80:80'
     command: ["serve", "--env", "${ENV}", "--hostname", "0.0.0.0", "--port", "80"]
--- a/app.yml
+++ b/app.yml
@@ -178,6 +178,19 @@ services:
       networks:
         - backend
 
+  redis:
+    # Eventually we'll probably want to use redis/redis-stack-server:7.4.0-v1 here,
+    # which excludes the RedisInsight GUI front-end.
+    image: redis/redis-stack:7.4.0-v1
+    environment:
+      REDIS_ARGS: '--maxmemory 4GB --maxmemory-policy allkeys-lru'
+    ports:
+      - '6379:6379'
+      - '8001:8001'
+    deploy:
+      resources:
+        limits:
+          memory: 4.2GB
 
   migrate:
     image: registry.gitlab.com/finestructure/swiftpackageindex:${VERSION}
--- a/app.yml
+++ b/app.yml
@@ -161,22 +161,22 @@ services:
       - backend
 
   alerting:
-      image: registry.gitlab.com/finestructure/swiftpackageindex:${VERSION}
-      <<: *shared
-      depends_on:
-        - migrate
-      entrypoint: ["/bin/bash"]
-      command: ["-c", "--",
+    image: registry.gitlab.com/finestructure/swiftpackageindex:${VERSION}
+    <<: *shared
+    depends_on:
+      - migrate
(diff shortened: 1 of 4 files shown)


finestructure:
I'll ad hoc deploy this from the branch to see how it goes in dev.

app.yml
```diff
@@ -180,6 +180,19 @@ services:
...
 
+  redis:
+    # Eventually we'll probably want to use redis/redis-stack-server:7.4.0-v1 here,
+    # which excludes the RedisInsight GUI front-end.
+    image: redis/redis-stack:7.4.0-v1
+    environment:
+      REDIS_ARGS: '--maxmemory 4GB --maxmemory-policy allkeys-lru'
+    ports:
+      - '6379:6379'
+      - '8001:8001'
```
finestructure, line 191: We'll need to clarify the situation around the ports before deploying this in prod. I don't think they're open due to the firewall sitting in front of it but it needs double checking.

We should be able to connect to the 8001 port for the insights GUI via an ssh tunnel/port forward. I did that i the past but not in a while - need to dig up the commands again.

finestructure:
Having some trouble getting this to work with service order. I suspect `depends_on` is looking for a health check and the redis container doesn't have one.

I think I know how to fix this but I'll leave it until tomorrow.

app.yml
heckj: If it’s a swarm, there may need to be firewalls opened between the hosts - I don’t think it’s available in prod based on the current network security groups.

heckj:
did some reading when I got settled, and depends_on without a health check only verified start order, as you presumed. I found this example of using a CLI heathcheck for redis:

redis:   
   image: redis:latest    
  ports:      - "6379:6379"    
  healthcheck:      
test: ["CMD", "redis-cli", "ping"]      
interval: 10s      
timeout: 5s      
retries: 3      
start_period: 10s

finestructure:
Ok, this is now working:

```
[ INFO ] Redis set/get result: 2025-01-03 08:43:49 +0000 [component: server]
```

The issue was that `docker swarm deploy` is [stuck on the v3 docker compose file format](https://docs.docker.com/engine/swarm/stack-deploy/) and didn't support the new keys I was trying.

And using just `redis` as a depends_on entry is problematic, because it seems that the default redis container doesn't implement a health check which this relies on. So what happens is that the server doesn't launch if that dependency is set because it doesn't see redis being healthy.

That'd be fixable but I've simply worked around it by putting the temporary test write in a `Task { }` with a delay.

We won't really have to worry about launch order for the real deployment so this is a non-issue.

finestructure:
Ok, connecting to the redis GUI also works via

```
ssh -N -i .ssh/bastion_id_rsa -L 8001:d1:8001 docker-user@bastion
```

The port is only open to connections from bastion.

Screenshot 2025-01-03 at 10 10 49

app.yml
```diff
@@ -161,23 +161,38 @@ services:
...
+    deploy:
+      resources:
+        limits:
+          memory: 0.5GB
+      restart_policy:
           max_attempts: 5
-      networks:
-        - backend
+    networks:
+      - backend
```
finestructure, line 179: The `alerting:` section was indented with 4 spaces rather than 2 spaces like the rest.

app.yml
finestructure: All the swarm hosts are in the same NSG and I'm pretty sure we only need to open 8001 (to Bastion) in case we want to use the Redis GUI.

I've verified that this is working as intended on dev (which is technically also a swarm albeit with just one node).

daveverwer (approved):

PR #3580, opened 2025-01-04
adding redis to local development 

heckj:
I may be "jumping the gun" here- but wanted to explore poking at redis locally to work out the middleware using it.

- this adds a REDIS_HOST environment variable, and references that instead of the string "redis". I've added it to the template for dev and test, and the app.yaml, but I'm not certain about the production ENV variables and what we might want/need to do to set that to "redis" in those locations.
- adds a redis instance in the makefile and commands to spin up and down
- adjusts the code in configure to use the environment variable

With this in place, I was able to spin up the local environment on my machine in docker, with an additional local redis instance exposed on 6379 localhost.


finestructure:
Let's hold off on this, Joe. I'm preparing a [first "real" use of Redis](https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/pull/3582) for our current documentation reference cache. This sets up the infrastructure needed (although it's quite tied to this particular use case for the moment - for example, the nested `Redis` actor should move to the top level once I'm done testing this setup).

finestructure:
Sorry, I just realised I misread the point of this PR, see my comment here: https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/pull/3582#discussion_r1903277660

I think all we need to do is weave the `Environment.get("REDIS_HOST")` into the `Redis` actor where it's using the hard-coded string right now and then we're good.

heckj:
First day I've been able to have any mental capacity to catch back up. Updated this PR to work over the `main` branch, still leaning into something to make redis available locally for dev work. Actor impl works perfectly, for me, but I wasn't entirely sure how you wanted to handle configuring it dynamically - I thought to reach for an environment variable, following the database host access pieces, but noticed that was also a dependency setup, and wasn't sure if dependencies in dependencies were even a thing, or if a straight `processinfo` access point inside the actor setup code would be acceptable.

finestructure:
`@Dependency` works inside a dependency and we should always aim to funnel side effects through them. I'll fix this up and update the `Redis` bit to work with these changes here.

finestructure (approved):

finestructure:
Apologies for the delay Joe, I completely lost track of this one!

daveverwer:
We should also add something to the `LOCAL_DEVELOPMENT_SETUP.md` file that explains in what circumstances and which bits of the code you'd need a Redis server and how to use the make commands.

PR #3582, opened 2025-01-05
Store current reference cache in redis

finestructure:
This starts using Redis in place of the per node current reference cache for doc links.

I've ad hoc deployed this revision to dev and triggered the code path where it writes to Redis:

Screenshot 2025-01-05 at 11 00 03

I'll update the PR now to replace the in-memory cache.

Visit https://staging.swiftpackageindex.com/SwiftPackageIndex/SemanticVersion/~/documentation/semanticversion to hit the cache code path.


Sources/App/Core/Dependencies/RedisClient.swift
```diff
@@ -0,0 +1,127 @@
...
+    private init() async throws {
+        let connection = RedisConnection.make(
+            configuration: try .init(hostname: Redis.hostname),
+            boundEventLoop: NIOSingletons.posixEventLoopGroup.any()
+        )
+        self.client = try await connection.get()
+    }
+
+    // This hostname has to match the redir service name in app.yml.
+    static let hostname = "redis"
```
finestructure, line 105: We could make this configurable but there's not much point as it's not really dependent on anything, just a static name defined in `app.yml`.

Sources/App/Core/Dependencies/RedisClient.swift
```diff
@@ -0,0 +1,127 @@
...
+        }
+    }
+
+    private init() async throws {
+        let connection = RedisConnection.make(
+            configuration: try .init(hostname: Redis.hostname),
+            boundEventLoop: NIOSingletons.posixEventLoopGroup.any()
+        )
+        self.client = try await connection.get()
+    }
```
finestructure, line 102: Please give this whole `actor Redis` a thorough review, in particular how it's setting up the `shared` singleton with retries and the `async throws` initialiser that's using the `@preconcurrency import`ed `RedisConnection`.

In particular, `@preconcurrency` is required, because `RedisConnection` isn't `Sendable`. I _think_ this is correctly set up but good to have a proper review of this.

Unfortunately, there's not much I can see us do around this in terms of testing other than verifying that writing to and reading from Redis works on `dev` (which it does).

Sources/App/Core/Dependencies/CurrentReferenceCacheClient.swift
```diff
@@ -0,0 +1,86 @@
...
+extension CurrentReferenceCacheClient: DependencyKey {
+    static let timeToLive: Duration = .seconds(5*60)
+
+    static var liveValue: CurrentReferenceCacheClient {
+        .init(
+            set: { owner, repository, reference async in
+                @Dependency(\.redis) var redis
+                await redis.set(key: getKey(owner: owner, repository: repository),
+                                value: reference,
+                                expiresIn: timeToLive)
```
finestructure, line 35: @heckj , this is how you'd interface with Redis once this lands. We wouldn't be using `app.redis` at all but rather the private `Redis` actor via the `RedisClient` dependency, replacing your PR #3580.

The only thing this doesn't cover is setup for local use of Redis, i.e. during development. It should be easy to add though, either by making the hostname configurable or just setting a `redis` alias in `/etc/hosts` on the local machine. The latter is certainly easiest and fastest to get going.

Sources/App/Core/Dependencies/RedisClient.swift
```diff
@@ -0,0 +1,127 @@
...
+import NIOCore
+@preconcurrency import RediStack
+import Dependencies
+import DependenciesMacros
+
+
+@DependencyClient
+struct RedisClient {
+    var set: @Sendable (_ key: String, _ value: String?, Duration?) async -> Void
+    var get: @Sendable (_ key: String) async -> String?
```
finestructure, line 24: I've opted for the simplest interface, just taking `String` keys and values for starters. We can add overloads that take `T: Codable` which funnel strings through these methods if needed.

Sources/App/Core/Dependencies/CurrentReferenceCacheClient.swift
finestructure: Correction, I just realised I misread your PR #3580. I was focused on the part

```diff
-        app.redis.configuration = try RedisConfiguration(hostname: "redis")
+        if let redisHost = Environment.get("REDIS_HOST") {
+            app.redis.configuration = try RedisConfiguration(hostname: redisHost)
+        } else {
+            app.logger.warning("REDIS_HOST not set, caching disabled.")
+        }
```

but really the point of it is to add a bringup for a local Redis and the hostname config.

daveverwer (approved):

Sources/App/Core/Dependencies/RedisClient.swift
heckj: This was the only bit that I thought we might want to tweak and make into an environment variable, primarily so that a purely local development setup could reference LOCALHOST:6379 and had redis running in a container.

finestructure: I've simply set `127.0.0.1  redis` in my `/etc/hosts` locally but of course making in configurable is better.

PR #3610, opened 2025-01-15
Added an `uptime-check` page that the automated uptime checkers can use

daveverwer:
Merge after #3609

Adds a simple uptime check route that exercises the database that we should monitor with the Azure and StatusPage checkers.

I'll set this page to cache this for 5 minutes which will prevent it being abused, but it will be a much more accurate test than the home page, which is cached for many hours.

### DIFF
--- a/Sources/App/Controllers/UptimeCheckController.swift
+++ b/Sources/App/Controllers/UptimeCheckController.swift
@@ -1,0 +1,30 @@
+// Copyright Dave Verwer, Sven A. Schmidt, and other contributors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import Fluent
+import Plot
+import Vapor
+
+enum UptimeCheckController {
+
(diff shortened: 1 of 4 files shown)


Sources/App/Controllers/UptimeCheckController.swift
```diff
@@ -0,0 +1,30 @@
...
+
+    @Sendable
+    static func show(req: Request) async throws -> String {
+        // A package page query is a good test of whether the site is healthy.
+        let (_, _) = try await API.PackageController.GetRoute
+            .query(on: req.db,
+                   owner: "SwiftPackageIndex",
+                   repository: "SemanticVersion")
+        return "Success"
+    }
```
finestructure, line 29: Our smoke test is already exercising the db via package page and search requests. I'm wondering why we'd need this extra route?

If we do this, I should call it `/healthcheck`.

finestructure:
FWIW, we know that the CF isn't preventing the uptime checker from alerting. I don't think we need a special route with special caching.

Sources/App/Controllers/UptimeCheckController.swift
daveverwer: Also, and you know this already, the smoke test only happenes at deploy time, the uptime check is a constant monitor.

finestructure: Not quite, the smoke tests also run every four hours. As to the Azure probes, they can probe a package page in addition if we feel that's important. I'm simply asking it wouldn't be simpler and more realistic to do just that.

I'll tick the PR either way, it was just a question if you've considered this.

finestructure (approved):

Sources/App/Controllers/UptimeCheckController.swift
finestructure: I'd meant to tick the PR earlier but forgot after commenting.

PR #3615, opened 2025-01-17
follow-on to explain redis in the LOCAL_DEVELOPMENT_SETUP documentation

heckj:
follow-on to #3580 - explaining the bits of adding redis.

Notably - that it's NOT required to run the site locally, but available if you want to set that up or work on the relevant code paths that use it.


daveverwer (approved):
Thanks Joe!

PR #3627, opened 2025-01-23
Added package collections to the keyword result pages

### LINKED ISSUE #1245: Add package collections to the keyword result pages
For example https://swiftpackageindex.com/keywords/networking

The collection itself should be available at `https://swiftpackageindex.com/keywords/networking/collection.json` and the visuals for it should be similar in design to how we present author collections, for now.

Screenshot 2021-09-06 at 17 48 47@2x

luis-gustavo:
# Issue 

Closes #1245 

# Description

This pull request adds package collections to the keyword result pages.

# Screenshots

* Keyword Page

* Collection JSON
[collection.json](https://github.com/user-attachments/files/18520814/collection.json)


Sources/App/Core/PackageCollection+VersionResult.swift
```diff
@@ -55,6 +55,8 @@ extension PackageCollection.VersionResult {
         switch filter {
             case let .author(owner):
                 query.filter(Repository.self, \Repository.$owner, .custom("ilike"), owner)
+            case let .keyword(keyword):
+                query.filter(Repository.self, \Repository.$keywords, .custom("@>"), [keyword])
```
daveverwer, line 59: I tested this and it's a case sensitive match of a full keyword, which seems like the correct approach to me 👍

daveverwer (approved):
This is really great @luis-gustavo, thank you so much!

I tested it all locally with signing enabled, and it all worked great.

daveverwer:
Thinking about this a bit more, there is a situation that we will need to deal with.

With author collections, the maxiumum number of packages per collection is low but with keywords, it can be really high. For example, the `swift` keyword has 3.7k packages associated with it. Generating a collection this large is not practical.

We could either disable collections for very large keywords or limit the number of packages in a keyword collection, maybe ordered by package score. In either case, we should be careful to explain the limitation in the UI before people copy the collection URL.

daveverwer:
I just ran this locally for `swift` and it took almost 5 minutes and generated a 27MB file 😬

daveverwer:
Looking at the data, there are only 5 keywords with that have more than 300 packages.

Generating a collection for 300 packages takes less than 5 seconds and generates a file that’s around 2MB. That seems reasonable to me, given that we aggressively cache all package collection generation with Cloudflare.

So, let’s limit package collection generation to a maximum of 300 packages and clip anything over that. We should also add a note to any keyword page (and author page, for when we have authors that reach the limit) explaining the limit. We should only add the note when there is more than 300 items in the collection so as not to add UI complexity where it’s not needed.

finestructure:
That still feels like a lot and a long time for a request.

More importantly though, have we tested how Xcode behaves with a package collection of that size? It shouldn't be a problem but sometimes things scale in weird ways and it's worth checking first.

finestructure:
Also, I assume it was 5 seconds on your machine? We should test it on dev, too, because it almost certainly is going to be slower.

daveverwer:
> We can tweak the number once we see performance on production hardware, but 300 seems like a good starting number.

I said this:

> We can tweak the number once we see performance on production hardware, but 300 seems like a good starting number.

luis-gustavo:
Thanks for the feedback @daveverwer and @finestructure. Following what Dave said, I have added a 300 limit to the package collection.

This is how the UI is looking right now:



Please lmk if that works and your feedback will be highly appreciated!

luis-gustavo:
@daveverwer

> So, let’s limit package collection generation to a maximum of 300 packages and clip anything over that. We should also add a note to any keyword page (and author page, for when we have authors that reach the limit) explaining the limit. We should only add the note when there is more than 300 items in the collection so as not to add UI complexity where it’s not needed.

I initially planned to display the note only when a collection exceeds 300 items. However, since we fetch results in pages (typically 20 items) and query 20 extra to check for a next page, there’s no straightforward way to determine if a specific keyword has more than 300 packages. Querying all results just to check this doesn’t seem efficient. I’ll leave the decision to you on whether this is worth implementing.

daveverwer:
> I initially planned to display the note only when a collection exceeds 300 items. However, since we fetch results in pages (typically 20 items) and query 20 extra to check for a next page, there’s no straightforward way to determine if a specific keyword has more than 300 packages.

You're right that this would be slow. However, there is a relatively performant way to get this for keyword pages, but not author pages. In the `weighted_keywords` table (which should really be named `tallied_keywords`, or just `keywords`) each keyword has a count, so we could do a quick query for one record to get that.

There is no equivalent for authors, so we should just leave author collections alone with no note and no trimming as they are far away from needing this at the moment. The max packages one author has is currently 79.

luis-gustavo:
Hi @daveverwer, I’ve implemented the changes as you suggested. I’m now using the weighted_keywords table to fetch the package count, and the disclaimer only appears when the number of packages exceeds 300, as you mentioned. Could you please review it again? Thanks!

Sources/App/Views/Keyword/KeywordShow+View.swift
```diff
@@ -66,8 +79,16 @@ enum KeywordShow {
                 )
             )
         }
-    }
 
+        private func showPackageLimitDisclaimer() -> Node<HTML.BodyContext> {
```
daveverwer, line 83: Nit: Maybe `packageCollectionLimitNote`? We tend not to name these with `show`

Sources/App/Views/Keyword/KeywordShow+View.swift
```diff
@@ -66,8 +79,16 @@ enum KeywordShow {
...
+            guard model.totalPackageCount >= Constants.maxKeywordPackageCollectionCount else { return .empty }
```
daveverwer, line 84: I think this should be a simple greater than rather than greater than or equal to. We say it will be limited to 300 in the note.

Sources/App/Views/Keyword/KeywordShow+View.swift
```diff
@@ -66,8 +79,16 @@ enum KeywordShow {
...
+            return .p(
+                .h6(
```
daveverwer, line 86: Let's not put this in an `h6`, it can just be a `p` tag.

I'd suggest this text:

```html
<p><strong>Note:</strong> Package collections are limited to a maximum of 300 items and this keyword has more than 300 packages associated with it. This collection includes only the top 300 packages for this keyword sorted by package score.</p>
```

daveverwer (changes requested):

daveverwer:
> Also, I assume it was 5 seconds on your machine? We should test it on dev, too, because it almost certainly is going to be slower.

It's faster than my local machine on staging. I tested the `swift` keyword and the `ios` keyword collections on staging just now and they both took ~2.2 seconds to respond. I also tested both in Xcode and it handled them fine.

Screenshot 2025-01-27 at 14 16 34@2x

The only odd thing was that neither collection had exaclty 300 items in it. The `swift` collection had 294 items:

Screenshot 2025-01-27 at 14 17 29@2x

and the ios one had 287 items:

Screenshot 2025-01-27 at 14 18 17@2x

luis-gustavo:
Hi @daveverwer, thanks again for the feedbacks. I've implemented the changes as you suggested. Hopefully, now it should be good to go.

Sources/App/Views/Keyword/KeywordShow+View.swift
```diff
@@ -66,8 +79,15 @@ enum KeywordShow {
                 )
             )
         }
-    }
 
+        private func packageCollectionLimitNote() -> Node<HTML.BodyContext> {
+            guard model.totalPackageCount > Constants.maxKeywordPackageCollectionCount else { return .empty }
+            return .p(
+                .strong("Note: Package collections are limited to a maximum of \(Constants.maxKeywordPackageCollectionCount) packages. If a keyword has more than \(Constants.maxKeywordPackageCollectionCount) packages, only the top \(Constants.maxKeywordPackageCollectionCount) packages will be included based on their package score."
```
daveverwer, line 86: Apologies, I meant for the strong to be only around the "Note:"

```suggestion
                .strong("Note: "),
                .text("Package collections are limited to a maximum of \(Constants.maxKeywordPackageCollectionCount) packages. If a keyword has more than \(Constants.maxKeywordPackageCollectionCount) packages, only the top \(Constants.maxKeywordPackageCollectionCount) packages will be included based on their package score."
```

luis-gustavo: Sure thing, updated!

daveverwer (approved):
Thanks so much for this, @luis-gustavo! We really appreciate it.

PR #3646, opened 2025-01-30
Issue 3469 dependency transition 26

finestructure:
This completes the transition to `@Dependencies` by moving the logging dependencies to `LoggerClient`.

Todo:

- [x] simplify LoggerClient
- [x] ad hoc deploy and check that logs on dev are with `component` metadata etc as expected


finestructure:
All good.

Screenshot 2025-01-31 at 09 06 54

daveverwer (approved):

PR #3648, opened 2025-01-31
Added latest badge on latest documentation page

### LINKED ISSUE #2461: Display additional version info the documentation page header
As discussed with @daveverwer, the documentation page could help visitors quickly identify if they are looking at the latest version. I was made aware of the banner displayed in the older version pages, but the latest version lacks an equivalent affordance.

**Potential Solutions**

The simplest solution would be to include said info between parentheses in front of the version number in the header (ex. `0.4.3 (latest)` or similar.

Something a bit fancier would be a label with a separate visual styling in front of the version handle.

**Further Considerations**

This same pattern could be used to add information about the latest pre-release version or default branches.

luis-gustavo:
# Issue 

Closes #2461 

# Description

This pull request adds the latest badge on the documentation page.

# Screenshots

* With Badge

* Without Badge


daveverwer (approved):
Thanks so much for this @luis-gustavo!

I made a few small changes to this this morning. First, the LATEST badge was baseline centred which makes it look off-centre when the background is added.

Screenshot 2025-02-03 at 11 03 38@2x

I also changed the colour of the background a bit for contrast. The end result now looks like this:

Screenshot 2025-02-03 at 15 09 54@2x

I also removed a debug line that had been accidentally left in.

It’s good to go. Thanks so much for your contribution!

PR #3666, opened 2025-02-17
Stop Google attempting to index the API

daveverwer:
We were no longer using the AI bot rules so I removed them to keep it clear that we are not dealing with that in robots.


Public/robots.txt
```diff
@@ -1,25 +1,4 @@
...
-Disallow: /
-
-User-agent: Google-Extended
-Disallow: /
-
-User-agent: Cohere-ai
-Disallow: /
-
-User-agent: Seekr
-Disallow: /
```
finestructure, line 25: Mmm hold on, we _don't_ want the AI bots crawling the site, do we?

daveverwer: I wrote this in the comment:

> We were no longer using the AI bot rules so I removed them to keep it clear that we are not dealing with that in robots.

These rules are not what is blocking the bots, and this is not a comprehensive list, so it's misleading for us to have these rules here. They are not being used.

finestructure: Maybe we should indicate that, but I think it's also a good place to indicate _intent_. `robots.txt` is the only place where you can communicate you objection to AI crawling quite explicitly and if it doesn't do any harm I'd be in favour of doing so.

finestructure (approved):

PR #3667, opened 2025-02-17
Simplify LoggerClient implementation

finestructure:
The `prepareDependencies` mechanism that's built into swift-dependencies actually does pretty much what I baked into `LoggerClient`.

To do:

- [ ] ad hoc deploy to dev
- [ ] check logs are ok

PR #3672, opened 2025-02-18
Simplify LoggerClient implementation

finestructure:
For some reason I can't reopen #3667 . Reproducing it here:

The `prepareDependencies` mechanism that's built into swift-dependencies actually does pretty much what I baked into `LoggerClient`.

To do:

- [ ] ad hoc deploy to dev
- [ ] check logs are ok

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,9 +27,9 @@ concurrency:
 jobs:
   test-linux:
     name: Test
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04-arm
     container:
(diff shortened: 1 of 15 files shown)


finestructure:
May have to abandon this for now. For some reason this works locally, on both macOS and Linux (arm), but fails in CI, even if I run it on an arm GH runner. I can't think of anything that's different. It _did_ fail locally at one point but not anymore. Might try and reset tomorrow.

The change isn't critical but it's a nice simplification that'll get rid of cruft and ensure the logger is set up exactly once. So it's worth having but not at any cost.

daveverwer (approved):

finestructure:
Closing in favour of #3729

PR #3718, opened 2025-03-10
Issue 3655 swift testing cleanup

finestructure:
- Drop VaporToOpenAPI
- Remove AppTestCase (moving relevant bits to TestSupport and Database+ext


Package.swift
```diff
@@ -36,7 +36,6 @@ let package = Package(
         .package(url: "https://github.com/SwiftPackageIndex/SemanticVersion.git", from: "0.3.0"),
         .package(url: "https://github.com/SwiftPackageIndex/ShellOut.git", from: "3.1.4"),
         .package(url: "https://github.com/swiftlang/swift-package-manager.git", branch: "release/5.10"),
-        .package(url: "https://github.com/dankinsoid/VaporToOpenAPI.git", from: "4.4.4"),
```
finestructure, line 39: VaporToOpenAPI is not thread-safe. It causes crashes when running tests in parallel. Rather than create `openapi.json` dynamically on the fly from the routes, we should switch to using the [swift-openapi-generator](https://swiftpackageindex.com/apple/swift-openapi-generator).

Since the OpenAPI spec isn't a critical piece I've simply removed it for now.

daveverwer (approved):

PR #3727, opened 2025-03-14
Pull forward test changes

finestructure:
This pull forward changes required to parallelise tests. The point here is to reduce the noise of _that_ PR with things that aren't required but can already be applied ahead of time.


Tests/AppTests/AllTests.swift
```diff
@@ -0,0 +1,121 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import Testing
+import Dependencies
+
+
+@Suite(.dependency(\.date.now, .t0)) struct AllTests { }
```
finestructure, line 19: In order to be able to bring up a number of test dbs at the start of the app tests and tear them down afterwards, all the tests need to belong to a single suite. The easiest way to do this is to nest all existing tests under a new top level suite.

This suite will then later gain the trait

```diff
- @Suite(.dependency(\.date.now, .t0)) struct AllTests { }
+ @Suite(.setupDatabasePool, .dependency(\.date.now, .t0)) struct AllTests { }
```

NB: `AllTests` isn't an ideal name I've just realised, because we have a couple of other test suites that aren't nested. I'll rename this to `AllAppTests` after all is done. It'd be too much rebasing for little gain at this point.

Tests/AppTests/AnalyzeErrorTests.swift
```diff
@@ -21,20 +21,23 @@ import Testing
 import Vapor
 
 
+extension AllTests {
+    @Suite struct AnalyzeErrorTests {
+        let socialPosts = LockIsolated<[String]>([])
+    }
+}
```
finestructure, line 28: Some of the nested `@Suite`s can't be defined in `AllTests.swift`, because they hold properties (which can't be declared in extensions).

Tests/AppTests/GitlabBuilderTests.swift
```diff
@@ -154,56 +154,3 @@ import Vapor
     }
 
 }
-
-
-@Suite 
-
-
-struct LiveGitlabBuilderTests {
```
finestructure, line 162: No change, just pulling it into its own file.

daveverwer (approved):

PR #3729, opened 2025-03-16
Simplify logger client 2

finestructure:
This is actually a required change in order to run tests concurrently successfully.

Trying to pull this change forward with a few fixes that hopefully allow it to work on CI as well now.

Superseedes #3672

To do:

- [x] ad hoc deploy to dev
- [x] check logs are ok

### DIFF
--- a/Sources/App/Commands/Alerting.swift
+++ b/Sources/App/Commands/Alerting.swift
@@ -42,8 +42,10 @@ enum Alerting {
         }
 
         func run(using context: CommandContext, signature: Signature) async throws {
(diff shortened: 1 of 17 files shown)


daveverwer (approved):

finestructure:
Server component logs are updating but none of the other jobs are after ad hoc deployment:

CleanShot 2025-03-16 at 18 03 06@2x

finestructure:
Fixed

CleanShot 2025-03-16 at 19 01 57@2x

PR #3738, opened 2025-03-19
Prometheus dependency

finestructure:
Another instance where we were using global state in tests. They started cross-talking when run in parallel.

### DIFF
--- a/Tests/AppTests/MetricsTests.swift
+++ b/Tests/AppTests/MetricsTests.swift
@@ -164,7 +164,6 @@ extension AllTests.MetricsTests {
 
                 // validation
                 #expect((AppMetrics.buildTriggerDurationSeconds?.get()) ?? 0 > 0)
-                print(AppMetrics.buildTriggerDurationSeconds!.get())
             }
         }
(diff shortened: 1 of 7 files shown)


Tests/AppTests/MetricsTests.swift
```diff
@@ -61,18 +61,6 @@ extension AllTests.MetricsTests {
...
-            )
-            let initialAddedTag = try #require(
-                AppMetrics.analyzeVersionsAddedCount?.get(.versionLabels(kind: .tag))
-            )
-            let initialDeletedBranch = try #require(
-                AppMetrics.analyzeVersionsDeletedCount?.get(.versionLabels(kind: .branch))
-            )
-            let initialDeletedTag = try #require(
-                AppMetrics.analyzeVersionsDeletedCount?.get(.versionLabels(kind: .tag))
-            )
```
finestructure, line 75: We don't need to baseline the counts anymore because we have a new, 0-based prometheus client on each test run.

daveverwer (approved):

PR #3749, opened 2025-04-06
Parallelise tests

finestructure:
CleanShot 2025-04-06 at 10 46 49@2x

NB that part of that speedup is due to also dropping TSan for now.

- [x] update local dev docs about test db setup

Comparison of the latest run on this branch:

```
Suite AllTests passed after 56.172 seconds
```

vs `main`:

```
Suite AllTests passed after 368.440 seconds
```

Shaved off more than 5mins, 6.6x improvement.


Makefile
```diff
@@ -35,7 +35,7 @@ run:
 
 test: xcbeautify
 	set -o pipefail \
-	&& swift test --disable-automatic-resolution --sanitize=thread --no-parallel \
```
finestructure, line 38: We have to drop TSan again, it keeps triggering when tests are run in parallel. It doesn't seem to be in code directly, it's in NIO. I don't think it's misuse on our end and I can't imagine it's a NIO problem, so perhaps a false positive.

daveverwer (approved):

PR #3750, opened 2025-04-06
Build with Swift 6.1

finestructure:
This just switches the building and running of spi-server to 6.1.

I'm pulling this forward, because it's needed for parallel testing (and also the Xcode 16.3 update).

### DIFF
--- a/Tests/AppTests/AllTests.swift
+++ b/Tests/AppTests/AllTests.swift
@@ -15,12 +15,14 @@
 @testable import App
 
 import Dependencies
+import SnapshotTesting
 import Testing
 
 
 @Suite(
     .dependency(\.date.now, .t0),
-    .dependency(\.metricsSystem, .mock)
+    .dependency(\.metricsSystem, .mock),
+    .snapshots(record: .failed)
 ) struct AllTests { }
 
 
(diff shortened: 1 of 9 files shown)


Tests/AppTests/AllTests.swift
```diff
@@ -15,12 +15,14 @@
...
 import Dependencies
+import SnapshotTesting
 import Testing
 
 
 @Suite(
     .dependency(\.date.now, .t0),
-    .dependency(\.metricsSystem, .mock)
+    .dependency(\.metricsSystem, .mock),
+    .snapshots(record: .failed)
```
finestructure, line 25: This is a great quality of life change: The if a snapshot differs, it will automatically record a new snapshot and then _fail_.

That means that in order to update snapshots, we don't need to edit tests anymore. We just need to check if there are snapshot differences and commit or revert them as needed.

Tests/AppTests/__Snapshots__/AnalyzerTests/dumpPackage_format.linux.json
```diff
@@ -204,5 +209,8 @@
   ],
   "toolsVersion" : {
     "_version" : "5.9.0"
-  }
+  },
+  "traits" : [
+
+  ]
```
finestructure, line 215: Swift 6.1 package dump format changes.

daveverwer (approved):

PR #3788, opened 2025-05-02
Re enable collection signing

### LINKED ISSUE #3755: Server hangs when running with Swift 6.1 image
The server process seems to be hanging when deploying the new **`swift:6.1.0-jammy`** based base image.

Compare this launch of the 2.116.27 docker image (**`swift:6.0.3-jammy`**) with 2.117.0 (**`swift:6.1.0-jammy`**):

Image

In both cases, the process has no configuration and should immediately raise an internal server error. However, the Swift 6.1 based image just hangs.

finestructure:
Fixes #3755


Sources/App/Controllers/API/API+PackageCollectionController.swift
```diff
@@ -60,7 +61,7 @@ extension API {
 }
 
 
-extension PackageCollectionModel.V1.Collection: Vapor.Content {}
+extension PackageCollectionSigning.Model.SignedCollection: @retroactive @unchecked Sendable, Vapor.Content {}
```
finestructure, line 64: We should open a PR upstream to conform `PackageCollectionModel.V1.Collection` to `Sendable` so we can avoid this `@retroactive @unchecked Sendable` here.

daveverwer (approved):

PR #3791, opened 2025-05-07
Add run 16 results

finestructure:
I'm not quite sure what to make of the results. It looks like 6.1 fixes at least some of the false positives and the total number of packages with zero warnings has gone up noticeably. However, the total number of errors is also still going up quite noticeably 🤷‍♂️

CleanShot 2025-05-07 at 09 22 22@2x

Unrelated to the values, the graph seems to have stopped showing the month labels on the x axis. Also, I think we should try and suppress the lighter grey vertical month grid lines. There are a lot of vertical lines and I think the event ones should have more prominence.


daveverwer:
That chart needs some work. I only planned for the end of the year 😬 I'll make a separate PR for that.

daveverwer (approved):

PR #3807, opened 2025-05-22
Compatibility matrix redesign for new platforms

daveverwer:
There are a few options with the matrix redesign. I wanted to combine the header with the compatibility information and have moved the platform and Swift version names inside the cells, coloured with the existing colours:

Screenshot 2025-05-22 at 17 43 55@2x

Unfortunately, it’s not very clear the difference between a pending/unknown state and a failed state for someone who is unfamiliar with the site. We could do something like this:

Screenshot 2025-05-22 at 18 01 10@2x

Alternatively, there is a version in `f2945ee05e91d19fd2d004cee9d32d7bbb4d51ad` that keeps the headers separated, and looks like this:

Screenshot 2025-05-22 at 18 14 11@2x

Unfortunately, that takes up significantly more vertical space, especially when compatibility does not match between versions. While it’s nice and clear, it pushes the README way too far down.

Screenshot 2025-05-22 at 18 13 17@2x

### Other Improvements

This redesign also changes how we label the versions above the matrices. They’re now separated with a “/” instead of a sentence and it copes with extremely long tag names by truncating the _longest_ tag when they don’t fit on one line:

Screenshot 2025-05-22 at 19 38 16@2x

### DIFF
--- a/FrontEnd/main.scss
+++ b/FrontEnd/main.scss
@@ -22,6 +22,7 @@ $mobile-breakpoint: 740px;
 @import 'styles/breadcrumbs';
 @import 'styles/build_logs';
 @import 'styles/build_monitor';
+@import 'styles/build_results';
 @import 'styles/copyable_input';
 @import 'styles/error';
 @import 'styles/github_highlighting';
--- a/FrontEnd/styles/build_results.scss
+++ b/FrontEnd/styles/build_results.scss
@@ -1,0 +1,154 @@
+// Copyright Dave Verwer, Sven A. Schmidt, and other contributors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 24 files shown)


finestructure:
Looks great! Just a tiny nit - has something about the branch icon baselines changed? They seem a little better aligned in the existing version (dark mode version):

CleanShot 2025-05-22 at 23 08 12@2x
CleanShot 2025-05-22 at 23 07 52@2x

finestructure (approved):

daveverwer:
> Looks great! Just a tiny nit - has something about the branch icon baselines changed? They seem a little better aligned in the existing version (dark mode version):
> 
> CleanShot 2025-05-22 at 23 08 12@2x CleanShot 2025-05-22 at 23 07 52@2x

Fixed:

Screenshot 2025-05-23 at 08 51 38@2x

daveverwer:
I went with this version in the end:

Screenshot 2025-05-22 at 17 43 55@2x

The "(Pending)" felt awkward and the information is available immediately and clearly displayed as soon as anyone clicks anywhere in the matrix.

finestructure:
It was from your own screenshot above actually

daveverwer:
> It was from your own screenshot above actually

Ah, that was to demonstrate the layout I went with and didn't have the other fix applied.

PR #3813, opened 2025-05-30
Speed up GH build pipeline

finestructure:
Unfortunately, we can't benefit from this new flag that uses pre-built SwiftSyntax libraries at the moment.

Due to our dependence on SwiftPM's branch release/6.1, we depend on SwiftSyntax release/6.1, which does not have a prebuilt library and we end up building it ourselves, as before 😞

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,4 +1,4 @@
-FROM registry.gitlab.com/finestructure/spi-base:224946a41aedd5106245666d7261a5218ceac0fc
+FROM registry.gitlab.com/finestructure/spi-base:1.2.2
 
 # Install SPM build dependencies
 RUN apt-get update && apt-get install -y curl git make unzip \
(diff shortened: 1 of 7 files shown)


finestructure:
I've asked on the forums if prebuilts might become available for release/6.1. If not, that's another reason to try and drop SwiftPM as a dependency.

https://forums.swift.org/t/preview-swift-syntax-prebuilts-for-macros/80202/21

finestructure:
Actually, this might work. We have good reason to fork SwiftPM already - there are two open PRs we need and they're unlikely to be be merged and cherrypicked to release/6.1 anyway:

- https://github.com/swiftlang/swift-package-manager/pull/8635
- https://github.com/swiftlang/swift-package-manager/pull/8680

Especially the latter we need. We are currently not getting Postgres-NIO packages updates due to that.

So since we need that fork, we might as well change SwiftPM's dependency on SwiftSyntax to a release version that has a prebuilt instead of the release/6.1 branch. For what we actually use from SwiftPM this will make no difference.

finestructure:
Ah, that looks very interesting! I’ve been meaning to look into caching and his seems to be exactly what we need :) Thanks @MahdiBM !

finestructure:
Much better, now QueryPerformaceTests is the bottle neck

CleanShot 2025-06-06 at 13 28 17@2x

daveverwer (approved):

PR #3818, opened 2025-06-05
Alternative intro

finestructure:
I've drafted an alternative intro that might work better?

I've separated out that part of the changes and some other light edits in a separate commit.


daveverwer (approved):
Much better intro 👍 Thanks

PR #3866, opened 2025-09-02
Update builds.md Swift Migration broken link

w-goog:
I wasn't able to find the original page on Archive.org etc. in order to link to the precise step, so I opted to link to the overview page.


daveverwer:
Thanks for this and well spotted, @w-goog! I've emailed you a CLA and we'll get this merged as soon as we have that.

PR #3922, opened 2026-01-01
Remove cache mount

finestructure:
This was still failing (in a different way):

```
ERROR: failed to solve: process "/bin/sh -c cp \"$(swift build --package-path /build -c release --show-bin-path)/Run\" ./" did not complete successfully: exit code: 1
```

https://gitlab.com/finestructure/swiftpackageindex/-/jobs/12584043933

I'm not sure if the cache mount was doing anything useful in our case but we can check that out again later.


finestructure:
That fixed it: https://gitlab.com/finestructure/swiftpackageindex/-/pipelines/2240681935

PR #3924, opened 2026-01-05
Package update

finestructure:


finestructure:
I suspect the build is failing because of the cache. I'll disable the `--disable-automatic-resolution` option temporarily.

PR #3977, opened 2026-03-02
Pin footer to bottom of viewport on short-content pages

DePasqualeOrg:
On pages with little content (e.g. the 404 page), the footer floats in the middle of the viewport instead of staying at the bottom.

This change uses a flexbox sticky footer pattern: `body` becomes a flex column with `min-height: 100vh`, and `main` gets `flex-grow: 1` to fill the remaining space.

## Changes

- `FrontEnd/styles/base.scss`: Add `display: flex`, `flex-direction: column`, and `min-height: 100vh` to `body`
- `FrontEnd/styles/layout.scss`: Add `flex-grow: 1` to `main`

## Before

before

## After

after

### DIFF
--- a/FrontEnd/styles/base.scss
+++ b/FrontEnd/styles/base.scss
@@ -21,6 +21,9 @@
 }
 
 body {
+    display: flex;
+    flex-direction: column;
+    min-height: 100vh;
     font-family:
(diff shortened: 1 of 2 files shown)


daveverwer:
Can you give me permission to edit this PR, please? There should have been a checkbox to allow the maintainers to make edits. I have a fix locally but can't push it to your branch.

Thanks

daveverwer:
I opened it as a new PR instead, no need to change the permissions. It still contains your original commit.

Closing in favour of #3980

PR #3979, opened 2026-03-02
Add support for EUPL-1.2 license

trbernstein:
The European Union Public License has had an updated version 1.2 for a long time now and projects using that license are shown as "Unknown License", while in fact, that license is already well-known to SwitchPackageIndex, as it supports EUPL in version 1.1.


daveverwer:
Thanks so much for this contribution, @trbernstein. We had a couple of PRs open on this subject and so I decided to do a comprehensive update of all licenses from choosealicense.com rather than merge the individual PRs.

Thank you so much for your contribution, though! We really appreciate it. 

Closing in favour of #3983.

PR #4032, opened 2026-04-14
Mitigate gitlab 500s part 6

finestructure:
This reverts the pageSize decrease I pushed yesterday in favour of changing the decompression size limit for the HTTP client.

TBD what the new size should be. I'm not sure if 1MB isn't too low given this also affects documentation.

Alternatively we could _not_ apply this change to the documentation requests but I wonder if it was affected by hitting the `limit` error as well but we just didn't see it.

### DIFF
--- a/Sources/App/Core/Constants.swift
+++ b/Sources/App/Core/Constants.swift
@@ -49,4 +49,8 @@ enum Constants {
 
     // package
     static let maxKeywordPackageCollectionCount = 300
+
+    // HTTP client
+    // 2026-04-14 sas: For details, see: https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/issues/4024#issuecomment-4237684071
+    static let httpDecompressionSizeLimit = 1_000_000
 }
--- a/Sources/App/Core/Dependencies/HTTPClient.swift
+++ b/Sources/App/Core/Dependencies/HTTPClient.swift
@@ -37,14 +37,14 @@ extension HTTPClient: DependencyKey {
         .init(
             get: { url, headers in
                 let req = try Request(url: url, method: .GET, headers: headers)
-                return try await Vapor.HTTPClient.shared.execute(request: req).get()
+                return try await shared.execute(request: req).get()
             },
             post: { url, headers, body in
                 let req = try Request(url: url, method: .POST, headers: headers, body: body.map({.data($0)}))
-                return try await Vapor.HTTPClient.shared.execute(request: req).get()
+                return try await shared.execute(request: req).get()
             },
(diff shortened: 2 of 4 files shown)


Sources/App/Core/Dependencies/HTTPClient.swift
```diff
@@ -37,14 +37,14 @@ extension HTTPClient: DependencyKey {
...
+                return try await shared.execute(request: req).get()
             },
             post: { url, headers, body in
                 let req = try Request(url: url, method: .POST, headers: headers, body: body.map({.data($0)}))
-                return try await Vapor.HTTPClient.shared.execute(request: req).get()
+                return try await shared.execute(request: req).get()
             },
             fetchDocumentation: { url in
-                try await Vapor.HTTPClient.shared.get(url: url.string).get()
+                try await shared.get(url: url.string).get()
```
finestructure: I wonder if we also hit this `limit` problem in our DocC-Proxy when we're forwarding documentation content sometimes.

Sources/App/Core/Constants.swift
```diff
@@ -49,4 +49,8 @@ enum Constants {
 
     // package
     static let maxKeywordPackageCollectionCount = 300
+
+    // HTTP client
+    // 2026-04-14 sas: For details, see: https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/issues/4024#issuecomment-4237684071
+    static let httpDecompressionSizeLimit = 1_000_000
```
finestructure, line 55: I'm unsure about what limit to set here. The default was `.ratio(25)`, which, if my reading [of the source code](https://github.com/apple/swift-nio-extras/blob/5a48717e29f62cb8326d6d42e46b562ca93847a6/Sources/NIOHTTPCompression/HTTPDecompression.swift#L44-L45) is correct, rejects requests that compress really well.

Since this is being used to forward documentation, 1M might be too small. If my understanding is correct, this would effectively be a file size limit for files we can serve from documentation. Or maybe this is a chunk size and NIO buffers to disk? I can see images and especially video exceeding limits here.

Our prod machines have 16GB of RAM, so we definitely have room to bump this up.

daveverwer (approved):

Sources/App/Core/Dependencies/HTTPClient.swift
finestructure: I've reverted the change to `HTTPClient.post` and `HTTPClient.fetchDocumentation` for now, because it's unclear how they would be affected and the pressing issue is fixing the `limit` errors when checking the pending jobs count.

PR #4041, opened 2026-04-21
Switch compatible/incompatible/other/none license detection for known/unknown/none

daveverwer:
As the title says, this removes the old `incompatible`/`compatible` license categorisation and also cleans up the `other` case into `known`, `unknown`, and `none`.

The biggest impact of this change is on search filtering by license. There is a migration which consolidates `incompatible`/`compatible` into `known` in the scoring struct, and also one which converts `other` to `unknown` in both the scoring data and also in the `repositories` table. The `incompatible`/`compatible` migration is not able to be rolled back perfectly, but will roll back into a state that would then be corrected as future analysis cycles happen.

There are UI changes, too, including a new license icon.

## Known license

Known License

## Unknown license

Unknown License

## No license found

No License Found


Tests/AppTests/LicenseTests.swift
```diff
@@ -38,54 +38,33 @@ extension AllTests.LicenseTests {
...
     @Test func shortName() throws {
         #expect(License.mit.shortName == "MIT")
         #expect(License.agpl_3_0.shortName == "AGPL 3.0")
-        #expect(License.other.shortName == "Unknown license")
-        #expect(License.none.shortName == "No license")
+        #expect(License.unknown.shortName == "Unknown license")
+        #expect(License.none.shortName == "No license found")
     }
 
-    @Test func isCompatibleWithAppStore() throws {
```
daveverwer, line 64: This test felt a little redundant now that we’re not testing for a defined set of `compatible`/`incompatible` licenses so I replaced it with a smaller set of tests exercising just two known licenses and the `unknown`/`none` cases.

Tests/AppTests/SearchFilterTests.swift
```diff
@@ -276,21 +259,18 @@ extension AllTests.SearchFilterTests {
         }
     }
 
-    @Test func licenseFilter_other() async throws {
+    @Test func licenseFilter_unknown() async throws {
         try await withSPIApp { app in
-            let filter = try LicenseSearchFilter(expression: .init(operator: .is, value: "other"))
+            let filter = try LicenseSearchFilter(expression: .init(operator: .is, value: "unknown"))
             #expect(filter.key == .license)
-            #expect(filter.predicate == .init(operator: .in,
```
daveverwer, line 283: I found a slightly larger problem with these tests while fixing up this test. I noticed that no matter what value is passed to `bindableValue` in this `#expect`, the test passes based on the number of parameters passed in, not the _value_ of the parameters.

So, for example, this code passes:

```
            #expect(filter.predicate == .init(operator: .in,
                                              bindableValue: .array(["nonsense"]),
                                              displayValue: "unknown"))
```

And this code fails:

```
            #expect(filter.predicate == .init(operator: .in,
                                              bindableValue: .array(["utter", "nonsense"]),
                                              displayValue: "unknown"))
```

This is because the `==` override on `App.SearchFilter.Predicate.BoundValue` is calling `renderGenericSQL`, which when a single value is passed in, renders `$1` for both sides of the equality operator. This does test _something_, but in a very misleading way.

However, the lines below:

```
            // test view representation
            #expect(filter.viewModel.description == "license is unknown")

            // test sql representation
            #expect(app.db.renderSQL(filter.leftHandSide) == #""license""#)
            #expect(app.db.renderSQL(filter.sqlOperator) == "IN")
            #expect(app.db.binds(filter.rightHandSide) == ["unknown"])
```

As I understand it, these lines should fully test what the `predicate` check does, but in a better way. It still covers `displayValue` (via the `description` expectation), `leftHandSide`, and `sqlOperator`.

I believe the best thing to do is to remove this `.predicate` check, which is what I have done here.

It’s also worth noting that we have many other tests throughout the rest of this file, so if we decide that what I did here by removing this test is correct, I’ll open another PR after it’s merged to fix the remainder of the tests.

Sources/App/Core/Score.swift
```diff
@@ -55,12 +55,10 @@ enum Score {
...
         }
 
-        // Is the license open-source and compatible with the App Store?
+        // Is the license a known open-source license?
         switch candidate.licenseKind {
-            case .compatibleWithAppStore:
+            case .known:
                 scoreBreakdown[.license] = 10
-            case .incompatibleWithAppStore:
-                scoreBreakdown[.license] = 3
```
finestructure, line 63: Should we keep the score of 3 for _having_ a license? That's pretty much what the existing code did.

Sources/App/Migrations/084/UpdateScoreDetailsLicenseKind.swift
```diff
@@ -0,0 +1,75 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import Fluent
+import SQLKit
+
+
+struct UpdateScoreDetailsLicenseKind: AsyncMigration {
```
finestructure: Small nit: the _real_ migration is `UpdateRepositoriesLicense` whereas the update to `packages.score_details` is an optimisation to pull forward the updates on a derived field that the analysis job would compute subsequently.

We should probably name the migration after the important step (and could consider dropping the others entirely).

Tests/AppTests/SearchFilterTests.swift
finestructure: Ah, that's bad, I see what went wrong here. I abused the SQL rendering to work around the `any` making `BoundValue` hard to make equatable.

It's still not ideal but I've fixed it (will push the changes in a second) so that we correctly distinguish between `.value()` and `.array()`. Since this is just for test assertion purposes we can get away with it.

The `#expect(filter.predicate == ...` assertion is worth keeping in my opinion, because it shows that we're parsing the input values into proper intermediate types. Otherwise `bindableValue` wouldn't be checked anywhere, I believe.

Sources/App/Core/Score.swift
daveverwer: Every package in the index *should* get this, so I see it more as applying a -10 penalty for having no license at all.

In my opinion, unlicensed code should be a serious hit to the score, not just -3. I'd suggest we keep it as 10.

finestructure (approved):

Sources/App/Migrations/084/UpdateScoreDetailsLicenseKind.swift
daveverwer: I'd be happy to rename the migration, but the changes to score are important as the old data causes a crash when deserialising score data.

daveverwer: I'll call it `UpdateRepositoriesLicenseAndScoreDetails`, does that work?

Sources/App/Core/Score.swift
finestructure: What I mean is, shouldn't `unknown` take the place of `other`, like so:

```swift
        switch candidate.licenseKind {
            case .known:
                scoreBreakdown[.license] = 10
            case .unknown:
                scoreBreakdown[.license] = 3
            case .none:
                break
        }
```

daveverwer: Fixed and added a test case to check it.

daveverwer:
Tested on staging and deploying as 2.123.7

PR #4064, opened 2026-05-25
Reset text color for all paragraphs in an alert box

SimplyDanny:
On https://swiftpackageindex.com/realm/SwiftLint you can find one note box that looks like

Screenshot 2026-05-25 at 13 05 34

where the second and third paragraph should not be in blue.

The change makes sure, that only the title receives the highlight color.

### DIFF
--- a/FrontEnd/styles/_readme.scss
+++ b/FrontEnd/styles/_readme.scss
@@ -110,7 +110,7 @@
         font-weight: 500;
     }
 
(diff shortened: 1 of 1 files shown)


daveverwer:
@cla-bot check

daveverwer (approved):
Thank you so much for your contribution, @SimplyDanny! We really appreciate it.

PR #4083, opened 2026-06-24
Add AWS native functionality

kcieplak:
* Based on the configuration variable AWS_S3_DIRECT_ACCESS read build logs, documentation and readme files directly from S3, rather than the S3 web frontend.
  * Add handling of root urls to map to index.html
* Based on AWS_USE_IAM_ROLE use instance roles to access S3 rather than access key/secret
* Based on ENABLE_PACKAGE_UPLOAD_PRESIGNED_URLS and ENABLE_BUILD_LOGS_PRESIGNED_URLS, create presigned URLs to pass to builders for logs and docs upload.
* Add functionality to the Dockerfile to be able to build spi in debug vs release
* Do not add back cookies in responses, the site does not use cookies.
* Make the gitlab project id for build triggers configurable via GITLAB_PROJECT_ID *


Sources/App/Core/Dependencies/HTTPClient.swift
```diff
@@ -104,6 +112,84 @@ private let globallySharedHTTPClient: Vapor.HTTPClient = {
...
+extension HTTPClient {
+    static func fetchWithIAMAuth(url: URI) async throws -> Response {
+        // Create AWS client with default credential provider
+        let awsClient = AWSClient(
+            credentialProvider: .default,
+            httpClientProvider: .createNew
+        )
+
+        defer {
+            Task {
```
finestructure, line 125: This raises a warning

```
Unstructured throwing task created by 'init(name:priority:operation:)' is not used, which may accidentally ignore errors thrown inside the task
```

We have a construct

```
try await run { ... }
defer: {
   try await client.shutdown()
}
```

(see above in the same file, line 63ff) in order to deal with async defer.

finestructure: Having some issues getting tests to pass with this branch. I'll see if it's on my end and if there are any changes, I'll roll them all up into a PR on top of this PR rather than sprinkle lots of comments on this one here.

daveverwer (approved):

PR #4089, opened 2026-07-01
Review suggestions/fixes

finestructure:
- fixes warnings
- updates and fixes tests

⚠️ This does not include new tests for the additional code paths around pre-signed documentation loading.

I've also disabled the Query Performance Test action, because it cannot run successfully at the moment.

CleanShot 2026-07-01 at 15 13 26@2x


vjdyjp8drj-lgtm (changes requested):
Nice

daveverwer:
I believe this can be closed as the commits here were merged as part of #4115.

PR #4116, opened 2026-07-31
Allow the hiding of compatibility information for the latest version of Swift

daveverwer:
Regular view:

Screenshot 2026-07-31 at 18 06 52

With `HIDE_LATEST_SWIFT_VERSION_BUILD_DATA` set:

Screenshot 2026-07-31 at 18 09 04

Obviously this will hide 6.4 and not 6.3 once #4112 is merged.

### DIFF
--- a/FrontEnd/styles/_matrix.scss
+++ b/FrontEnd/styles/_matrix.scss
@@ -36,6 +36,12 @@
     padding: 20px 0;
     list-style: none;
 
+    &.hide-latest-swift-version {
+        .results .result:first-child {
+            display: none;
+        }
+    }
+
     .version {
         display: flex;
         flex-direction: column;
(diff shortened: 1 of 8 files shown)


app.yml
```diff
@@ -61,6 +61,7 @@ x-shared: &shared
     GITLAB_API_TOKEN: ${GITLAB_API_TOKEN}
     GITLAB_PIPELINE_TOKEN: ${GITLAB_PIPELINE_TOKEN}
     GITLAB_PIPELINE_LIMIT: ${GITLAB_PIPELINE_LIMIT}
+    HIDE_LATEST_SWIFT_VERSION_BUILD_DATA: ${HIDE_LATEST_SWIFT_VERSION_BUILD_DATA}
```
finestructure, line 64: @kcieplak - I believe this env variable in `app.yml` isn't effective anymore and it would need to be replicated in the ops project, wouldn't it?

finestructure (approved):

app.yml
daveverwer: Ah, that's interesting and feels like something which might be easy to forget in future updates. I wonder if we should try to find a way to have these variables somewhere in this project, too?

Issue #3944, opened 2026-02-01
404 on documentation page for main branch


DePasqualeOrg:
I get a 404 error for the [root of my main branch](https://swiftpackageindex.com/depasqualeorg/swift-mcp/main) documentation pages.

However, the latest [build results](https://swiftpackageindex.com/DePasqualeOrg/swift-mcp/builds) show the docs icon for the main branch, indicating that the build succeeded.

I sometimes also get intermittent "An unknown error occurred" pages, which work on subsequent reloads.

finestructure:
Hi @DePasqualeOrg , thanks for the report.

That URL you're using isn't quite correct, it should be:

https://swiftpackageindex.com/DePasqualeOrg/swift-mcp/main/documentation

and that URL worked for me when I tested it just now.

Where did you come across it, is it from somewhere on the site?

PR #4126, opened 2026-08-07
Handle non-AbortError errors in ErrorMiddleware

MsfPablo:
Refs #3944

### Problem

`ErrorMiddleware` only catches `AbortError`s. Any other error thrown from a route escapes the middleware, is converted into a raw 500 by Vapor's internal error handling, and is **never logged** by the application — so intermittent failures leave no trace.

This also explains the "An unknown error occurred" pages reported in #3944: swift-docc-render shows "page not found" for a 404 on a `/data/…json` request, but "An unknown error occurred" for [any other error status](https://github.com/swiftlang/swift-docc-render/blob/main/src/utils/data.js#L97-L105). An unlogged raw 500 from a non-`AbortError` produces exactly that, with nothing in the logs to diagnose it.

(Credit to @DePasqualeOrg, who tracked this down in https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/issues/3944#issuecomment.)

### Change

`ErrorMiddleware` now catches every error, not just `AbortError`. Non-`AbortError`s are treated as `.internalServerError`, logged at `critical` with the underlying error description, and rendered with the usual HTML error page. `AbortError`s with a status below 400 keep propagating, as before.

The user-facing page only shows the generic "500 - Internal Server Error" text — the underlying error description goes to the log, not the response.

### Notes

- This is a visibility/robustness fix; it doesn't itself identify what is intermittently failing, but it makes the failure show up in the logs.
- `swift build` passes locally. I couldn't run the test suite here (the toolchain in my environment can't resolve the `Testing` module for the test targets, which also fails on a clean checkout), so the added test is only verified by inspection — please let CI have the final word.

### DIFF
--- a/Sources/App/Core/ErrorMiddleware.swift
+++ b/Sources/App/Core/ErrorMiddleware.swift
@@ -25,9 +25,14 @@ final class ErrorMiddleware: AsyncMiddleware {
     func respond(to req: Request, chainingTo next: AsyncResponder) async throws -> Response {
         do {
             return try await next.respond(to: req)
-        } catch let error as AbortError where error.status.code >= 400 {
-            let statusCode = error.status.code
-            let isCritical = (statusCode >= 500)
+        } catch let error as AbortError where error.status.code < 400 {
+            throw error
+        } catch {
+            // Errors that aren't `AbortError`s would otherwise escape this middleware and be
+            // turned into an unlogged, raw 500 by Vapor's default error handling.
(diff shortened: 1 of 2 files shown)


finestructure:
Thanks for the PR, @MsfPablo ! We have a few high priority PRs that we need to take of at the moment but we'll be looking at this as soon as possible.

PR #4130, opened 2026-08-10
Sync changes for deployment

kcieplak:
* Deployment is done in dev/prod pipelines so remove that pipeline code.
* Update the project spi-base image location.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,7 +18,7 @@
 # ================================
 # Build image
 # ================================
-FROM registry.gitlab.com/finestructure/spi-base:2.2.1 AS build
+FROM registry.gitlab.com/saguaro1/spi-base:2.2.2 AS build
 ARG COMPILATION_MODE="release"
 
 # Set up a build area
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -18,7 +18,7 @@
 # ================================
 # Build image
 # ================================
-FROM registry.gitlab.com/finestructure/spi-base:2.2.1 AS build
+FROM registry.gitlab.com/saguaro1/spi-base:2.2.2 AS build
```
finestructure, line 21: I don't think we can't switch these URLs over until the spi-base repo is public. CI should be failing with this set.

Dockerfile
```diff
@@ -65,7 +65,7 @@ RUN [ -d /build/Resources ] && { mv /build/Resources ./Resources && chmod -R a-w
 # ================================
 # Run image
 # ================================
-FROM registry.gitlab.com/finestructure/spi-base:2.2.1
+FROM registry.gitlab.com/saguaro1/spi-base:2.2.2
```
finestructure, line 68: dto

finestructure (approved):

Dockerfile
finestructure: Wait, how did the tests pass for these? 🤔

PR #4131, opened 2026-08-10
Retry 400 rate limit triggers

finestructure:


Sources/App/Core/Gitlab.swift
```diff
@@ -178,6 +164,65 @@ extension Gitlab.Builder {
...
+                    let webUrl = try JSONDecoder().decode(Response.self, from: responseBody).webUrl
+                    let res = Build.TriggerResponse(status: response.status, webUrl: webUrl)
+                    logger.info("Triggered build: \(res.webUrl)")
+                    return res
+                case .badRequest:
+                    struct ErrorResponse: Content { var message: String }
+                    let message = try JSONDecoder().decode(ErrorResponse.self, from: responseBody).message
+                    if message.lowercased().contains("too many pipelines created"),
+                       let request {
+                        // retry once
```
kcieplak, line 203: Just once?  Should this not be an exponential backoff type retry?  Try, and delay by N*I, then try again up to a maximum or N retries?

finestructure: I didn't want to overdo it. In my anecdotal observation a single retry worked. We're triggering quite a few jobs per batch and if we retry each one multiple times incl backoffs that process becomes quite long-running.

It's not really that important that each trigger passes right now (they'll be retried again anyway, on the build trigger's schedule), it's just that we want to increase our success rate at filling the queue.

kcieplak (approved):

PR #4133, opened 2026-08-11
Remove Plausible-based analytics

finestructure:
Trying to report analytics events could fail the smoke tests if `ANALYTICS_BACKEND_REPORTING_SITE_ID` wasn't set (and potentially with it set, too, that's unclear).

This removes all the reporting logic. It was tailored for Plausible, which we're not using anymore.

Alternatively, I could leave the instrumentation in and only disable the event posting for now in case we want to hook this up to a different system in the future.


finestructure:
@daveverwer if there's any chance we'll want to hook up API reporting to a different system, let me know I'll produce a minimal patch that preserves the instrumentation.

daveverwer:
Yes, I'd like to at least leave the option in to report some of these things *somewhere* in the future.

Having the analytics disabled but still leaving a trace of what they previously reported is useful context as to what we once did.

daveverwer (approved):

PR #4134, opened 2026-08-11
Adds the "What is a package registry?" blog post

daveverwer:
What is a package registry? on the Swift Package Index Blog – Swift Package Index

### DIFF
--- a/Resources/Blog/Posts/what-is-a-package-registry.md
+++ b/Resources/Blog/Posts/what-is-a-package-registry.md
@@ -12,7 +12,7 @@ Most language ecosystems have a registry. JavaScript has [npm](https://www.npmjs
 
 ### Depending on a Git-based package
 
-Like most package managers, SwiftPM supports using Git-based packages and registry-based packages. You will probably be familiar with how that appears in a `Package.swift` file:
+You will probably be familiar with how that appears in a `Package.swift` file:
 
 ```swift
 dependencies: [
@@ -60,7 +60,7 @@ swift package-registry set https://tuist.dev/api/registry/swift
 
 *Note: This command sets a default package registry for the current project, not [globally](https://github.com/swiftlang/swift-package-manager/blob/main/Documentation/PackageRegistry/PackageRegistryUsage.md#configuring-a-registry).*
 
-**Step 3:** Switch the dependency. Open the `Package.swift` file and change the `dependencies` section to use [`package(id:from:)`](https://developer.apple.com/documentation/packagedescription/package/dependency/package(id:from:)) instead of [`package(url:from:)`](https://developer.apple.com/documentation/packagedescription/package/dependency/package(url:from:)):
+**Step 3:** Switch the dependency. Open the `Package.swift` file and make two changes. Switch the `dependencies` section to use [`package(id:from:)`](https://developer.apple.com/documentation/packagedescription/package/dependency/package(id:from:)) instead of [`package(url:from:)`](https://developer.apple.com/documentation/packagedescription/package/dependency/package(url:from:)):
 
 ```swift
 dependencies: [
@@ -68,7 +68,7 @@ dependencies: [
 ]
 ```
 
-**Step 4:** Switch the dependency’s reference in the `executableTarget`:
+Then switch the dependency’s reference in the `executableTarget`:
 
 ```swift
 .executableTarget(
@@ -79,7 +79,7 @@ dependencies: [
 )
 ```
 
-**Step 5:** Run the tool:
+**Step 4:** Run the tool:
 
 ```shell
 swift run
--- a/Resources/Blog/Posts/what-is-a-package-registry.md
+++ b/Resources/Blog/Posts/what-is-a-package-registry.md
@@ -6,13 +6,13 @@ As we move towards that goal, we thought it would be worth revisiting what a pac
 
 ### What is a package registry?
 
-There are three main parts of the Swift package tooling ecosystem. SwiftPM resolves your dependencies ready for the Swift compiler to build. A package index (like [Swift Package Index](https://swiftpackageindex.com/)) helps you discover packages and learn about what they do. A package registry *hosts and serves* the packages themselves. It’s the infrastructure that, when asked for version 1.0.2 of a specific package, serves back a verified, immutable archive of source code.
+There are three main parts of the Swift package tooling ecosystem. SwiftPM resolves your dependencies, fetching their source code so the compiler can build them. A package index (like [Swift Package Index](https://swiftpackageindex.com/)) helps you discover packages and learn what they do. A package registry *hosts and serves* the packages themselves. It’s the infrastructure that, when asked for version 1.0.2 of a specific package, serves back a verified, immutable archive of source code.
 
 Most language ecosystems have a registry. JavaScript has [npm](https://www.npmjs.com/), Rust has [crates.io](https://crates.io/), Ruby has [RubyGems](https://rubygems.org/), and Python has [PyPI](https://pypi.org/).
 
-### Depending on a git-based package
+### Depending on a Git-based package
 
-Like most package managers, SwiftPM supports using git-based packages and registry-based packages. You will probably be familiar with how a git-based dependency looks in a `Package.swift` file:
+Like most package managers, SwiftPM supports using Git-based packages and registry-based packages. You will probably be familiar with how that appears in a `Package.swift` file:
 
 ```swift
 dependencies: [
@@ -20,7 +20,7 @@ dependencies: [
 ]
 ```
 
-During package resolution, SwiftPM clones git-based dependencies from their source repository and checks out a matching tag. It then supplies the source to the compiler, ready to build your app. This is simple and requires no extra infrastructure, but there are inherent disadvantages. Git checkouts can be slow and transfer more data than is needed, and git tags are mutable. SwiftPM uses [Trust on First Use](https://en.wikipedia.org/wiki/Trust_on_first_use) (TOFU) checks that help mitigate problems stemming from mutable tags, but they only catch mutations after first package resolution.
+During package resolution, SwiftPM clones Git-based dependencies from their source repository and checks out a matching tag. It then supplies the source code to the compiler, ready to build your app. This is simple and requires no extra infrastructure, but there are inherent disadvantages. Git checkouts can be slow and transfer more data than is needed. Git tags are also mutable, and while SwiftPM uses [Trust on First Use](https://en.wikipedia.org/wiki/Trust_on_first_use) (TOFU) checks that help mitigate that problem, TOFU only catches mutations that happen after first package resolution.
 
 ### Depending on a registry-based package
 
@@ -32,19 +32,19 @@ dependencies: [
 ]
 ```
 
-Package registries address the main downsides of git-based dependencies. Package versions are immutable once published, a registry delivers source code in an archive without any git history, minimizing download size. SwiftPM added [support for package registries](https://github.com/swiftlang/swift-package-manager/blob/main/Documentation/PackageRegistry/PackageRegistryUsage.md) in Swift 5.7, so it’s available for use right now.
+Package registries address the main downsides of Git-based dependencies. Package versions are immutable once published. A registry delivers source code in an archive without any Git history, which minimizes the download size. SwiftPM added [support for package registries](https://github.com/swiftlang/swift-package-manager/blob/main/Documentation/PackageRegistry/PackageRegistryUsage.md) in Swift 5.7, so it’s available for use right now.
 
 Fetching packages from a registry has several advantages, but some things don’t change:
 
 * Your source code still lives wherever you choose to host it, be that GitHub, GitLab, or a self-hosted solution.
 * Packages deliver source code, not pre-built binaries. You still compile your app’s code on your device.
-* Git-based package resolution continues to work, and apps can mix registry dependencies and git-based ones.
+* Git-based package resolution continues to work, and apps can mix registry dependencies and Git-based ones.
 
 It’s worth noting that package registries also don’t create a walled garden. SwiftPM can register per-project registries, global registries, scope-specific registries, and anyone can implement a package registry for SwiftPM. There are already package registry implementations from [JFrog Artifactory](https://jfrog.com/integrations/swift-repository/), [AWS CodeArtifact](https://docs.aws.amazon.com/codeartifact/latest/ug/using-swift.html), [Cloudsmith](https://cloudsmith.com/product/formats/swift), and a community package cache hosted in the [Tuist registry](https://tuist.dev/en/docs/guides/features/registry).
 
 ### Walkthrough: Depending on a package from a registry
 
-In this follow-along example, we will build a trivial SwiftPM-based command-line tool and switch the Swift Argument Parser dependency from git-based to registry-based. We’ll use the Tuist registry here as a publicly available registry that has archived every package in the [Swift Package Index Package List](https://github.com/SwiftPackageIndex/PackageList/blob/main/packages.json).
+In this follow-along example, we will build a trivial SwiftPM-based command-line tool and switch the Swift Argument Parser dependency from Git-based to registry-based. We’ll use the Tuist registry here as a publicly available registry that has archived every package in the [Swift Package Index Package List](https://github.com/SwiftPackageIndex/PackageList/blob/main/packages.json).
 
 **Step 1:** Create a “Hello, world!” package. Make a new directory and run:
 
--- a/Resources/Blog/Posts/what-is-a-package-registry.md
+++ b/Resources/Blog/Posts/what-is-a-package-registry.md
@@ -46,19 +46,19 @@ It’s worth noting that package registries also don’t create a walled garden.
 
 In this follow-along example, we will build a trivial SwiftPM-based command-line tool and switch the Swift Argument Parser dependency from git-based to registry-based. We’ll use the Tuist registry here as a publicly available registry that has archived every package in the [Swift Package Index Package List](https://github.com/SwiftPackageIndex/PackageList/blob/main/packages.json).
 
-**Step 1:** Add a registry to SwiftPM:
+**Step 1:** Create a “Hello, world!” package. Make a new directory and run:
+
+```shell
+swift package init --type tool --name TestPackage
+```
+
+**Step 2:** Add a registry to SwiftPM:
 
 ```shell
 swift package-registry set https://tuist.dev/api/registry/swift
 ```
 
 *Note: This command sets a default package registry for the current project, not [globally](https://github.com/swiftlang/swift-package-manager/blob/main/Documentation/PackageRegistry/PackageRegistryUsage.md#configuring-a-registry).*
-
-**Step 2:** Create a “Hello, world!” package. Make a new directory and run:
-
-```shell
-swift package init --type tool --name TestPackage
-```
 
 **Step 3:** Switch the dependency. Open the `Package.swift` file and change the `dependencies` section to use [`package(id:from:)`](https://developer.apple.com/documentation/packagedescription/package/dependency/package(id:from:)) instead of [`package(url:from:)`](https://developer.apple.com/documentation/packagedescription/package/dependency/package(url:from:)):
 
--- a/Resources/Blog/Posts/what-is-a-package-registry.md
+++ b/Resources/Blog/Posts/what-is-a-package-registry.md
@@ -1,0 +1,90 @@
+When we recently [announced that Swift Package Index was joining Apple](https://swiftpackageindex.com/blog/swift-package-index-joins-apple), you may have noticed one very significant line in that announcement post:
+
+> Together, we’re building a comprehensive package registry to serve the Swift community’s evolving needs.
+
+As we move towards that goal, we thought it would be worth revisiting what a package registry is and how you can use one today.
+
+### What is a package registry?
+
+There are three main parts of the Swift package tooling ecosystem. SwiftPM resolves your dependencies ready for the Swift compiler to build. A package index (like [Swift Package Index](https://swiftpackageindex.com/)) helps you discover packages and learn about what they do. A package registry *hosts and serves* the packages themselves. It’s the infrastructure that, when asked for version 1.0.2 of a specific package, serves back a verified, immutable archive of source code.
+
+Most language ecosystems have a registry. JavaScript has [npm](https://www.npmjs.com/), Rust has [crates.io](https://crates.io/), Ruby has [RubyGems](https://rubygems.org/), and Python has [PyPI](https://pypi.org/).
+
+### Depending on a git-based package
+
+Like most package managers, SwiftPM supports using git-based packages and registry-based packages. You will probably be familiar with how a git-based dependency looks in a `Package.swift` file:
+
+```swift
+dependencies: [
+    .package(url: "https://github.com/swiftlang/swift-testing.git", from: "6.3.2")
+]
+```
+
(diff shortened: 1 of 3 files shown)


daveverwer:
Reviews are welcome but we won't merge this until we're ready to publish it. Ideally tomorrow.

Resources/Blog/Posts/what-is-a-package-registry.md
```diff
@@ -0,0 +1,90 @@
+When we recently [announced that Swift Package Index was joining Apple](https://swiftpackageindex.com/blog/swift-package-index-joins-apple), you may have noticed one very significant line in that announcement post:
+
+> Together, we’re building a comprehensive package registry to serve the Swift community’s evolving needs.
+
+As we move towards that goal, we thought it would be worth revisiting what a package registry is and how you can use one today.
+
+### What is a package registry?
+
+There are three main parts of the Swift package tooling ecosystem. SwiftPM resolves your dependencies ready for the Swift compiler to build. A package index (like [Swift Package Index](https://swiftpackageindex.com/)) helps you discover packages and learn about what they do. A package registry *hosts and serves* the packages themselves. It’s the infrastructure that, when asked for version 1.0.2 of a specific package, serves back a verified, immutable archive of source code.
```
thePianoKid: I find `SwiftPM resolves your dependencies ready for the Swift compiler to build.` a bit hard to read. 

Maybe something like `SwiftPM resolves your dependencies so the Swift compiler can build them.`

thePianoKid:
```suggestion
There are three main parts of the Swift package tooling ecosystem. SwiftPM resolves your dependencies ready for the Swift compiler to build. A package index (like [Swift Package Index](https://swiftpackageindex.com/)) helps you discover packages and learn what they do. A package registry *hosts and serves* the packages themselves. It’s the infrastructure that, when asked for version 1.0.2 of a specific package, serves back a verified, immutable archive of source code.
```

Resources/Blog/Posts/what-is-a-package-registry.md
```diff
@@ -0,0 +1,90 @@
...
+
+Like most package managers, SwiftPM supports using git-based packages and registry-based packages. You will probably be familiar with how a git-based dependency looks in a `Package.swift` file:
+
+```swift
+dependencies: [
+    .package(url: "https://github.com/swiftlang/swift-testing.git", from: "6.3.2")
+]
+```
+
+During package resolution, SwiftPM clones git-based dependencies from their source repository and checks out a matching tag. It then supplies the source to the compiler, ready to build your app. This is simple and requires no extra infrastructure, but there are inherent disadvantages. Git checkouts can be slow and transfer more data than is needed, and git tags are mutable. SwiftPM uses [Trust on First Use](https://en.wikipedia.org/wiki/Trust_on_first_use) (TOFU) checks that help mitigate problems stemming from mutable tags, but they only catch mutations after first package resolution.
```
thePianoKid: May be explicit about how SPM finds and checks out the tag?

thePianoKid: > Git checkouts can be slow and transfer more data than is needed, and git tags are mutable.

Two separate thoughts in this sentence. Something like this: 

Git checkouts can be slow and transfer more data than is needed. Another problem is git tags are mutable: an attacker could make the most recent version point to a commit with their malicious code. SwiftPM uses [Trust on First Use]...

Resources/Blog/Posts/what-is-a-package-registry.md
```diff
@@ -0,0 +1,90 @@
...
+
+### Walkthrough: Depending on a package from a registry
+
+In this follow-along example, we will build a trivial SwiftPM-based command-line tool and switch the Swift Argument Parser dependency from git-based to registry-based. We’ll use the Tuist registry here as a publicly available registry that has archived every package in the [Swift Package Index Package List](https://github.com/SwiftPackageIndex/PackageList/blob/main/packages.json).
+
+**Step 1:** Add a registry to SwiftPM:
+
+```shell
+swift package-registry set https://tuist.dev/api/registry/swift
+```
```
thePianoKid: I think this should be after you create a Swift package. When I run this in a dir w/ nothing in it, I get 

```
Error: Could not find Package.swift in this directory or any of its parent directories.
```

Resources/Blog/Posts/what-is-a-package-registry.md
heckj: "SwiftPM resolves your dependencies ready for the Swift compiler to build." reads slightly oddly to me - like it's missing a word. What would you think of:


```suggestion
There are three main parts of the Swift package tooling ecosystem. SwiftPM resolves your dependencies  for the Swift compiler to build your code. A package index (like [Swift Package Index](https://swiftpackageindex.com/)) helps you discover packages and learn about what they do. A package registry *hosts and serves* the packages themselves. It’s the infrastructure that, when asked for version 1.0.2 of a specific package, serves back a verified, immutable archive of source code.
```

Resources/Blog/Posts/what-is-a-package-registry.md
```diff
@@ -0,0 +1,90 @@
...
+
+Here’s the same dependency import from a registry:
+
+```swift
+dependencies: [
+    .package(id: "swiftlang.swift-testing", from: "6.3.2")
+]
+```
+
+Package registries address the main downsides of git-based dependencies. Package versions are immutable once published, a registry delivers source code in an archive without any git history, minimizing download size. SwiftPM added [support for package registries](https://github.com/swiftlang/swift-package-manager/blob/main/Documentation/PackageRegistry/PackageRegistryUsage.md) in Swift 5.7, so it’s available for use right now.
```
heckj:
```suggestion
Package registries address the main downsides of Git-based dependencies. Package versions are immutable once published. A registry delivers source code in an archive without any Git history, which minimizes the download size. SwiftPM added [support for package registries](https://github.com/swiftlang/swift-package-manager/blob/main/Documentation/PackageRegistry/PackageRegistryUsage.md) in Swift 5.7, so it’s available for use right now.
```

Recommend breaking up the second into two distinct sentences.
What do you think of switching "minimizing" to "minimizes the download size" - which I think makes that benefit more explicit.
capitalizing Git in Git-based for consistency.

Resources/Blog/Posts/what-is-a-package-registry.md
```diff
@@ -0,0 +1,90 @@
...
+
+### What is a package registry?
+
+There are three main parts of the Swift package tooling ecosystem. SwiftPM resolves your dependencies ready for the Swift compiler to build. A package index (like [Swift Package Index](https://swiftpackageindex.com/)) helps you discover packages and learn about what they do. A package registry *hosts and serves* the packages themselves. It’s the infrastructure that, when asked for version 1.0.2 of a specific package, serves back a verified, immutable archive of source code.
+
+Most language ecosystems have a registry. JavaScript has [npm](https://www.npmjs.com/), Rust has [crates.io](https://crates.io/), Ruby has [RubyGems](https://rubygems.org/), and Python has [PyPI](https://pypi.org/).
+
+### Depending on a git-based package
+
+Like most package managers, SwiftPM supports using git-based packages and registry-based packages. You will probably be familiar with how a git-based dependency looks in a `Package.swift` file:
```
heckj:
```suggestion
Like most package managers, SwiftPM supports using Git-based packages and registry-based packages. You will probably be familiar with how that appears in a `Package.swift` file:
```

(capitalizing "Git" since you're not talking about the explicit command, and reducing the replication in the next sentence)

Resources/Blog/Posts/what-is-a-package-registry.md
heckj:
```suggestion
During package resolution, SwiftPM clones Git-based dependencies from their source repository and checks out a matching tag. It then supplies the source to the compiler, ready to build your app. This is simple and requires no extra infrastructure, but there are inherent disadvantages. Git checkouts can be slow and transfer more data than is needed, and Git tags are mutable. SwiftPM uses [Trust on First Use](https://en.wikipedia.org/wiki/Trust_on_first_use) (TOFU) checks that help mitigate problems stemming from mutable tags, but they only catch mutations after first package resolution.
```

Resources/Blog/Posts/what-is-a-package-registry.md
heckj:
```suggestion
Package registries address the main downsides of git-based dependencies. Package versions are immutable once published, a registry delivers source code in an archive without any Git history, minimizing download size. SwiftPM added [support for package registries](https://github.com/swiftlang/swift-package-manager/blob/main/Documentation/PackageRegistry/PackageRegistryUsage.md) in Swift 5.7, so it’s available for use right now.
```

Resources/Blog/Posts/what-is-a-package-registry.md
```diff
@@ -0,0 +1,90 @@
...
+]
+```
+
+Package registries address the main downsides of git-based dependencies. Package versions are immutable once published, a registry delivers source code in an archive without any git history, minimizing download size. SwiftPM added [support for package registries](https://github.com/swiftlang/swift-package-manager/blob/main/Documentation/PackageRegistry/PackageRegistryUsage.md) in Swift 5.7, so it’s available for use right now.
+
+Fetching packages from a registry has several advantages, but some things don’t change:
+
+* Your source code still lives wherever you choose to host it, be that GitHub, GitLab, or a self-hosted solution.
+* Packages deliver source code, not pre-built binaries. You still compile your app’s code on your device.
+* Git-based package resolution continues to work, and apps can mix registry dependencies and git-based ones.
```
heckj:
```suggestion
* Git-based package resolution continues to work, and apps can mix registry dependencies and Git-based ones.
```

Resources/Blog/Posts/what-is-a-package-registry.md
```diff
@@ -0,0 +1,90 @@
...
+
+It’s worth noting that package registries also don’t create a walled garden. SwiftPM can register per-project registries, global registries, scope-specific registries, and anyone can implement a package registry for SwiftPM. There are already package registry implementations from [JFrog Artifactory](https://jfrog.com/integrations/swift-repository/), [AWS CodeArtifact](https://docs.aws.amazon.com/codeartifact/latest/ug/using-swift.html), [Cloudsmith](https://cloudsmith.com/product/formats/swift), and a community package cache hosted in the [Tuist registry](https://tuist.dev/en/docs/guides/features/registry).
+
+### Walkthrough: Depending on a package from a registry
+
+In this follow-along example, we will build a trivial SwiftPM-based command-line tool and switch the Swift Argument Parser dependency from git-based to registry-based. We’ll use the Tuist registry here as a publicly available registry that has archived every package in the [Swift Package Index Package List](https://github.com/SwiftPackageIndex/PackageList/blob/main/packages.json).
```
heckj: you'll totally laugh, but I had to struggle to *not* suggest editing out the use of "we" ;-)

heckj (commented):
A couple of nits for grammar and capitalization, but broadly lovely. After reading this in depth though, I was left with two thoughts:

1) Does a registry sort out all the issues with transitive dependencies as well, or does *every* version you want use in a package need to be pre-loaded somewhere in a registry to begin with? Classic example in my world is "I'm using SwiftNIO..." - and we all know how many dependencies that can cascade into. What's it take to get the "faster download" win for those larger scenarios, and how do I get there?

The biggest wins sound like it's most effective for projects with a lot of dependencies, (or lots of transitive dependencies) after you've done all the lifting to convert your Package, and cajoled or harangued others into updating theirs so that transitive dependencies work as well. (or am I mistaking that's a requirement for transitive deps?)

When done, the download speeds should be faster, so a "package resolve" will be "a lot" (?) faster when you're doing a clean build.  Reduced local build times, reduced CI times, etc. Are there secondary wins here that are compelling that we want to highlight? The "immutability" and security perspective, for example?

2) The article highlights that there are already registries out there - even ready to use, and support for them - so what's this implementation going to be doing that the others aren't? The "We're going to work on..." setup sort of begs the question that something needs doing, but the suggestion of upcoming work and effort makes me infer that they're potentially lacking in some way, and I'm not clear why or what the gap is. What does "comprehensive" mean in the lede for this article

Resources/Blog/Posts/what-is-a-package-registry.md
daveverwer: This is such a good catch, @thePianoKid. Thanks for catching it. I did these two steps in the opposite order when making the project and then when writing about it thought "it makes sense to add the registry first". I had no idea that'd break it! Thank you for testing it ❤️

Resources/Blog/Posts/what-is-a-package-registry.md
daveverwer: Thanks. I went with:

> SwiftPM resolves your dependencies, fetching their source code so the compiler can build them.

Resources/Blog/Posts/what-is-a-package-registry.md
daveverwer: There's quite a bit that this could cover. Were you interested in hitting any specific points?

Resources/Blog/Posts/what-is-a-package-registry.md
daveverwer: Fixed, thanks.

Resources/Blog/Posts/what-is-a-package-registry.md
daveverwer: Thanks. I went with this rewrite:

> Git checkouts can be slow and transfer more data than is needed. Git tags are also mutable, and while SwiftPM uses [Trust on First Use](https://en.wikipedia.org/wiki/Trust_on_first_use) (TOFU) checks that help mitigate that problem, TOFU only catches mutations that happen after first package resolution.

Resources/Blog/Posts/what-is-a-package-registry.md
daveverwer: Yes, @davelester and I chatted about this too and agreed to leave it. I think it works here.

Resources/Blog/Posts/what-is-a-package-registry.md
daveverwer: Thanks, fixed!

Resources/Blog/Posts/what-is-a-package-registry.md
daveverwer: Thanks, fixed!

Resources/Blog/Posts/what-is-a-package-registry.md
daveverwer: Thanks, fixed!

daveverwer: Fixed, thanks!

Resources/Blog/Posts/what-is-a-package-registry.md
daveverwer: Thanks, fixed!

Resources/Blog/Posts/what-is-a-package-registry.md
```diff
@@ -0,0 +1,90 @@
...
+
+**Step 2:** Add a registry to SwiftPM:
+
+```shell
+swift package-registry set https://tuist.dev/api/registry/swift
+```
+
+*Note: This command sets a default package registry for the current project, not [globally](https://github.com/swiftlang/swift-package-manager/blob/main/Documentation/PackageRegistry/PackageRegistryUsage.md#configuring-a-registry).*
+
+**Step 3:** Switch the dependency. Open the `Package.swift` file and make two changes. Switch the `dependencies` section to use [`package(id:from:)`](https://developer.apple.com/documentation/packagedescription/package/dependency/package(id:from:)) instead of [`package(url:from:)`](https://developer.apple.com/documentation/packagedescription/package/dependency/package(url:from:)):
```
finestructure, line 63: I'm not sure if this blog post is the place to address it but a question this raises (that I don't actually know the answer to), and perhaps not just for me: what does this mean for transient dependencies?

I.e. say `apple.swift-argument-parser` had dependencies itself in its `Package.swift` that aren't declared as "registry" dependencies. How can you, or can you even, ensure that all of your dependencies are registry-based?

I'd imagine this is being addressed in the spec, so maybe it's worth a sentence and a link?

Resources/Blog/Posts/what-is-a-package-registry.md
```diff
@@ -0,0 +1,94 @@
...
+dependencies: [
+    .package(url: "https://github.com/swiftlang/swift-testing.git", from: "6.3.2")
+]
+```
+
+During package resolution, SwiftPM clones Git-based dependencies from their source repository and checks out a matching tag. It then supplies the source code to the compiler, ready to build your app. This is simple and requires no extra infrastructure, but there are inherent disadvantages. Git checkouts can be slow and transfer more data than is needed. Git tags are also mutable, and while SwiftPM uses [Trust on First Use](https://en.wikipedia.org/wiki/Trust_on_first_use) (TOFU) checks that help mitigate that problem, TOFU only catches mutations that happen after first package resolution.
+
+### Depending on a registry-based package
+
+Here’s the same dependency import from a registry:
```
davelester, line 27:
```suggestion
Here’s the same dependency, declared using a registry:
```

Resources/Blog/Posts/what-is-a-package-registry.md
```diff
@@ -0,0 +1,94 @@
...
+
+```shell
+swift run
+```
+
+You’ll see SwiftPM resolve the package dependency using the registry, build the tool, and output “Hello, world!”. Congratulations, you just used a package registry!
+
+### Why build a package registry?
+
+Given that you just used a community package registry, you might wonder why we are building a new one. The Tuist registry is a valuable resource for the community as a read-only cache implemented on a registry, and we think there's more value to be had with a full implementation. The key difference is that a fully implemented registry allows packages to be published from SwiftPM using `swift package publish`. It's much more than that, though, involving thinking about developer identity, supply chain security, package scope ownership, release management, and more.
```
davelester, line 92:
```suggestion
Given that you just used a community package registry, you might wonder why we are building a new one. The Tuist registry is a valuable resource for the community as a read-only cache implemented as a registry, and we think there's more value in a full implementation. The key difference is that a fully implemented registry allows packages to be published from SwiftPM using `swift package publish`. It's much more than that, though. A full registry involves thinking about developer identity, supply chain security, package scope ownership, release management, and more.
```

davelester: A few changes here:
 * re: on -> as, I believe this revision is correct, but will defer to the team.
 * two additional changes are stylistic / simplification but OK as-is!

Resources/Blog/Posts/what-is-a-package-registry.md
```diff
@@ -0,0 +1,94 @@
...
+
+Fetching packages from a registry has several advantages, but some things don’t change:
+
+* Your source code still lives wherever you choose to host it, be that GitHub, GitLab, or a self-hosted solution.
+* Packages deliver source code, not pre-built binaries. You still compile your app’s code on your device.
+* Git-based package resolution continues to work, and apps can mix registry dependencies and Git-based ones.
+
+It’s worth noting that package registries also don’t create a walled garden. SwiftPM can register per-project registries, global registries, scope-specific registries, and anyone can implement a package registry for SwiftPM. There are already package registry implementations from [JFrog Artifactory](https://jfrog.com/integrations/swift-repository/), [AWS CodeArtifact](https://docs.aws.amazon.com/codeartifact/latest/ug/using-swift.html), [Cloudsmith](https://cloudsmith.com/product/formats/swift), and a community package cache hosted in the [Tuist registry](https://tuist.dev/en/docs/guides/features/registry).
```
samuelmurray, line 43: I'm not suggesting this is meant to be a comprehensive list, but Sonatype Nexus also supports SPM registries since earlier this year: https://help.sonatype.com/en/swift-repositories.html

samuelmurray:
> 1. Does a registry sort out all the issues with transitive dependencies as well, or does _every_ version you want use in a package need to be pre-loaded somewhere in a registry to begin with? Classic example in my world is "I'm using SwiftNIO..." - and we all know how many dependencies that can cascade into. What's it take to get the "faster download" win for those larger scenarios, and how do I get there?
> 
> The biggest wins sound like it's most effective for projects with a lot of dependencies, (or lots of transitive dependencies) after you've done all the lifting to convert your Package, and cajoled or harangued others into updating theirs so that transitive dependencies work as well. (or am I mistaking that's a requirement for transitive deps?)

@heckj I think I can answer this, having played around some with a private registry. Short answer: a registry can handle transitive dependencies even if they are stated as git-URLs, if all metadata are correct.

First, let's take a scenario where all your dependencies and transitive dependencies are actually published in the registry. If all of the packages are re-written to have all their dependencies declared as package identifiers (`"swiftlang.swift-testing"`), then it obviously works. But it's not unlikely that some still declare their dependencies as git URLs, as you point out. Let's say you (transitively) get a dependency to https://github.com/apple/swift-log.git, and perhaps also have a direct dependency to `"apple.swift-log"`. As long as that package is uploaded to the registry with the [`repositoryURLs` metadata attribute](https://docs.swift.org/swiftpm/documentation/packagemanagerdocs/registryserverspecification#Appendix-B-Package-Release-Metadata-JSON-Schema), then the registry will manage to deduplicate this dependency, and translate between the Git URL and package identifier, as explained here: https://docs.swift.org/swiftpm/documentation/packagemanagerdocs/registryserverspecification#451-URL-to-package-identifier-mappings. The exact behaviour will depend on what flag you pass to the swift cli, i.e. `--disable-scm-to-registry-transformation` (default), `--use-registry-identity-for-scm`, or `--replace-scm-with-registry`. (With https://github.com/swiftlang/swift-package-manager/pull/9966, it seems future(?) versions of swift will also allow this to be configured globally.)

Another scenario would be that some of your (sub-)dependencies are not published in the registry. Let's say https://github.com/apple/swift-log has not yet been published, and you have a dependency to it. Then, you can't state that dependency as `"apple.swift-log"`, as there is no way for the registry to know about that package (no metadata has been published). So all packages depending on `swift-log` will have to use its git URL. SPM can resolve packages using both a registry and git, for the same project, so that should not be an issue either.

> does _every_ version you want use in a package need to be pre-loaded

If a package has a lot of previous git-based versions (i.e. tags) but only publish new versions to the registry, I'm not sure what will happen. I'm guessing that the registry will then only know about the published versions, and so in some scenarios the resolution might fail, if some exact/upper bound limits prevents the published versions from being picked. I haven't tried it out, but I would be somewhat surprised if the git-URL was used as a backup to find more versions, if the package is found in the registry.

Resources/Blog/Posts/what-is-a-package-registry.md
samuelmurray: Perhaps this comment gives the answer? https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/pull/4134#issuecomment-5284367966
Though https://github.com/swiftlang/swift-package-manager/issues/9953 is also relevant, if you mean that you want to ensure that no dependency is resolved from github.com

Resources/Blog/Posts/what-is-a-package-registry.md
daveverwer: Fixed, thanks!

Resources/Blog/Posts/what-is-a-package-registry.md
daveverwer: Thanks! I added them.

Resources/Blog/Posts/what-is-a-package-registry.md
daveverwer: Thanks! Fixed.

daveverwer:
Thanks so much for writing this up @samuelmurray! This is an important issue that likely deserves its own blog post in the future. ❤️

plemarquand (approved):

PR #4135, opened 2026-08-12
Run migrations in a transaction

plemarquand:
FluentKit's migrator calls `prepare` and then writes the `migrations`
row as two separate operations with no enclosing transaction. A
multi-step migration that throws part way commits the steps that already
succeeded *and* records nothing, so the re-run fails on "column already
exists" against a half-changed schema.

Add all migrations to an array, and then run each one through a
`TransactionalMigration` to force them to run as a transaction.

Some migrations (036, 037, 044) already run in transactions, but nesting
s safe: Postgres has no nested transactions and FluentKit makes
the inner call a passthrough.


finestructure:
FYI Paul, I'll push your branch into the server project itself so that tests can run!

finestructure (approved):
That's great, thanks Paul!

Just one note regarding deployment: Since we're not really in a good place ops-wise right now, I'd prefer if we didn't deploy this straight away.

Everything looks good wrt to ensuring we're not re-applying migrations etc but any issue here comes with big ramifications so there's a bit of risk associated with it :)

(FWIW, I ran a `migrate --revert` and a `migrate` locally on a PROD copy with this branch and it all worked fine.)

plemarquand:
Thanks Sven, no rush on deployment, whenever we're in a better spot we can move forward with it.

finestructure:
Apologies for the delay, @plemarquand - I've now rebased your branch on `main` and pushed it into the SPI repo: #4164 

Closing this one in favour of the rebased version.

PR #4136, opened 2026-08-13
Fix typo in example in FAQ

samuelmurray:
The query in the example searches for "testing", but the text says that the query will find packages matching "layout".

This is fixed so that the link text, the link URL, and text all match.


daveverwer (approved):
Thank you! Well spotted 👍

PR #4147, opened 2026-08-31
Harden analysis

finestructure:
This ended up being slightly more complicated than expected. While we _do_ have docker on the machines running analysis, the analysis job actually runs in a container, which _doesn't_. This required the change of the base image ([merge request](https://gitlab.com/saguaro1/spi-base/-/merge_requests/3)) to bring in the docker CLI and a slightly more complicated way of calling it, in order to bring the checked out package sourced into the container.

I'll add some notes to the source locations in the PR. There might be simpler ways of doing this that I'm not seeing at the moment.


.devcontainer/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM registry.gitlab.com/finestructure/spi-base:2.2.1
+FROM registry.gitlab.com/saguaro1/spi-base:27688b20334f10a3528183f6eca12adf8baea508
```
finestructure: - [x] Merge and tag spi-base and update the references when we're sure this is the way to go.

Sources/App/Commands/Analyze.swift
```diff
@@ -555,19 +555,53 @@ extension Analyze {
...
+        @Dependency(\.uuid) var uuid
         guard fileManager.fileExists(atPath: path + "/Package.swift") else {
             // It's important to check for Package.swift - otherwise `dump-package` will go
             // up the tree through parent directories to find one
             throw AppError.invalidRevision(nil, "no Package.swift")
         }
-        let json = try await shell.run(command: .swiftDumpPackage, at: path)
-        return try JSONDecoder().decode(Manifest.self, from: Data(json.utf8))
+
+        do {
```
finestructure, line 599: Below is the new dump-package sequence. A more readable version is as follows:

```
docker volume create scratch
docker run --rm -v scratch:/scratch --name temp -d busybox /bin/sleep 20
docker cp checkout temp:/scratch/package-dir
docker rm -f temp
docker run --rm --volume=scratch:/scratch --workdir=/scratch/package-dir --network=none swiftlang/swift@sha256:805c5f57d2cb73adfee6821813afbc411d414e885122dd13a741c5192e74ce57 swift package dump-package
docker volume rm -f scratch
```

It's a little awkward having to spin up the busybox container but I couldn't think of another way to copy the package directory onto the scratch volume. It feels a bit convoluted.

The checkouts _are_ actually all on a docker volume already but that's a deployment detail that's not obvious here and I'm not sure we should make use of it. If we ever change how things get deployed it's easy to accidentally break dump-package.

I've opted for the locally contained extra copy, even though it adds quite a few extra steps to the mechanism.

Do you see way we could perhaps at least avoid having to spin up the temp container?

Also, we technically don't need to copy the whole package directory. Just the `Package.swift` and (optionally) `Package-*.swift` files would be enough but `docker copy` doesn't support wild cards and I wanted to avoid having to list and loop, for now at least.

Tests/AppTests/AllTests.swift
```diff
@@ -19,6 +19,7 @@ import SnapshotTesting
 import Testing
 
 
+#warning("pre-run cleanup of temp-UUID() containers and scratch-UUID() volumes")
 @Suite(
```
finestructure, line 23: We may want to run a container cleanup before (or after) the test run so we don't litter a dev machine in case tests get interrupted but I deferred that. Due to the uuid in the names there's no risk of collisions just the annoyance of leaving stuff around if tests fail to exit cleanly.

Tests/AppTests/AnalyzerTests.swift
```diff
@@ -1137,7 +1150,12 @@ extension AllTests.AnalyzerTests {
                     .appendingPathComponent("5.9-Package-swift").path
                 let fname = tempDir.appending("/Package.swift")
                 try await ShellOut.shellOut(to: .copyFile(from: fixture, to: fname))
-                var json = try await ShellClient.liveValue.run(command: .swiftDumpPackage, at: tempDir)
+                #warning("This should actually also use the analysis image, because the Swift version might differ")
+                var json = try await ShellClient.liveValue.run(
```
finestructure: After fixing this test to use the image's Swift version I realised we really should be using the Swift version from the analysis image instead. I'll fix this up but I didn't want to delay the PR until I do.

Tests/AppTests/AllTests.swift
finestructure: Do we perhaps also want to run a clean up step on each invocation of the analysis job (not per package-dump call but before we run an analysis batch)? That way if anything should go wrong we won't litter the deployment machine with temp containers and scratch volumes.

Sources/App/Commands/Analyze.swift
```diff
@@ -555,19 +555,53 @@ extension Analyze {
...
+            let scratchVolume = "scratch-\(uuid())"
+            let tempContainer = "temp-\(uuid())"
+            let packageDir = "package-dir"
+            let json = try await run {
+                try await shell.run(command: .docker("volume", "create", scratchVolume), at: path)
+                try await shell.run(command: .docker(
+                    "run", "--rm", "-v", "\(scratchVolume):/scratch", "--name", tempContainer, "-d", "busybox", "/bin/sleep", "20"
```
plemarquand: The sleep 20 worries me a bit. If docker cp on a large checkout runs long the container will dissapear and we'll see it as a failure.

Could we bind-mount the checkout read-only instead of copying it in?

```
let json = try await shell.run(command: .docker(
  "run", "--rm",
  "--volume=\(path):/pkg:ro",
   "--workdir=/pkg",
   "--network=none",
   SwiftVersion.analysisDockerImage,
   "swift", "package",
   "--scratch-path", "/tmp/scratch", "--cache-path", "/tmp/cache",
   "dump-package"
), at: .cwd)
```

This has the added benefit of keeping the volume read-only.

If sharing sharing the checkouts mount isn't an option, docker create still gets rid of the volume and the sleep:

```
  docker create --network=none --workdir=/pkg <image> swift package dump-package 
  docker cp <path> <id>:/pkg
  docker start --attach <id>
  docker rm --force <id>
```

Sources/App/Commands/Analyze.swift
```diff
@@ -555,19 +555,53 @@ extension Analyze {
...
         }
-        let json = try await shell.run(command: .swiftDumpPackage, at: path)
-        return try JSONDecoder().decode(Manifest.self, from: Data(json.utf8))
+
+        do {
+            let scratchVolume = "scratch-\(uuid())"
+            let tempContainer = "temp-\(uuid())"
+            let packageDir = "package-dir"
+            let json = try await run {
+                try await shell.run(command: .docker("volume", "create", scratchVolume), at: path)
```
plemarquand: Any non-zero exit code from any of these `shell.run` calls will be translated in to "this is not a valid package". Made sense when it was just the `dump-package` call, but now that we're doing docker we should disambiguate a "failed to pull the image" error from a real dump-package failure so we don't miscategorize packages when there are docker issues.

Sources/App/Core/SwiftVersion+Build.swift
```diff
@@ -44,6 +44,13 @@ extension SwiftVersion {
         }
     }
 
+    /// This is the docker image we are using to run package dump in the analysis stage. It should always be the same as the docker image we use in the PackageList validation action.
+    /// https://github.com/SwiftPackageIndex/PackageList/blob/04a0d49d7747979f9a08cb86a3ab187fd019227b/.github/workflows/ci.yml#L13
+    static var analysisDockerImage: String {
```
plemarquand, line 49: nit: may fit better in `Constants.swift`

Sources/App/Commands/Analyze.swift
finestructure: ~~Yeah, it's not ideal. Actually, we can just use `sleep infinite` since we kill the container anyway.~~

A bind mount won't work, unfortunately. The problem is that we're launching the container from within a container, so the container-local path is invalid for the bind mount - it needs a host-valid path.

```
  docker create --network=none --workdir=/pkg <image> swift package dump-package 
  docker cp <path> <id>:/pkg
  docker start --attach <id>
  docker rm --force <id>
```

Oh nice, that two phase bring-up is much better!

Sources/App/Commands/Analyze.swift
finestructure: Ah yes, good catch. This might go away with the two-phase launch but I'll double check.

Sources/App/Core/SwiftVersion+Build.swift
finestructure: Yeah, that's another option. The reason I kept it here (and would rather leave it) is because on Swift version upgrades, _this_ is the file where we do the top-level Swift version bump that then triggers all the changes in the other files. So location wise it's better to keep it here but I could stick it in an extension on `Constants`?

plemarquand: ah makes sense to keep them bundled in the same place then

thePianoKid:
Could we add an E2E test to CI that runs a malicious package through all package ingestion steps and asserts that no creds were leaked?

Might be out of scope for this PR, but worth doing imo.

finestructure:
Yeah, that'd be a nice one to add. I've made a note to look into that once the dust has settled :)

Sources/App/Commands/Analyze.swift
finestructure: Ok, turns out `docker cp` behaves differently when the target isn't a volume - at least that's what I think makes the difference.

Before, we ran `docker cp foo scratch:/scratch/pkg` and we had the _contents_ of `foo` inside `/scratch/pkg`.

Now when I run `docker cp foo container:/pkg` I end up with the contents in `/pkg/foo/`. I've tried trailing slashes and nested directories, it really seems to be a difference in how the copy works.

What _does_ work is `docker cp foo/Package.swift container:/pkg`, which gives us `/pkg/Package.swift` inside the container. So I'll just do that, which is the better thing to do anyway.

finestructure:
I believe I've resolved all open PR review comments. I'll look into the optional parts but if we want to go ahead and merge and deploy, this should be ready.

Sources/App/Commands/Analyze.swift
```diff
@@ -553,23 +553,105 @@ extension Analyze {
...
+
+    static func startPackageDumpContainer(at path: String, containerName: String) async throws(ManifestError) -> String {
+        @Dependency(\.shell) var shell
+        do {
+            return try await shell.run(command: .docker(
+                "start",
+                "--attach", containerName
+            ), at: path)
+        } catch {
+            throw .packageDumpError("\(error)")
```
plemarquand, line 656: This should be `dockerFailure` not `packageDumpError`.

Sources/App/Commands/Analyze.swift
```diff
@@ -553,23 +553,105 @@ extension Analyze {
...
+    /// - Parameters:
+    ///   - path: path to the package
+    /// - Throws: Shell errors or AppError.invalidRevision if there is no Package.swift file
+    /// - Returns: `Manifest` data
+    static func parseManifest(at path: String) async throws -> Manifest {
+        let json: String
+        do {
+            json = try await dumpPackage(at: path)
+        } catch .noManifestFound, .packageDumpError {
+            throw AppError.invalidRevision(nil, "no Package.swift")
```
plemarquand, line 566: `.packageDumpError` should be its own catch and have a different error message than `"no Package.swift"`.

Sources/App/Commands/Analyze.swift
```diff
@@ -553,23 +553,105 @@ extension Analyze {
...
         @Dependency(\.fileManager) var fileManager
+        @Dependency(\.logger) var logger
         @Dependency(\.shell) var shell
-        guard fileManager.fileExists(atPath: path + "/Package.swift") else {
-            // It's important to check for Package.swift - otherwise `dump-package` will go
-            // up the tree through parent directories to find one
-            throw AppError.invalidRevision(nil, "no Package.swift")
+        @Dependency(\.uuid) var uuid
+
+        let manifests = (try? fileManager.contentsOfDirectory(atPath: path)
```
plemarquand: I think this filter needs to be narrowed; a file like "PackageDescription.swift" would match. The pattern SwiftPM uses is here: https://github.com/swiftlang/swift-package-manager/blob/main/Sources/PackageRegistryCommand/PackageRegistryCommand%2BPublish.swift#L437

finestructure: Yeah, this isn't as narrow as it could be but I thought I'd keep this simpler even if we copy an extra file or two. After all, we've just trimmed it down from copying the whole source archive? 😆 I'll see if I can filter directly based on the regex!

Sources/App/Commands/Analyze.swift
finestructure: It can be both and we can't easily distinguish the two here... A genuine package dump failure would throw here but also some docker related problem. I think it's fairly safe to ignore the docker error as any general docker related issue wouldn't even get this far. Or can you think of a sensible way to pry the two apart (ideally one that doesn't involve grepping for docker errors etc)?

plemarquand: Ah, I misread what this was doing, I thought this was still setup but this actually runs the command. Makes sense

plemarquand (approved):

Sources/App/Commands/Analyze.swift
```diff
@@ -553,23 +553,109 @@ extension Analyze {
...
+        @Dependency(\.logger) var logger
         @Dependency(\.shell) var shell
-        guard fileManager.fileExists(atPath: path + "/Package.swift") else {
-            // It's important to check for Package.swift - otherwise `dump-package` will go
-            // up the tree through parent directories to find one
-            throw AppError.invalidRevision(nil, "no Package.swift")
+        @Dependency(\.uuid) var uuid
+
+        let toplevelContents = (try? fileManager.contentsOfDirectory(atPath: path)) ?? []
+        let versionSpecificPackageManifestPattern = /#"^Package@swift-(\d+)(?:\.(\d+))?(?:\.(\d+))?.swift$"#/
```
finestructure, line 591: NB: keeping this here, because sticking it in a `static let` raises concurrency errors.

PR #4149, opened 2026-09-02
Bind mount docker socket.

kcieplak:
Harden analysis -#4147 runs the package-dump in a sub-container, and this requires the docker socket available in the analysis container.

Add bind mount of the docker socket.


finestructure:
Closing this as we've now removed `app.yml` entirely, because it was unused.

PR #4153, opened 2026-09-04
Exclude releases with future commit dates from the Recent Releases view

### LINKED ISSUE #4152: Ensure that dates in the future don't sit at the top of "Recent Releases"
Currently, we have a package with git commit and tag dates that are in the future: 

Image

Image

These are genuine commit dates. From GitHub:

Image

We should not modify these dates on the package page or elsewhere, but we should remove releases in the future from the "Recent Releases" list on the site homepage.

daveverwer:
Fixes #4152

### DIFF
--- a/Tests/AppTests/RecentViewsTests.swift
+++ b/Tests/AppTests/RecentViewsTests.swift
@@ -148,6 +148,57 @@ extension AllTests.RecentViewsTests {
         }
     }
 
+    @Test func recentReleases_excludes_future_releases() async throws {
+        // Ensure releases with commit dates in the future don't sit at the top of the list
+        // https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/issues/4152
+        try await withSPIApp { app in
+            // setup
+            do {  // 1st package has a future release and should be represented by its latest past release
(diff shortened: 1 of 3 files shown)


Tests/AppTests/RecentViewsTests.swift
```diff
@@ -148,6 +148,57 @@ extension AllTests.RecentViewsTests {
...
+            do {  // 1st package has a future release and should be represented by its latest past release
+                let pkg = Package(id: UUID(), url: "1")
+                try await pkg.save(on: app.db)
+                try await Repository(package: pkg,
+                                     defaultBranch: "default",
+                                     name: "1",
+                                     owner: "foo",
+                                     summary: "pkg 1").create(on: app.db)
+                try await Version(package: pkg,
+                                  commitDate: Date(timeIntervalSince1970: 0),
```
finestructure: No need to change this, just FYI, we also have a `.t0` extension on Date.

finestructure (approved):

Tests/AppTests/RecentViewsTests.swift
daveverwer: Changed it anyway 👍 Thanks

PR #4156, opened 2026-09-11
Clamp pagination parameters

plemarquand:
It is possible to supply extremely large values for pageSize that can return large volumes of search results.

Add Pagination, holding the bounds for page (1...1000) and pageSize (1...1024), and clamp in two places: the query decoders for search, keywords and custom collections.


Tests/AppTests/CustomCollectionControllerTests.swift
```diff
@@ -94,6 +94,30 @@ extension AllTests.CustomCollectionControllerTests {
         }
     }
 
+    @Test func Query_pagination_clamping() throws {
+        let decoder = URLEncodedFormDecoder()
+
+        do {  // below range
+            let query = try decoder.decode(CustomCollectionsController.Query.self, from: "page=0&pageSize=-1")
+            #expect(query.page == Pagination.pageRange.lowerBound)
+            #expect(query.pageSize == Pagination.pageSizeRange.lowerBound)
```
finestructure, line 103: Just wondering: do we want to use `lowerBound` (a large value) on bad input or rather use a smaller default page (I believe we have 10 or 20 or something elsewhere)?

finestructure: Oh, hold on, `lowerBound`, so `1`, right?

finestructure (approved):

finestructure:
Just realised this wasn't in the SPI org, so tests weren't running. Re-pushing the branch there for merging.

PR #4160, opened 2026-09-14
Clamp pagination parameters

plemarquand:
It is possible to supply extremely large values for pageSize that can return large volumes of search results.

Add Pagination, holding the bounds for page (1...1000) and pageSize (1...1024), and clamp in two places: the query decoders for search, keywords and custom collections.


finestructure:
Ah sorry Paul, I meant to update your PR with the push but got distracted by the reconciliation issue 😅

finestructure (approved):

